PR #28, opened 2024-08-29
Improve configuration using environment variables

evanj80-illumio:
Fix typos in `CONTRIBUTING.md`.
Fix helm deleting a new line in `clustercreds-secret.yaml`.
Rename `/cluster/onboard` URL path into `/k8_cluster/onboard`.
Use Viper to get configuration from environment variables.
Fix `OnboardResponse` to match the field names sent by the server.
Add error catching in case of a non 200 response for `Onboard`.
convert snake_case -> camelCase for our requests and responses


internal/controller/credentials_manager.go
```diff
@@ -73,14 +74,32 @@ func (am *CredentialsManager) Onboard(ctx context.Context, TlsSkipVerify bool, O
...
 	}
-	defer resp.Body.Close()
 
+	switch resp.StatusCode {
+	case http.StatusOK:
+		// 200 OK - Continue processing the response
+	case http.StatusUnauthorized:
+		// 401 Unauthorized
+		err := errors.New("unauthorized: invalid credentials")
+		am.Logger.Error(err, "Received 401 Unauthorized")
```
rlenglet, line 89: @aaronIllumio
I missed all those calls to `Logger.Error` in your PR https://github.com/illumio/cloud-operator/pull/31.
The signature is different between `logr` and `zap`. The signature in `zap` takes the string as the first parameter:
https://github.com/uber-go/zap/blob/master/logger.go#L261
So each call like:
```
Logger.Error(err, "...")
```
must be modified into:
```
Logger.Error("...", zap.Error(err))
```
or, if it's a suggared logger:
```
Logger.Errorw("...", "error", err)
```
Please fix this in another PR.

internal/controller/streaming_manager.go
```diff
@@ -26,7 +26,20 @@ type streamManager struct {
...
 
-// TODO: Create a struct that holds all of the env variables to more easily pass them in with static types
+type EnvironmentConfig struct {
+	// Whether to skip TLS certificate verification when starting a stream.
+	TlsSkipVerify bool
+	// URL of the onboarding endpoint.
+	OnboardingEndpoint string
+	// URL of the token endpoint.
+	TokenEndpoint string
+	// Client ID for onboarding.
```
rlenglet, line 36:
```suggestion
	// Client ID for onboarding. "" if not specified, i.e. if the operator is not meant to onboard itself.
```

internal/controller/streaming_manager.go
```diff
@@ -26,7 +26,20 @@ type streamManager struct {
...
+	OnboardingClientId string
+	// Client secret for onboarding.
```
rlenglet, line 38:
```suggestion
	// Client secret for onboarding. "" if not specified, i.e. if the operator is not meant to onboard itself.
```

rlenglet (dismissed):
LGTM. A couple more nnits about comments.
One action item for @aaronIllumio re: migrating logging to Zap.

internal/controller/credentials_manager.go
rlenglet: @evanj80-illumio please at least fix the logs that you modify or add in this PR.

evanj80-illumio: Yep sounds good my logs are done

aaronIllumio (approved):
LPTM

PR #29, opened 2024-09-04
Add liveness check to cloud-operator

evanj80-illumio:
Add k8s HTTP liveness probe to check that:
1. cloud-operator is able to receive the HTTP request at all
2. cloud-operator has not entered a deadlock


cmd/main.go
```diff
@@ -103,6 +113,14 @@ func main() {
 	if err := agent.Listen(agent.Options{}); err != nil {
 		logger.Errorw("Failed to start gops agent", "error", err)
 	}
+	http.HandleFunc("/healthz", healthHandler)
+
+	errChan := make(chan error, 1)
+
+	go func() {
+		errChan <- http.ListenAndServe(":8080", nil)
+		logger.Error(errChan, "healthz check server failed")
```
rlenglet, line 122: 1. You can't just silently let the goroutine exit silently in this case.
2. We switched the whole codebase to Zap. Use the correct method signature.
```suggestion
		logger.Fatal("healthz check server failed", zap.Error(errChan))
```

cmd/main.go
```diff
@@ -34,6 +35,15 @@ import (
...
 
+// healthHandler checks the health of the server and returns a status code accordingly
+func healthHandler(w http.ResponseWriter, r *http.Request) {
+	if controller.ServerIsHealthy() {
+		w.WriteHeader(http.StatusOK)
+	} else {
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+}
+
```
rlenglet, line 46: It would be better to inject the check function instead of referring to it statically, so you could unittest this.
```suggestion
// newHealthHandler returns an HTTP HandlerFunc that checks the health of the server by calling the given function and returns a status code accordingly
func newHealthHandler(checkFunc func() bool) http.HandlerFunc {
	return func(w http.ResponseWriter, r *http.Request) {
		if checkFunc() {
			w.WriteHeader(http.StatusOK)
		} else {
			w.WriteHeader(http.StatusInternalServerError)
		}
	}
}

```

cmd/main.go
```diff
@@ -103,6 +113,14 @@ func main() {
 	if err := agent.Listen(agent.Options{}); err != nil {
 		logger.Errorw("Failed to start gops agent", "error", err)
 	}
+	http.HandleFunc("/healthz", healthHandler)
```
rlenglet, line 116:
```suggestion
	http.HandleFunc("/healthz", newHealthHandler(controller.ServerIsHealthy))
```

internal/controller/streaming_manager.go
```diff
@@ -42,6 +47,15 @@ type EnvironmentConfig struct {
 }
 
 var resourceTypes = [2]string{"pods", "nodes"}
+var dd = &deadlockDetector{}
+
+// ServerIsHealthy checks if a deadlock has occured within the threaded resource listing process.
+func ServerIsHealthy() bool {
+	if dd.processingResources && time.Since(dd.timeStarted) > 5*time.Minute {
```
rlenglet, line 54: All accesses to that global variable must be protected by a Mutex.
Otherwise, you won't be able to see updates from this other goroutine.
Add a Mutex to the `deadlockDetector` struct, and `RLock` it here while reading the other fields.

internal/controller/streaming_manager.go
```diff
@@ -85,6 +99,8 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context) error {
 	}
 
 	snapshotCompleted.Add(1)
+	dd.timeStarted = time.Now()
```
rlenglet, line 102: `Lock` the `Mutex` while modifying those fields. Otherwise, other goroutines, e.g. the healthz handler won't be able to read those updates.

internal/controller/streaming_manager.go
```diff
@@ -103,6 +119,8 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context) error {
 	}
 	allResourcesSnapshotted.Wait()
 	err = resourceLister.sendResourceSnapshotComplete()
+	dd.timeStarted = time.Now()
```
rlenglet, line 122: Same here.

rlenglet, line 122: You must do the same before the error return in case `sendClusterMetadata` returns an error, above.
Every `return` statement must be covered.
I'd have suggested to do this in a `defer`, but that wouldn't work since this method blocks on `<-Done()` below until the stream closes?!
What else can we do to ensure that these fields are updated before every `return`?

rlenglet (changes requested):
Overall, that's great.
The logic works.

Fix the concurrency issue with reading/writing the shared global variable.
And I suggested to inject the check function into the HTTP handler func to make it unittestable.

cmd/main.go
rlenglet: Oh wait, I misread this.
`ListenAndServe` returns an `error`, and you're writing this error into `errChan`.
In that case, just get the error returned by `ListenAndServe` and both write it into `errChan` and pass it to `Error` as an Error field.

internal/controller/streaming_manager.go
evanj80-illumio: I see what you are saying but have a few questions

- Lets say a goroutine fails part way through its requests and cancels context, do we really need to cover those cases since now this healthcheck will just restart the pod anyway? On top of that the it will just restart the whole process again in the backoff loop and reset thet timestamp in that deadlock object after initial bootup?

rlenglet (dismissed):

evanj80-illumio:
Updated the context placement and verified watchers are working as expected

rlenglet (approved):
Great!

PR #30, opened 2024-09-04
Stream logs to server

aaronIllumio:


api/illumio/cloud/k8scluster/v1/resource_config.proto
```diff
@@ -94,8 +94,25 @@ message KubernetesResourceMutation {
   }
 }
 
+enum LogLevels {
+  LOG_LEVELS_INFO_UNSPECIFIED = 0;
+  LOG_LEVELS_DEBUG = 1;
+  LOG_LEVELS_ERROR = 2;
+  LOG_LEVELS_PANIC = 3;
+}
```
rlenglet, line 102: Also add `WARN`?

api/illumio/cloud/k8scluster/v1/resource_config.proto
```diff
@@ -94,8 +94,25 @@ message KubernetesResourceMutation {
...
+
+message KubernetesLogsRequest {
+  string logs = 1;
+}
```
rlenglet, line 106: 1) Keep the little structure and metadata we still get in a Zap hook, cf. [the `Entry` type](https://github.com/uber-go/zap/blob/d6b2d14fc541ed17b31b794afc139f03648842ca/zapcore/entry.go#L143-L150).
2) Remove the `Kubernetes` prefix as this doesn't have anything to do with k8s.
3) Future-proof the API by making the request a `oneof`, like `SendKubernetesResourcesRequest`, so we can send new types of messages into this stream later.
```suggestion
message Log {
  LogLevel level = 1;
  google.protobuf.Timestamp time = 2;
  string message = 3;
}

message SendLogsRequest {
  oneof request {
    Log log = 1;
  }
}
```

Unfortunately, in an `Entry`, the message is already serialized, i.e. we have already lost the structure (message + fields). But if you use a JSON encoder, the `message` can still be a JSON object, so it will be somewhat structured.

api/illumio/cloud/k8scluster/v1/resource_config.proto
```diff
@@ -94,8 +94,25 @@ message KubernetesResourceMutation {
...
+
+message KubernetesLogsResponse {
+  LogLevels level = 1;
+}
```
rlenglet, line 110: Same as for the request, future-proof the API by making this a `oneof`.
```suggestion
message SendLogsResponse {
  oneof response {
    // set_log_level sets the operator's log level to the given level.
    SetLogLevel set_log_level = 1;
  }
}

message SetLogLevel {
  LogLevel level = 1;
}
```

rlenglet (changes requested):
Reviewed the protobufs only so far.

api/illumio/cloud/k8scluster/v1/resource_config.proto
rlenglet: I'd actually name the `message` field `json_message` to clarify that this contains a JSON object.

fakeserver/main.go
```diff
@@ -86,6 +86,29 @@ func (s *server) SendKubernetesResources(stream pb.KubernetesInfoService_SendKub
...
+			return nil
+		}
+		if err != nil {
+			logger.Error("Error from stream", zap.Error(err))
+			return err // Return the error to terminate the stream
+		}
+
+		logs := req.Logs
+		logger.Info("Log messages")
+		logger.Info(logs)
```
rlenglet, line 108: 1. Log all the fields from the protobuf.
2. Use the correct log level from the received message.
3. Assume that the text message is JSON, decode it into a `json.RawMessage`, and log it as a structured `Object` field.
And you should do the same in the real server.
```suggestion
		var level zapcore.Level = ... // Map from the log level from the protobuf message's log level
		var message json.RawMessage // Decode the protobuf message's message string from JSON into this RawMessage...
		if ce := logger.Check(level, "Received log from cloud-operator"); ce != nil {
			ce.Write(
				zap.String("cluster_id", ...),
				zap.Object("message", message),
			)
		}
```

rlenglet (changes requested):
Some comments on how to log messages in fakeserver, which should be the same logic used in the actual server.

fakeserver/main.go
rlenglet: Hmmm calling `zap.Object` with a `json.RawMessage` needs to be tested, but it should be possible to nest the message as a JSON object.

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// ListenToLogStream will wait for responses from server and will update log level
+// depending respone contents
+func (b *BufferedGrpcWriteSyncer) ListenToLogStream() error {
+	for {
+		res, err := b.client.Recv()
+		if err == io.EOF {
+			// The client has closed the stream
+			b.logger.Info("Server has closed stream")
```
rlenglet, line 150:
```suggestion
			b.logger.Info("Server has closed the SendLogs stream")
```

internal/controller/resource_manager.go
```diff
@@ -26,8 +26,8 @@ type ResourceManager struct {
 
 // sendResourceSnapshotComplete sends a message to indicate that the initial inventory snapshot has been completely streamed into the given stream.
 func (rm *ResourceManager) sendResourceSnapshotComplete() error {
-	if err := rm.streamManager.instance.stream.Send(&pb.SendKubernetesResourcesRequest{Request: &pb.SendKubernetesResourcesRequest_ResourceSnapshotComplete{}}); err != nil {
-		rm.logger.Errorw("Falied to send resource snapshot complete", "error", err)
+	if err := rm.streamManager.instance.resourceStream.Send(&pb.SendKubernetesResourcesRequest{Request: &pb.SendKubernetesResourcesRequest_ResourceSnapshotComplete{}}); err != nil {
+		rm.logger.Error(err, "Falied to send resource snapshot complete")
```
rlenglet, line 30:
```suggestion
		rm.logger.Errorw("Failed to send resource snapshot complete", "error", err)
```

internal/controller/resource_manager.go
```diff
@@ -47,8 +47,8 @@ func (rm *ResourceManager) sendClusterMetadata(ctx context.Context) error {
 	if err != nil {
 		rm.logger.Errorw("Error getting Kubernetes version", "error", err)
 	}
-	if err := rm.streamManager.instance.stream.Send(&pb.SendKubernetesResourcesRequest{Request: &pb.SendKubernetesResourcesRequest_ClusterMetadata{ClusterMetadata: &pb.KubernetesClusterMetadata{Uid: clusterUid, KubernetesVersion: kubernetesVersion.String(), OperatorVersion: version.Version()}}}); err != nil {
-		rm.logger.Errorw("Failed to send cluster metadata", "error", err)
+	if err := rm.streamManager.instance.resourceStream.Send(&pb.SendKubernetesResourcesRequest{Request: &pb.SendKubernetesResourcesRequest_ClusterMetadata{ClusterMetadata: &pb.KubernetesClusterMetadata{Uid: clusterUid, KubernetesVersion: kubernetesVersion.String(), OperatorVersion: version.Version()}}}); err != nil {
+		rm.logger.Error(err, "Failed to send cluster metadata")
```
rlenglet, line 51:
```suggestion
		rm.logger.Errorw("Failed to send cluster metadata", "error", err)
```

internal/controller/streaming_helper.go
```diff
@@ -10,8 +10,8 @@ import (
 // sendObjectMetaData sends a KubernetesMetadata to CloudSecure into the given stream.
 // Its used for the intial boot up of the operator so that is can stream everything currently in the cluster.
 func sendObjectMetaData(sm *streamManager, metadata *pb.KubernetesObjectMetadata) error {
-	if err := sm.instance.stream.Send(&pb.SendKubernetesResourcesRequest{Request: &pb.SendKubernetesResourcesRequest_ResourceMetadata{ResourceMetadata: metadata}}); err != nil {
-		sm.logger.Errorw("Failed to send resource metadata", "error", err)
+	if err := sm.instance.resourceStream.Send(&pb.SendKubernetesResourcesRequest{Request: &pb.SendKubernetesResourcesRequest_ResourceMetadata{ResourceMetadata: metadata}}); err != nil {
+		sm.logger.Error(err, "Failed to send resource metadata")
```
rlenglet, line 14:
```suggestion
		sm.logger.Errorw("Failed to send resource metadata", "error", err)
```

rlenglet (changes requested):
Finished reviewing the rest of the PR...
I like your logic in `BufferedGrpcWriteSyncer`!
But I'm afraid that you won't have any guarantee that each call to `Write` will get the contents of exactly one full log entry.
It would be safer (and IMHO cleaner) to implement this as a Zap hook instead.

fakeserver/main.go
```diff
@@ -53,6 +55,21 @@ type server struct {
...
 
+// LogEntry represents the structure of a zapcore.Entry with additional fields
+type LogEntry struct {
+	Level   string          `json:"Level"`
+	Time    string          `json:"Time"`
+	Message string          `json:"Message"`
+	Caller  string          `json:"Caller"`
+	Stack   string          `json:"Stack"`
+	Fields  json.RawMessage `json:"Fields"` // For additional fields
+}
```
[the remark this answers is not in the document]
rlenglet: Actually, apart from the `level`, you don't need to do anything special with any the fields. Just make this a `map[string]any`:
```suggestion
// LogEntry represents the structure of a zapcore.Entry encoded using Zap's JSON encoder with the production encoder config.
type LogEntry map[string]any

// Level returns the log level of the LogEntry.
func (l LogEntry) Level() (zapcore.Level, error) {
	var levelStr string

	if levelStr, found := l["level"]; !found {
		return zapcore.InvalidLevel, errors.New("no level field found in log entry")
	}

	var level zapcore.Level
	if err := level.UnmarshalText([]byte(levelStr); err != nil {
		return zapcore.InvalidLevel, fmt.Errorf("invalid level field %s found in log entry: %w", levelStr, err)
	}

	return level
}
```

fakeserver/main.go
```diff
@@ -158,6 +212,54 @@ func tokenAuthStreamInterceptor(expectedToken string) grpc.StreamServerIntercept
...
+	var data map[string]interface{}
+	if err := json.Unmarshal(r.Raw, &data); err != nil {
+		return err
+	}
+	// Use the encoder to add the fields to the log
+	for k, v := range data {
+		enc.AddReflected(k, v)
+	}
+	return nil
+}
```
rlenglet, line 226:
```suggestion
```

fakeserver/main.go
```diff
@@ -158,6 +212,54 @@ func tokenAuthStreamInterceptor(expectedToken string) grpc.StreamServerIntercept
...
+
+func recordStreamedLog(log *pb.Log, logger zap.Logger) error {
+	// Decode the JSON-encoded string into a LogEntry struct
+	var logEntry LogEntry
```
rlenglet, line 230:
```suggestion
	var logEntry LogEntry = make(map[string]any)
```

fakeserver/main.go
```diff
@@ -158,6 +212,54 @@ func tokenAuthStreamInterceptor(expectedToken string) grpc.StreamServerIntercept
...
+
+	// Wrap the rawMessage in RawMessageWrapper
+	wrappedRawMessage := &RawMessageWrapper{Raw: rawMessage}
+
+	if ce := logger.Check(level, "Received log from cloud-operator"); ce != nil {
+		ce.Write(
+			//zap.String("cluster_id", ...),
+			zap.Object("message", wrappedRawMessage),
+		)
+	}
```
rlenglet, line 259: As I commented above, I think you can avoid doing this wrapping by just having `LogEntry` implement `ObjectMarshaler` directly.
```suggestion
	if ce := logger.Check(level, "Received log from cloud-operator"); ce != nil {
		ce.Write(
			//zap.String("cluster_id", ...),
			zap.Object("message", logEntry),
		)
	}
```

rlenglet (changes requested):

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+	client    pb.KubernetesInfoService_SendLogsClient
+	conn      *grpc.ClientConn
+	buffer    []zapcore.Entry
+	mutex     sync.Mutex
+	done      chan struct{}
+	logger    *zap.SugaredLogger
+	logLevel  *zap.AtomicLevel
+	encoder   zapcore.Encoder
+	lostLogs  bool
+	lostBytes int
```
rlenglet, line 37: You're counting log entries (which is good!), not bytes.
```suggestion
	lostLogEntries int
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+			b.lostBytes += maxBufferSize - len(b.buffer)
+		}
+		return
+	}
+
+	for _, logEntry := range b.buffer {
+		if err := b.sendLog(logEntry); err != nil {
+			b.logger.Errorw("Failed to send log",
+				"error", err,
+				zap.Any("logBufferSize", len(b.buffer)),
```
rlenglet, line 80:
```suggestion
				zap.Int("log_buffer_size", len(b.buffer)),
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+			)
```
rlenglet, line 81:
```suggestion
			b.logger.Error("Failed to send log entry",
				zap.Error(err),
				zap.Int("buffered_log_entries_count", len(b.buffer)),
			)
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+			b.lostLogs = true
+			b.lostBytes += len(logEntry.Message)
```
rlenglet, line 83: I think it's more interesting to count lost entries than lost bytes.
```suggestion
			b.lostLogEntries += 1
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+			return
```
rlenglet, line 84: You can't just return here, since you may have successfully sent some log entries before starting to fail.
And those entries are still in the buffer. You must remove them from the buffer before returning.

internal/controller/logger_manager.go
rlenglet, line 37: Make those local variables in the `flush` method. Nothing outside of that method should refer to them.
And their values should not be persisted across calls to `flush`.
```suggestion
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+	b.buffer = append(b.buffer, entry)
+	if len(b.buffer) >= maxBufferSize {
+		b.flush()
+	}
+}
+
+// lostLogMessage generates a message about lost logs
+func (b *BufferedGrpcWriteSyncer) lostLogMessage() string {
+	return "Lost logs due to buffer overflow. Total lost bytes: " + strconv.Itoa(b.lostBytes)
+}
```
rlenglet, line 204: No need for a separate method for this. Just inline this.
And pass the lost count as a Zap field. Keep logs structured and avoid doing string formatting in log messages.

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+		lostLogMessage := zapcore.Entry{
+			Level:   zapcore.ErrorLevel,
+			Time:    time.Now(),
+			Message: b.lostLogMessage(),
+		}
+		err := b.sendLog(lostLogMessage)
+		if err != nil {
+			b.logger.Warn("Error while sending log",
+				"error", err,
+			)
```
rlenglet, line 100:
```suggestion
			b.logger.Warn("Error while sending log", zap.Error(err))
```

api/illumio/cloud/k8scluster/v1/resource_config.proto
```diff
@@ -94,8 +94,56 @@ message KubernetesResourceMutation {
...
+  // Error log level indicating something went wrong.
+  LOG_LEVEL_ERROR = 4;
+}
+
+// Message representing a log entry.
+message LogEntry {
+  // The log level of the log entry.
+  LogLevel level = 1;
+  // The timestamp when the log entry was recorded.
+  google.protobuf.Timestamp time = 2;
```
rlenglet, line 116: You don't need those. The server now extracts the level from the `json_message`, and it doesn't use the `time`.
```suggestion
```

fakeserver/main.go
```diff
@@ -158,6 +224,39 @@ func tokenAuthStreamInterceptor(expectedToken string) grpc.StreamServerIntercept
...
+	for k, v := range l {
+		err := enc.AddReflected(k, v)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func recordStreamedLog(log *pb.LogEntry, logger zap.Logger) error {
```
rlenglet, line 238:
```suggestion
func logReceivedLogEntry(log *pb.LogEntry, logger zap.Logger) error {
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+func (b *BufferedGrpcWriteSyncer) sendLog(logEntry zapcore.Entry) error {
+	buf, err := b.encoder.EncodeEntry(logEntry, nil)
+	if err != nil {
+		return err
+	}
+	err = b.client.Send(&pb.SendLogsRequest{
+		Request: &pb.SendLogsRequest_LogEntry{
+			LogEntry: &pb.LogEntry{
+				Level:       pb.LogLevel(logEntry.Level),
+				Time:        timestamppb.New(logEntry.Time),
```
rlenglet, line 133: Those are not necessary.
```suggestion
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+		err := b.sendLog(lostLogMessage)
+		if err != nil {
+			b.logger.Warn("Error while sending log",
+				"error", err,
+			)
+		}
+		b.lostLogs = false
+		b.lostBytes = 0
+	}
```
rlenglet, line 163: Remove this. That should already be taken care of at the bottom of `flush`.
You should call `flush`, though, in case there are currently buffered entries.
```suggestion
	b.flush()
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+		switch res.Response.(type) {
+		case *pb.SendLogsResponse_SetLogLevel:
+			newLevel := res.GetSetLogLevel().Level
+			b.updateLogLevel(newLevel)
+		}
+	}
+}
+
+// bufferLog adds the log entry to in-memory buffer
+func (b *BufferedGrpcWriteSyncer) bufferLog(entry zapcore.Entry) {
```
rlenglet, line 191: `zapcore.Entry` is a `struct`, so pass it by reference:
```suggestion
func (b *BufferedGrpcWriteSyncer) bufferLogEntry(entry *zapcore.Entry) {
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+		b.logLevel.SetLevel(zapcore.WarnLevel)
+	default:
+		b.logger.Warn("Unknown log level received, defaulting to INFO")
+		b.logLevel.SetLevel(zapcore.InfoLevel)
+	}
+}
+
+// NewGrpclogger will define a new zap logger with multiple writesyncs
+// one to stdout and one for GRPC writestream
+func NewGrpclogger(grpcSyncer *BufferedGrpcWriteSyncer) *zap.SugaredLogger {
```
rlenglet, line 229: By convention, acronyms are CAPITALIZED in Golang symbols.
```suggestion
func NewGRPClogger(grpcSyncer *BufferedGrpcWriteSyncer) *zap.SugaredLogger {
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,267 @@
...
+// BufferedGrpcWriteSyncer is a custom zap writesync that writes to a grpc stream
+// In case stream is not connected it will buffer to memory
+type BufferedGrpcWriteSyncer struct {
+	client    pb.KubernetesInfoService_SendLogsClient
+	conn      *grpc.ClientConn
+	buffer    []zapcore.Entry
+	mutex     sync.Mutex
+	done      chan struct{}
+	logger    *zap.SugaredLogger
+	logLevel  *zap.AtomicLevel
```
rlenglet, line 34: No need to use a pointer here. Looks like this type is passed by value everywhere. It itself encapsulates a pointer, so it's all good.
```suggestion
	logLevel  zap.AtomicLevel
```

rlenglet (changes requested):
Looks better!
Some remaining issues, esp. around the handling of lost log entries.

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,236 @@
...
+	buffer   []*zapcore.Entry
+	mutex    sync.Mutex
+	done     chan struct{}
+	logger   *zap.SugaredLogger
+	logLevel zap.AtomicLevel
+	encoder  zapcore.Encoder
+}
+
+// NewBufferedGrpcWriteSyncer returns a new BufferedGrpcWriteSyncer
+func NewBufferedGrpcWriteSyncer(client pb.KubernetesInfoService_SendLogsClient, conn ClientConnInterface) *BufferedGrpcWriteSyncer {
```
rlenglet, line 41: Since this is always passed `(nil, nil)`, you could remove those parameters altogether.
```suggestion
func NewBufferedGrpcWriteSyncer() *BufferedGrpcWriteSyncer {
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+// UpdateClient will update BufferedGrpcWriteSyncer with new client stream and GRPC connection
+func (b *BufferedGrpcWriteSyncer) UpdateClient(client pb.KubernetesInfoService_SendLogsClient, conn ClientConnInterface) {
+	b.mutex.Lock()
+	b.client = client
+	b.conn = conn
+	b.mutex.Unlock()
+	b.flush()
+}
```
rlenglet, line 138: This must also create a new channel and assign it to `done`.

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	lostLogs := false
+	lostLogEntries := 0
+
+	for _, logEntry := range b.buffer {
+		if err := b.sendLog(logEntry); err != nil {
+			b.logger.Error("Failed to send log entry",
+				zap.Error(err),
+				zap.Int("buffered_log_entries_count", len(b.buffer)),
+			)
```
rlenglet, line 84: Remove this. This is emitting a log entry for every log entry that can't be sent.
The log at the bottom of the method is more useful, and less noisy.
```suggestion
```

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,236 @@
...
+	"time"
+
+	pb "github.com/illumio/cloud-operator/api/illumio/cloud/k8scluster/v1"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"google.golang.org/grpc/connectivity"
+)
+
+const (
+	maxBufferSize = 100
```
rlenglet, line 18:
```suggestion
	maxBufferSize = 2048
```

internal/controller/streaming_manager.go
```diff
@@ -32,6 +34,18 @@ type streamManager struct {
...
+}
+
+func (w *ClientConnWrapper) GetState() connectivity.State {
+	return w.conn.GetState()
+}
+
+func (w *ClientConnWrapper) Close() error {
+	return w.conn.Close()
+}
+
```
rlenglet, line 48: Remove. You should already be able to convert a `ClientConn` to a `ClientConnInterface`.

internal/controller/streaming_manager.go
```diff
@@ -186,11 +209,16 @@ func ExponentialStreamConnect(ctx context.Context, logger *zap.SugaredLogger, en
...
 		}
-		sm, err := NewStream(ctx, logger, conn)
+		sm, err := NewStreams(ctx, logger, conn)
 		if err != nil {
 			logger.Errorw("Failed to create a new stream", "error", err)
 			continue
 		}
+
+		// Update the gRPC client and connection in BufferedGrpcWriteSyncer
+		bufferedGrpcSyncer.UpdateClient(sm.instance.logStream, &ClientConnWrapper{conn: sm.instance.conn})
```
rlenglet, line 219:
```suggestion
		bufferedGrpcSyncer.UpdateClient(sm.instance.logStream, sm.instance.conn)
```

rlenglet (changes requested):

internal/controller/logger_manager.go
```diff
@@ -0,0 +1,236 @@
...
+			lostLogsErrs = append(lostLogsErrs, err)
+		}
+	}
+	b.buffer = b.buffer[:0]
+
+	// If previously logs were lost, send a log message to indicate that
+	if lostLogs {
+		b.logger.Error("Lost logs due to buffer overflow",
+			zap.Int("lost_log_entries", lostLogEntries),
+			zap.Error(lostLogsErrs[len(lostLogsErrs)-1]),
```
rlenglet, line 93:
```suggestion
	var lostLogsErr error

	for _, logEntry := range b.buffer {
		if err := b.sendLog(logEntry); err != nil {
			lostLogs = true
			lostLogEntries += 1
			lostLogsErr = err
		}
	}
	b.buffer = b.buffer[:0]

	// If previously logs were lost, send a log message to indicate that
	if lostLogs {
		b.logger.Error("Lost logs due to buffer overflow",
			zap.Int("lost_log_entries", lostLogEntries),
			zap.Error(lostLogsErr),
```

rlenglet (changes requested):

rlenglet (approved):
Thank you, looks great!

PR #32, opened 2024-09-05
Fix error log signatures for zap

aaronIllumio:
Fixing log signatures when transitioning from `logr.Logger` -> `zap.SugaredLogger`


internal/controller/credentials_manager.go
```diff
@@ -60,14 +60,14 @@ func (am *CredentialsManager) Onboard(ctx context.Context, TlsSkipVerify bool, O
 	// Convert the data to JSON
 	jsonData, err := json.Marshal(data)
 	if err != nil {
-		am.Logger.Error(err, "Unable to marshal json data")
+		am.Logger.Error("Unable to marshal json data", "error", err)
```
rlenglet, line 63: This doesn't do what you think. You need either to use `Errorw` (notice the `w` at the end), or explicitly pass a field:
```suggestion
		am.Logger.Errorw("Unable to marshal json data", "error", err)
```
or 
```suggestion
		am.Logger.Error("Unable to marshal json data", zap.Error(err))
```
Same with all other calls.

rlenglet (dismissed):

evanj80-illumio:
Noticed some weird spacing issues with the new logging. Once https://github.com/illumio/cloud-operator/pull/28 gets merged in please run through process of using fakeserver and operator and verify logs are easy to read.

evanj80-illumio (approved):
LGTM, also ran integration fakeserver tests and logs look good!

PR #34, opened 2024-09-09
Revise helm default values and publish Docker image into GHCR

evanj80-illumio:
Revise helm values function to be more in line with standard values.
Publish Docker image into GHCR.


.github/workflows/build_and_release.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract version from file
```
rlenglet, line 26: Extract this from the git `main` branch tag?
Cilium's workflows does this:
https://github.com/cilium/cilium/blob/896d0420f11c8eefe317e77a2a43805e63f2c3bd/.github/workflows/build-images-ci.yaml#L103-L115
https://github.com/cilium/cilium/blob/896d0420f11c8eefe317e77a2a43805e63f2c3bd/.github/workflows/build-images-ci.yaml#L171-L173

BTW they also do image signing using `cosign`, which you might want to look into:
https://github.com/cilium/cilium/blob/896d0420f11c8eefe317e77a2a43805e63f2c3bd/.github/workflows/build-images-ci.yaml#L221-L230

VERSION
rlenglet, line 1: Remove. You should get the version from the git `main` branch's tag, if there's one, or the commit SHA otherwise

.github/workflows/build_and_release.yml
rlenglet, line 1: Do you want to add publishing the Helm chart into GHCR in this PR, or will you do that in another PR?
Cf. https://helm.sh/docs/topics/registries/

cloud-operator/values.yaml
```diff
@@ -4,12 +4,9 @@ defaultNamespace: "illumio-cloud"
...
 priorityClassName: ""
 image:
-  repository: "illumioevanj80/operator"
-  tag: "latest"
-  pullPolicy: Always
-
-imagePullSecrets:
-  - name: regcred
+  repository: "ghcr.io/illumio/cloud-operator"
+  tag: "dev"
```
rlenglet, line 8: This tag will have to be updated at the time the chart is published into GHCR.

rlenglet (changes requested):
You should get the version from the git tag or commit SHA.

.github/workflows/build_and_release.yml
evanj80-illumio: Going to add in this PR

rlenglet: I found one GitHub action that does this. I don't think it is exactly what we want, but it's probably a good starting point for a custom workflow for our repo: https://github.com/appany/helm-oci-chart-releaser/blob/main/action.yaml

evanj80-illumio: Yeah this is good! Had seen something similar while I was looking last night

cloud-operator/Chart.yaml
```diff
@@ -3,6 +3,5 @@
 apiVersion: v1
 name: cloud-operator
 description: A Helm chart for deploying the Illumio cloud-operator application
-version: 0.0.1
-appVersion: "0.0.1"
+version: "1.0.0"
```
evanj80-illumio, line 6: Version here and version in values.yaml should mirror what the version is tagged in GHA from the build

.github/workflows/build_and_release.yml
```diff
@@ -0,0 +1,60 @@
+name: Build and Push Docker Image
+
+on:
+  push:
+    branches:
+      - main
```
evanj80-illumio, line 6: Look at terraform workflow, how to trigger this to only run on release

evanj80-illumio: with proper v* tags

.github/workflows/build_and_release.yml
```diff
@@ -35,16 +35,16 @@ jobs:
...
+      # - name: Determine version from Git
+      #   id: determine_version
+      #   run: |
+      #     # Get the latest tag reachable from the current commit
+      #     VERSION=$(git describe --tags --abbrev=0 2>/dev/null || echo "")
+      #     if [ -z "$VERSION" ]; then
+      #       # If no tag is found, use the commit SHA
+      #       VERSION=${GITHUB_SHA}
+      #     fi
+      #     echo "VERSION=$VERSION" >> $GITHUB_ENV
```
rlenglet, line 47: You should delete instead of commenting out.
```suggestion
```

.github/workflows/build_and_release.yml
rlenglet: For reference:
https://github.com/illumio/terraform-provider-illumio-cloudsecure/blob/0336a5268576a5210edc9a841906952eafb86bff/.github/workflows/release.yml#L13-L16

rlenglet (changes requested):
Looks good! I think the only remaining piece is updating the GHA to run only on tags on `main`?

evanj80-illumio:
> Looks good! I think the only remaining piece is updating the GHA to run only on tags on `main`?

Yes think so! Forgot to push those changes, looks like it was just local so will be pushing that now!

rlenglet (approved):
LGTM, thanks!

rlenglet (approved):

PR #36, opened 2024-09-11
Fix filename for helm push to OCI in Github Action

evanj80-illumio:
Helm push failing due to missing `-` and incorrect pathing.


evanj80-illumio:
Screenshot 2024-09-11 at 11 15 42 AM

rlenglet (dismissed):

evanj80-illumio:
Screenshot 2024-09-11 at 11 23 06 AM
Just double checked before merge. File path does not need to be updated due to the fact that helm package creates the package in the directory its called, not the helm chart directory it is packaging.

PR #38, opened 2024-09-12
Flush buffer before new log entry in Zap hook

aaronIllumio:
In the cloud-operator's Zap hook, when logging a new log entry, and the client is connected, the log entry is directly sent to the server, even if there are already log entries buffered that have not been flushed yet.
That results in logs being sent out of order to the server.

### DIFF
--- a/internal/controller/logger_manager.go
+++ b/internal/controller/logger_manager.go
@@ -219,6 +219,9 @@ func NewGRPClogger(grpcSyncer *BufferedGrpcWriteSyncer) *zap.SugaredLogger {
 			grpcSyncer.bufferLogEntry(&entry)
 			return nil
 		}
+
+		grpcSyncer.flush()
+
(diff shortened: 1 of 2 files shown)


internal/controller/logger_manager.go
```diff
@@ -219,6 +219,9 @@ func NewGRPClogger(grpcSyncer *BufferedGrpcWriteSyncer) *zap.SugaredLogger {
 			grpcSyncer.bufferLogEntry(&entry)
 			return nil
 		}
+
```
evanj80-illumio, line 222: Maybe add a comment here to briefly describe why your going to flush buffer every time this hook is called.

aaronIllumio: added

evanj80-illumio (approved):

rlenglet (approved):

rlenglet (approved):

rlenglet (approved):

PR #46, opened 2024-09-18
Rename files and symbols

aaronIllumio:
Most filenames and types in [the cloud-operator implementation](https://github.com/illumio/cloud-operator/tree/main/internal/controller) are suffixed with "manager" or "helper".
Those suffixes provide no information and make the code and the logs (the caller field in each log unnecessarily verbose.
Remove those suffixes and generally choose concise names for each type and each file.
For example, type CacheManager should be renamed into Cache.

Format timestamps in RFC 3339 format in logs.


internal/controller/client.go
evanj80-illumio, line 1: Based on the contents of this file maybe `client` is not even relevant at this point.

internal/controller/metadata.go
evanj80-illumio, line 1: Kind of the same comment as the `cluster`. Seems like this name doesnt really fit and neither does the old one based on the functions present (At one point this file had a lot to do with getting different data about k8s and now its a shell of what it was 😆 )

internal/controller/credentials_manager.go
rlenglet, line 1: This file really implements the logic for onboarding.
Managing credentials is only part of that logic.
I'd rename this file to reflect that, e.g. `onboarding.go`.

And now it becomes apparent that `SetUpOAuthConnection` doesn't belong here. It doesn't have anything to do with onboarding. It has something to do with secrets, so I'd rather move it to `secrets.go`.

internal/controller/credentials_manager_test.go
rlenglet, line 1: Same here, I'd rename this file into `onboarding_test.go`.

internal/controller/credentials_manager.go
rlenglet, line 1: I'm not very happy with the `CredentialsManager` name, esp. because of the "manager" suffix.
I don't think we should have a `CredentialsManager` struct in the first place. I'd make `Onboard` a top-level function and remove the `CredentialsManager` struct.

internal/controller/secrets.go
```diff
@@ -94,3 +100,30 @@ func (secrets *Secrets) WriteK8sSecret(ctx context.Context, keyData OnboardRespo
 	}
 	return nil
 }
+
+// NewClientSet returns a new Kubernetes clientset based on the execution environment.
+func NewClientSet() (*kubernetes.Clientset, error) {
```
rlenglet, line 105: Why is this function in a file named `secrets.go`?

internal/controller/resource_manager.go
```diff
@@ -83,8 +83,8 @@ func (r *ResourceManager) DyanmicListAndWatchResources(ctx context.Context, canc
 }
 
 // DynamicListResources lists a specifed resource dynamically and sends down the current gRPC stream.
-func (r *ResourceManager) DynamicListResources(ctx context.Context, resource string) (string, CacheManager, error) {
-	cm := CacheManager{cache: make(map[string][32]byte)}
+func (r *ResourceManager) DynamicListResources(ctx context.Context, resource string) (string, Cache, error) {
+	cm := Cache{cache: make(map[string][32]byte)}
```
rlenglet, line 87:
```suggestion
	cache := &Cache{cache: make(map[string][32]byte)}
```

internal/controller/resource_manager.go
```diff
@@ -83,8 +83,8 @@ func (r *ResourceManager) DyanmicListAndWatchResources(ctx context.Context, canc
 }
 
 // DynamicListResources lists a specifed resource dynamically and sends down the current gRPC stream.
-func (r *ResourceManager) DynamicListResources(ctx context.Context, resource string) (string, CacheManager, error) {
-	cm := CacheManager{cache: make(map[string][32]byte)}
+func (r *ResourceManager) DynamicListResources(ctx context.Context, resource string) (string, Cache, error) {
```
rlenglet, line 86: The `Cache` should be passed by reference.
```suggestion
func (r *ResourceManager) DynamicListResources(ctx context.Context, resource string) (string, *Cache, error) {
```

internal/controller/resource_manager.go
```diff
@@ -115,7 +115,7 @@ func (r *ResourceManager) DynamicListResources(ctx context.Context, resource str
 
 // watchEvents watches Kubernetes resources and updates cache based on events.
 // Any occurring errors are sent through errChanWatch. The watch stops when ctx is cancelled.
-func (r *ResourceManager) watchEvents(ctx context.Context, resource string, watchOptions metav1.ListOptions, c CacheManager) error {
+func (r *ResourceManager) watchEvents(ctx context.Context, resource string, watchOptions metav1.ListOptions, c Cache) error {
```
rlenglet, line 118:
```suggestion
func (r *ResourceManager) watchEvents(ctx context.Context, resource string, watchOptions metav1.ListOptions, cache *Cache) error {
```

internal/controller/streaming_manager.go
rlenglet, line 1: 1. I don't understand why we need to separate `streamClient` and `streamManager`. Merge those into a single struct. Naming that new type `streamClient` or even just `client`/`Client` sounds OK to me.
1. Rename the file into something simpler, e.g. `stream.go`.

rlenglet, line 1: I'd move `EnvironmentConfig` into its own `config.go` file.

rlenglet (changes requested):
Thanks for the cleanup!
I made a few suggestions to clean up further.

internal/controller/secrets.go
aaronIllumio: renamed to authenticator

internal/controller/testhelper/set_up_test_cluster.go
```diff
@@ -3,12 +3,14 @@
...
 
 import (
+	"fmt"
 	"os/exec"
 )
 
 // setupTestCluster creates a new KIND cluster for testing.
 func SetupTestCluster() error {
 	cmd := exec.Command("kind", "create", "cluster", "--name", "my-test-cluster", "--config", "../../kind-config.yaml")
+	fmt.Println(cmd)
```
evanj80-illumio, line 13: Please take this out

internal/controller/streams.go
```diff
@@ -62,7 +44,7 @@ func ServerIsHealthy() bool {
 }
 
 // NewStream returns a new stream.
-func NewStreams(ctx context.Context, logger *zap.SugaredLogger, conn *grpc.ClientConn) (*streamManager, error) {
+func NewStreams(ctx context.Context, logger *zap.SugaredLogger, conn *grpc.ClientConn) (*streamClient, error) {
```
rlenglet, line 47: Match the type name in the constructor name.
```suggestion
func NewStreamClient(ctx context.Context, logger *zap.SugaredLogger, conn *grpc.ClientConn) (*streamClient, error) {
```

rlenglet (changes requested):
A couple comments still need to be resolved, but this is already looking good!

rlenglet (approved):
LGTM, thank you!!

PR #48, opened 2024-09-19
List and watch more K8s resource types

evanj80-illumio:
List and watch more resource types to stream to CloudSecure.


cloud-operator/templates/clusterrole.yaml
```diff
@@ -8,65 +8,20 @@ metadata:
...
-  verbs: ["get", "list", "watch"]
-- apiGroups: ["authentication.k8s.io"]
-  resources: ["tokenreviews"]
-  verbs: ["get", "list", "watch"]
-- apiGroups: ["authorization.k8s.io"]
-  resources: ["localsubjectaccessreviews", "selfsubjectaccessreviews", "selfsubjectrulesreviews", "subjectaccessreviews"]
-  verbs: ["get", "list", "watch"]
-- apiGroups: ["autoscaling"]
-  resources: ["horizontalpodautoscalers"]
+  resources: ["replicasets","deployments","statefulsets", "daemonsets"]
```
rlenglet, line 14: Super-nit: consistent spacing.
```suggestion
  resources: ["replicasets", "deployments", "statefulsets", "daemonsets"]
```

cloud-operator/templates/clusterrole.yaml
```diff
@@ -8,65 +8,20 @@ metadata:
...
-- apiGroups: ["metrics.k8s.io"]
-  resources: ["nodes", "pods"]
+  resources: ["jobs", "cronjobs"]
   verbs: ["get", "list", "watch"]
 - apiGroups: ["networking.k8s.io"]
-  resources: ["ingressclasses", "ingresses", "networkpolicies"]
-  verbs: ["get", "list", "watch"]
-- apiGroups: ["node.k8s.io"]
-  resources: ["runtimeclasses"]
+  resources: ["networkpolicies","ingresses", "ingressclasses"]
```
rlenglet, line 20:
```suggestion
  resources: ["networkpolicies", "ingresses", "ingressclasses"]
```

internal/controller/resource_manager.go
```diff
@@ -24,6 +26,30 @@ type ResourceManager struct {
...
+		"jobs":                      "batch",
+		"cronjobs":                  "batch",
+		"services":                  "",
+		"networkpolicies":           "networking.k8s.io",
+		"customresourcedefinitions": "apiextensions.k8s.io",
+		"endpoints":                 "",
+		"gateways":                  "gateway.networking.k8s.io",
+		"httproutes":                "gateway.networking.k8s.io",
+		"gatewayclasses":            "gateway.networking.k8s.io",
+	}
```
rlenglet, line 50: Keep this list in alphabetic order, to make it easier to maintain.

internal/controller/resource_manager.go
```diff
@@ -74,6 +100,9 @@ func (r *ResourceManager) DyanmicListAndWatchResources(ctx context.Context, canc
 		Watch:           true,
 		ResourceVersion: resourceListVersion,
 	}
+	// Prevent us from overwhelming K8 api
+	randomDuration := time.Duration(rand.Intn(5-1+1)+1) * time.Second
+	time.Sleep(randomDuration)
```
rlenglet, line 105: Try using a proper rate limiter like e.g. https://pkg.go.dev/golang.org/x/time/rate instead of a hardcoded sleep.

internal/controller/resource_manager.go
```diff
@@ -116,7 +145,7 @@ func (r *ResourceManager) DynamicListResources(ctx context.Context, resource str
 // watchEvents watches Kubernetes resources and updates cache based on events.
 // Any occurring errors are sent through errChanWatch. The watch stops when ctx is cancelled.
 func (r *ResourceManager) watchEvents(ctx context.Context, resource string, watchOptions metav1.ListOptions, c CacheManager) error {
-	objGVR := schema.GroupVersionResource{Group: "", Version: "v1", Resource: resource}
+	objGVR := schema.GroupVersionResource{Group: resourceAPIGroupMap[resource], Version: "v1", Resource: resource}
```
rlenglet, line 148: Instead of doing that, why not pass both the API group and resource kind as parameters? You could also define a small struct type with that pair of fields, and pass that instead of just the resource kind.

internal/controller/streaming_manager.go
```diff
@@ -48,7 +48,9 @@ type EnvironmentConfig struct {
 	ClusterCreds string
 }
 
-var resourceTypes = [2]string{"pods", "nodes"}
+var resourceTypes = []string{"pods", "nodes", "serviceaccounts", "replicationcontrollers", "replicasets", "deployments", "statefulsets",
```
rlenglet, line 51: This is a recipe for bugs. Don't duplicate the list of supported resource kinds in two different places.
This list should be specified only once.

rlenglet (changes requested):
Some relatively minor comments.

internal/controller/resource_manager.go
```diff
@@ -112,9 +88,9 @@ func (r *ResourceManager) DyanmicListAndWatchResources(ctx context.Context, canc
 }
 
 // DynamicListResources lists a specifed resource dynamically and sends down the current gRPC stream.
-func (r *ResourceManager) DynamicListResources(ctx context.Context, resource string) (string, CacheManager, error) {
+func (r *ResourceManager) DynamicListResources(ctx context.Context, resource string, apiGroup string) (string, CacheManager, error) {
 	cm := CacheManager{cache: make(map[string][32]byte)}
-	objGVR := schema.GroupVersionResource{Group: resourceAPIGroupMap[resource], Version: "v1", Resource: resource}
+	objGVR := schema.GroupVersionResource{Group: apiGroup, Version: "v1", Resource: resource}
```
rlenglet, line 93: Seeing this, I think we shouldn't hardcode the `"v1"` version, just in case.
A cleaner design would be to make `resourceAPIGroupMap` a list of `schema.GroupVersionResource` values. You could then pass around those `schema.GroupVersionResource` values instead of API group + resource type.

Do this in a separate PR if you want.
This is not blocking the PR.

rlenglet (approved):
LGTM
Thanks for addressing my nitpicks. :)

internal/controller/resource_manager.go
evanj80-illumio: Good point. Will make a note of this!

rlenglet (approved):

PR #50, opened 2024-09-24
Enabling multi-platform builds 

evanj80-illumio:
Adding local and github action commands to build multi image builds


rlenglet (approved):
Awesome, thanks!

PR #51, opened 2024-09-24
Refactor stream reconnection logic

aaronIllumio:
Separate stream connection dropouts and issues from cloud-operator-wide reboot.


internal/controller/logger_manager.go
```diff
@@ -147,25 +147,27 @@ func (b *BufferedGrpcWriteSyncer) UpdateClient(client pb.KubernetesInfoService_S
...
-			b.logger.Errorw("Stream terminated",
-				"error", err,
-			)
-			return
-		}
-		switch res.Response.(type) {
-		case *pb.SendLogsResponse_SetLogLevel:
-			newLevel := res.GetSetLogLevel().Level
-			b.updateLogLevel(newLevel)
+		//nolint:gosimple,staticcheck // we need to use select because of future expansion to multiple cases
```
evanj80-illumio, line 153: Seems like staticcheck DID not want to list to your nolint 😆

internal/controller/streaming_manager.go
```diff
@@ -82,42 +84,8 @@ func ServerIsHealthy() bool {
...
-	}
-	return sm, nil
-}
-
-// BootUpStreamAndReconnect creates the clients needed to read K8s resources and
-// also creates all of the objects that help organize and pass info to the goroutines that are listing and watching
-// different resource types. This is also handling the asyc nature of listing resources and properly sending our commit
-// message on intial boot when we have the state of the cluster.
-func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel context.CancelFunc) error {
+// StreamResources handles the resource stream.
```
evanj80-illumio, line 87:
```suggestion
// StreamResources handles the resource stream that streams K8 resources and any events related to them
```

Something like that?

internal/controller/streaming_manager.go
```diff
@@ -192,75 +160,175 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
 		return err
 	}
 	snapshotCompleted.Done()
+
```
evanj80-illumio, line 163:
```suggestion
// Block until context is done, otherwise if watchers return an error and cancel context no errors are returned.
```

internal/controller/streaming_manager.go
```diff
@@ -192,75 +160,175 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+		select {
+		case <-resourceDone:
+		case <-logDone:
+		}
+
+		logger.Warn("All streams have been closed. Rebooting the entire connection.")
+	}
+}
+
+func streamAuth(ctx context.Context, logger *zap.SugaredLogger, envMap EnvironmentConfig) (*grpc.ClientConn, pb.KubernetesInfoServiceClient, error) {
```
evanj80-illumio, line 296: Maybe rename to NewAuthenticatedCon or something? @rlenglet opinions?

evanj80-illumio:
Im not at liberty to review this too much since i party authored it 😆

internal/controller/grpc_logger.go
```diff
@@ -226,20 +228,23 @@ func NewGRPClogger(grpcSyncer *BufferedGrpcWriteSyncer) *zap.SugaredLogger {
 
 	// Add a custom hook to handle logs for grpcSyncer
 	logger = logger.WithOptions(zap.Hooks(func(entry zapcore.Entry) error {
+		// Do not use logging inside the hook to avoid deadlock
```
rlenglet, line 231: Good call. 👍🏻

internal/controller/grpc_logger.go
```diff
@@ -226,20 +228,23 @@ func NewGRPClogger(grpcSyncer *BufferedGrpcWriteSyncer) *zap.SugaredLogger {
...
 		}
+		grpcSyncer.mutex.Unlock()
 
-		// Flush buffer so logs are sent in order to the server
-		grpcSyncer.flush()
-
-		if err := grpcSyncer.sendLog(&entry, nil); err != nil {
-			logger.Error("Error when sending logs to server", zap.Error(err))
+		if shouldBuffer {
 			grpcSyncer.bufferLogEntry(&entry)
```
rlenglet, line 247: Doing this after unlocking doesn't work, because `bufferLogEntry` doesn't lock, so this new logic is not thread-safe.
The lock must be held when calling `bufferLogEntry`.
What is the reason for moving outside of the Lock/Unlock section?

internal/controller/grpc_logger.go
```diff
@@ -226,20 +228,23 @@ func NewGRPClogger(grpcSyncer *BufferedGrpcWriteSyncer) *zap.SugaredLogger {
...
-			return err
 		}
```
rlenglet, line 248: Unlock only after calling `bufferLogEntry` to keep this thread-safe.
```suggestion
		}

		if shouldBuffer {
			grpcSyncer.bufferLogEntry(&entry)
		}

		grpcSyncer.mutex.Unlock()
```

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+func (sm *streamManager) StreamLogs(ctx context.Context, cancel context.CancelFunc) error {
+	errCh := make(chan error, 1)
+
+	go func() {
+		errCh <- sm.bufferedGrpcSyncer.ListenToLogStream()
+	}()
+
+	select {
+	case <-ctx.Done():
+		cancel()
```
rlenglet, line 182: This is unnecessary, because the context gets canceled anyway on every return by the caller, cf. https://github.com/illumio/cloud-operator/pull/51/files#diff-65d4066e2022c168a97a2083474d348fa6d5d1fffb6e8d7378282935ad187940R261
```suggestion
```

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+		return ctx.Err()
+	case err := <-errCh:
+		if err != nil {
+			cancel()
```
rlenglet, line 186: Same here, this is unnecessary.
```suggestion
```

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
 		return err
 	}
 	snapshotCompleted.Done()
+
+	<-ctx.Done()
+	return ctx.Err()
+}
+
+// StreamLogs handles the log stream.
+func (sm *streamManager) StreamLogs(ctx context.Context, cancel context.CancelFunc) error {
```
rlenglet, line 173: `cancel` is not needed, so remove it from the signature:
```suggestion
func (sm *streamManager) StreamLogs(ctx context.Context) error {
```

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+			cancel()
+			return err
+		}
+	}
+	return nil
+}
+
+// StreamLogs handles the log stream.
+func (sm *streamManager) StreamCiliumNetworkFlows(ctx context.Context, cancel context.CancelFunc, ciliumNamespace string) error {
+	defer cancel()
```
rlenglet, line 195: `cancel` is unnecessary in this method.
```suggestion
func (sm *streamManager) StreamCiliumNetworkFlows(ctx context.Context, ciliumNamespace string) error {
```

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+
+	sendCiliumNetworkFlowsStream, err := sm.streamClient.client.SendKubernetesNetworkFlows(ciliumCtx)
+	if err != nil {
+		logger.Errorw("Failed to connect to server", "error", err)
+		return err
+	}
+
+	sm.streamClient.ciliumNetworkFlowsStream = sendCiliumNetworkFlowsStream
+
+	err = sm.StreamCiliumNetworkFlows(ciliumCtx, ciliumCancel, sm.streamClient.ciliumNamespace)
```
rlenglet, line 227:
```suggestion
	err = sm.StreamCiliumNetworkFlows(ciliumCtx, sm.streamClient.ciliumNamespace)
```

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+	SendLogsStream, err := sm.streamClient.client.SendLogs(logCtx)
+	if err != nil {
+		logger.Errorw("Failed to connect to server", "error", err)
+		return err
+	}
+
+	sm.streamClient.logStream = SendLogsStream
+	sm.bufferedGrpcSyncer.UpdateClient(sm.streamClient.logStream, sm.streamClient.conn)
+
+	err = sm.StreamLogs(logCtx, logCancel)
```
rlenglet, line 272:
```suggestion
	err = sm.StreamLogs(logCtx)
```

rlenglet (changes requested):
First pass of review.
Summary: `StreamResources` does need to get the `cancel` parameter, but `StreamLogs` and `StreamCiliumNetworkFlows`. Removing `cancel` from those methods will simplify the design.

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+	if err != nil {
+		logger.Errorw("Failed to connect to server", "error", err)
+		return err
+	}
+
+	sm.streamClient.ciliumNetworkFlowsStream = sendCiliumNetworkFlowsStream
+
+	err = sm.StreamCiliumNetworkFlows(ciliumCtx, ciliumCancel, sm.streamClient.ciliumNamespace)
+	if err != nil {
+		logger.Errorw("Failed to bootup and Cilium network flows", "error", err)
```
rlenglet, line 229:
```suggestion
		logger.Errorw("Failed to bootup and stream Cilium network flows", "error", err)
```

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+
+	return nil
+}
+
+// Generic function to manage any stream with backoff and reconnection logic.
+func manageStream(logger *zap.SugaredLogger, connectAndStream func(*zap.SugaredLogger, *streamManager) error, sm *streamManager, done chan struct{}) {
 	var backoff = 1 * time.Second
-	authn := Authenticator{Logger: logger}
 	max := big.NewInt(3)
+	rebootCounter := 0
```
rlenglet, line 285:
```suggestion
	rebootCount := 0
```

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
-		if err != nil {
-			logger.Errorw("Could not read K8s credentials", "error", err)
-		}
-		if clientID == "" && clientSecret == "" {
-			OnboardingCredentials, err := authn.GetOnboardingCredentials(ctx, envMap.OnboardingClientId, envMap.OnboardingClientSecret)
-			if err != nil {
-				logger.Errorw("Failed to get onboarding credentials", "error", err)
-				continue
+			if rebootCounter >= 10 {
+				done <- struct{}{}
```
rlenglet, line 291: Usually, e.g. like for the `Context.Done()` channel, you should close the channel instead.
```suggestion
				close(done)
```

Right now, nothing is closing the channel, so those channels are leaking.

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
-				logger.Errorw("Failed to write secret to Kubernetes", "error", err)
-			}
-			clientID, clientSecret, err = authn.ReadCredentialsK8sSecrets(ctx, envMap.ClusterCreds)
-			if err != nil {
-				logger.Errorw("Could not read K8s credentials", "error", err)
+				logger.Errorw("Could not generate a random int", "error", err)
 				continue
 			}
+			result := randomInt.Int64()
+			sleep := 1*time.Second + backoff + time.Duration(result)*time.Millisecond // Add randomness
```
rlenglet, line 301: This jitter is only up to 3 milliseconds?! That's way too small.
It should be like 10% or 20% of the delay.
```suggestion
			sleep := backoff
			jitterPct := rand.Float64() * 0.20 // [0, 20%)
			sleep = time.Duration(float64(sleep) * (1. - jitterPct))
			if sleep < 1*time.Second {
				sleep = 1 * time.Second
			}
```

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+		logger.Errorw("Failed to bootup and stream logs", "error", err)
+		return err
+	}
+
+	return nil
+}
+
+// Generic function to manage any stream with backoff and reconnection logic.
+func manageStream(logger *zap.SugaredLogger, connectAndStream func(*zap.SugaredLogger, *streamManager) error, sm *streamManager, done chan struct{}) {
 	var backoff = 1 * time.Second
```
rlenglet, line 283: As an alternative to returning after 10 retries, you could make this function loop indefinitely and:
* Decrease the backoff automatically e.g. after some time without any errors, or after a successful connection. Otherwise, if the pod has any temporary connectivity issues, the backoff would remain forever high until the pod is restarted.
* Put a cap on the backoff, so it would not exceed a maximum. E.g. 30s, 1m, whatever.

Just a thought...

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
-		go bufferedGrpcSyncer.ListenToLogStream()
+		resourceDone := make(chan struct{})
+		logDone := make(chan struct{})
+		ciliumDone := make(chan struct{})
+		sm.bufferedGrpcSyncer.done = logDone
+
+		go manageStream(logger, connectAndStreamResources, sm, resourceDone)
+		go manageStream(logger, connectAndStreamLogs, sm, logDone)
+
+		// Need to block this until resources are done being committed. Using waitgroups? Then we need to pass that in? Or pullout the snapschat complete watigroup to here?
```
rlenglet, line 339: No, no need to wait.
```suggestion
```

internal/controller/streams.go
```diff
@@ -197,105 +164,225 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+		go manageStream(logger, connectAndStreamCiliumNetworkFlows, sm, ciliumDone)
+
+		<-ciliumDone
+		<-resourceDone
+		<-logDone
+
+		logger.Warn("All streams have been closed. Rebooting the entire connection.")
```
rlenglet, line 346: That logic doesn't work if we have issues with just one type of stream, as it could well be that the logs stream keeps terminating, we retry 10 times, and then `manageStream` returns for that one stream. But if the other streams were healthy, then we would never retry again to reopen the logs stream until we restart the pod.

It would make more sense to me to have `manageStream` loop indefinitely, and to cap the backoff delay, as I commented above.

You could still have `manageStream` return an error in the channels. In this case, it would now indicate a several failure, and that should cause restarting all streams.
Like:
```suggestion
		select {
		case <-ciliumDone:
		case <-resourceDone:
		case <-logDone:
		}

		logger.Warn("One or more streams have been closed; reopening the connection")
```

rlenglet (changes requested):

internal/controller/streaming_manager.go
rlenglet: Yeah, that sounds better.

internal/controller/streams.go
```diff
@@ -17,21 +18,24 @@ import (
...
-	conn                      *grpc.ClientConn
-	streamKubernetesResources pb.KubernetesInfoService_SendKubernetesResourcesClient
-	streamKubernetesFlows     pb.KubernetesInfoService_SendKubernetesNetworkFlowsClient
-	logStream                 pb.KubernetesInfoService_SendLogsClient
+	ciliumNamespace          string
+	ciliumNetworkFlowsStream pb.KubernetesInfoService_SendKubernetesNetworkFlowsClient
+	conn                     *grpc.ClientConn
+	client                   pb.KubernetesInfoServiceClient
+	logStream                pb.KubernetesInfoService_SendLogsClient
+	resourceStream           pb.KubernetesInfoService_SendKubernetesResourcesClient
```
rlenglet, line 26: Are those ever set to nil again when a stream is closed?
Why not? Why keep a reference to a closed stream?

internal/controller/streams.go
```diff
@@ -197,105 +165,269 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
 	max := big.NewInt(3)
+
+	resetTimer := time.NewTimer(resetPeriod)
 	for {
-		// Generate a random number
-		randomInt, err := rand.Int(rand.Reader, max)
-		if err != nil {
-			logger.Errorw("Could not generate a random int", "error", err)
+		select {
+		case <-resetTimer.C:
```
rlenglet, line 299: This case will almost never be selected because of the `default` below.
In a `select`, if all the `case`s are blocking, the `default` is selected.

internal/controller/streams.go
```diff
@@ -197,105 +165,269 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+				consecutiveFailures = 0
+				backoff = initialBackoff
 			}
 		}
-		conn, err := SetUpOAuthConnection(ctx, logger, envMap.TokenEndpoint, envMap.TlsSkipVerify, clientID, clientSecret)
+	}
+}
+
+// ExponentialStreamConnect will continue to reboot and restart the main operations within the operator if any disconnects or errors occur.
+func ExponentialStreamConnect(ctx context.Context, logger *zap.SugaredLogger, envMap EnvironmentConfig, bufferedGrpcSyncer *BufferedGrpcWriteSyncer) {
```
rlenglet, line 355: What is `ctx` used for?

rlenglet (changes requested):

internal/controller/streams.go
aaronIllumio: Ideally the log and resource stream should never be closed right? Both should be always connected or attempting to connect. I can set the ciliumStream to nil though

internal/controller/streams.go
aaronIllumio: It is used in NewAuthenticatedCon to get Oauth Token and grpc connection

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+
+				randomInt, err := rand.Int(rand.Reader, max)
+				if err != nil {
+					logger.Errorw("Could not generate a random int", "error", err)
+					sleepTimer.Reset(backoff) // Reset sleep timer before continuing loop
+					continue
+				}
+
+				jitterPct, _ := randomInt.Float64()
+				jitterPct = jitterPct * maxJitterPct
```
rlenglet, line 334:
```suggestion
				jitterPct := rand.Float64() * maxJitterPct // [0, maxJitterPct)
```

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+	if err != nil {
+		logger.Errorw("Failed to bootup and stream logs", "error", err)
+		return err
+	}
+
+	return nil
+}
+
+// Generic function to manage any stream with backoff and reconnection logic.
+func manageStream(logger *zap.SugaredLogger, connectAndStream func(*zap.SugaredLogger, *streamManager) error, sm *streamManager, done chan struct{}) {
```
rlenglet, line 286: You close `done` before every `return`.
You can as well do that in a `defer`.
```suggestion
func manageStream(logger *zap.SugaredLogger, connectAndStream func(*zap.SugaredLogger, *streamManager) error, sm *streamManager, done chan struct{}) {
	defer close(done)
```

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+			backoff = initialBackoff
+			resetTimer.Reset(resetPeriod)
+		case <-sleepTimer.C:
+			err := connectAndStream(logger, sm)
 			if err != nil {
-				logger.Errorw("Failed to get onboarding credentials", "error", err)
-				continue
+				if errors.Is(err, ErrStopRetries) {
+					logger.Info("Stopping retries for this stream as instructed.")
+					close(done)
```
rlenglet, line 315: Do this in a `defer`.
```suggestion
```

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+					return
+				}
+				logger.Errorw("Failed to establish stream connection; will retry", "error", err)
+				consecutiveFailures++
+
+				if consecutiveFailures >= severeErrorThreshold {
+					close(done)
```
rlenglet, line 322: Do this in a `defer`.
```suggestion
```

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
-				continue
+				if errors.Is(err, ErrStopRetries) {
+					logger.Info("Stopping retries for this stream as instructed.")
+					close(done)
+					return
+				}
+				logger.Errorw("Failed to establish stream connection; will retry", "error", err)
+				consecutiveFailures++
+
+				if consecutiveFailures >= severeErrorThreshold {
```
rlenglet, line 321:
```suggestion
				switch {
				case consecutiveFailures == 1:
					resetTimer.Reset(resetPeriod)
				case consecutiveFailures >= severeErrorThreshold {
```

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+
+				backoff *= 2
+				if backoff > maxBackoff {
+					backoff = maxBackoff
+				}
+			} else {
+				consecutiveFailures = 0
+				backoff = initialBackoff
+				// Reset sleep timer back to initial backoff interval
+				sleepTimer.Reset(initialBackoff)
```
rlenglet, line 355:
```suggestion
				resetTimer.Reset(resetPeriod)
				sleepTimer.Reset(initialBackoff)
```

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+
+			sm := &streamManager{
+				streamClient:       streamClient,
+				logger:             logger,
+				bufferedGrpcSyncer: bufferedGrpcSyncer,
 			}
+
+			resourceDone := make(chan struct{})
+			logDone := make(chan struct{})
+			ciliumDone := make(chan struct{})
```
rlenglet, line 390: You're leaking this channel inc ase Cilium flows are disabled.
```suggestion
			var chan struct{} ciliumDone
```

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+			sm.bufferedGrpcSyncer.done = logDone
+
+			go manageStream(logger, connectAndStreamResources, sm, resourceDone)
+			go manageStream(logger, connectAndStreamLogs, sm, logDone)
+			// Only start network flows stream if not disabled
+			if !sm.streamClient.disableNetworkFlows {
+				go manageStream(logger, connectAndStreamCiliumNetworkFlows, sm, ciliumDone)
```
rlenglet, line 397: Only create the channel if you need it.
```suggestion
				ciliumDone = make(chan struct{})
				go manageStream(logger, connectAndStreamCiliumNetworkFlows, sm, ciliumDone)
```

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+			} else {
+				ciliumDone = nil
+			}
+
+			select {
+			case <-ciliumDone:
+				if sm.streamClient.disableNetworkFlows {
+					ciliumDone = nil // Prevent further processing if disabled
+				}
```
rlenglet, line 406: That's not needed. The next iteration will reset it correctly.
```suggestion
```

rlenglet (changes requested):

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+			case <-resourceDone:
+			case <-logDone:
+			}
+
+			logger.Warn("One or more streams have been closed. Closing and reopening the connection to CloudSecure.")
```
rlenglet, line 411: Nit.
```suggestion
			logger.Warn("One or more streams have been closed; closing and reopening the connection to CloudSecure")
```

internal/controller/streams.go
```diff
@@ -197,105 +167,289 @@ func (sm *streamManager) BootUpStreamAndReconnect(ctx context.Context, cancel co
...
+			sm.streamClient.conn.Close()
 		}
-		conn, err := SetUpOAuthConnection(ctx, logger, envMap.TokenEndpoint, envMap.TlsSkipVerify, clientID, clientSecret)
+	}
+}
+
+// streamAuth handles getting a valid token and creating a connection
+func NewAuthenticatedCon(ctx context.Context, logger *zap.SugaredLogger, envMap EnvironmentConfig) (*grpc.ClientConn, pb.KubernetesInfoServiceClient, error) {
```
rlenglet, line 418:
```suggestion
// NewAuthenticatedConnection gets a valid token and creats a connection to CloudSecure.
func NewAuthenticatedConnection(ctx context.Context, logger *zap.SugaredLogger, envMap EnvironmentConfig) (*grpc.ClientConn, pb.KubernetesInfoServiceClient, error) {
```

rlenglet (changes requested):

rlenglet (dismissed):

rlenglet (approved):

PR #56, opened 2024-10-02
Include IP address of Pods in k8s resources of kind Pod

evanj80-illumio:
Adding pod ip addresses to object metadata specifically for pod resources.


.github/workflows/go-security-checks.yml
```diff
@@ -35,9 +35,6 @@ jobs:
       - name: Install golangci-lint
         uses: golangci/golangci-lint-action@v6
         with:
-          # Needed to address this issue https://github.com/golangci/golangci-lint-action/issues/244
```
evanj80-illumio, line 38: Linter gave an error so I think this problem has been resolved. Also checked issue and it seems like its resolved.

internal/controller/k8s_resources.go
```diff
@@ -126,6 +129,33 @@ func convertMetaObjectToMetadata(obj metav1.ObjectMeta, resource string) *pb.Kub
 	return objMetadata
 }
 
+// getPodIPAddresses uses a pod name and namespace to grab the hostIP addresses within the podStatus
+func getPodIPAddresses(ctx context.Context, logger *zap.SugaredLogger, podName string, namespace string) ([]v1.HostIP, error) {
```
rlenglet, line 133: Why are you re-querying the Pod?
You already receive the full Pod object from the watcher.
No need to re-query data you already have.

internal/controller/resource_manager.go
```diff
@@ -102,7 +102,15 @@ func (r *ResourceManager) DynamicListResources(ctx context.Context, resource str
 	}
 	for _, obj := range objs {
 		metadataObj := convertMetaObjectToMetadata(obj, resource)
-		err := sendObjectMetaData(r.streamManager, metadataObj)
+		if resource == "pods" {
+			hostIPs, err := getPodIPAddresses(ctx, r.logger, obj.GetName(), obj.GetNamespace())
```
rlenglet, line 106: `resource` should already contains that data. Don't re-query it.
Modify `convertMetaObjectToMetadata` to add the kind-specific data into the returned protobuf.
And rename `convertMetaObjectToMetadata` into `convertMetaObjectToInventoryData` or something like that to clarify its expanded scope.

internal/controller/resource_manager.go
```diff
@@ -147,7 +155,14 @@ func (r *ResourceManager) watchEvents(ctx context.Context, resource string, apiG
 			return err
 		}
 		metadataObj := convertMetaObjectToMetadata(*convertedData, resource)
-
+		if resource == "pods" {
```
rlenglet, line 158: Same here, move that logic into `convertMetaObjectToMetadata`.
You shouldn't have to re-query data you already have just above (`resource`).

rlenglet (changes requested):

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -87,11 +97,11 @@ message KubernetesResourceMutation {
   // The documented types
   oneof mutation {
```
rlenglet, line 98: Do we send the `kind_specific` data for every kind of mutation, or only for create/update?
I think we can omit that for `delete_resource` mutations, as a small optimization.
Clarify that in the comments.

This is really a nit. I'm OK with ignoring this.

internal/controller/k8s_resources.go
```diff
@@ -79,6 +83,7 @@ func convertObjectToMetadata(obj metav1.Object) metav1.ObjectMeta {
 // getObjectMetadataFromRuntimeObject safely extracts metadata from any Kubernetes runtime.Object.
 // It returns a pointer to a metav1.ObjectMeta structure if successful, along with any error encountered.
 func getObjectMetadataFromRuntimeObject(obj runtime.Object) (*metav1.ObjectMeta, error) {
+	fmt.Println(obj)
```
rlenglet, line 86: Remove. And thanks for checking that this doesn't contain anything beyond the metadata!
```suggestion
```

internal/controller/streaming_helper.go
```diff
@@ -29,8 +29,8 @@ func sendCiliumFlow(sm *streamManager, flow *pb.CiliumFlow) error {
 	return nil
 }
 
-// streamMutationObjectMetaData sends a resource mutation message into the given stream.
-func streamMutationObjectMetaData(sm *streamManager, metadata *pb.KubernetesObjectMetadata, eventType watch.EventType) error {
+// streamMutationObjectMetadata sends a resource mutation message into the given stream.
+func streamMutationObjectMetadata(sm *streamManager, metadata *pb.KubernetesObjectData, eventType watch.EventType) error {
```
rlenglet, line 33:
```suggestion
// streamMutationObjectData sends a resource mutation message into the given stream.
func streamMutationObjectData(sm *streamManager, metadata *pb.KubernetesObjectData, eventType watch.EventType) error {
```

internal/controller/streaming_helper.go
```diff
@@ -7,10 +7,10 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 )
 
-// sendObjectMetaData sends a KubernetesMetadata to CloudSecure into the given stream.
+// sendObjectMetadata sends a KubernetesObjectData to CloudSecure into the given stream.
```
rlenglet, line 10:
```suggestion
// sendObjectData sends a KubernetesObjectData to CloudSecure into the given stream.
```

internal/controller/streaming_helper.go
```diff
@@ -7,10 +7,10 @@ import (
...
 // Its used for the intial boot up of the operator so that is can stream everything currently in the cluster.
-func sendObjectMetaData(sm *streamManager, metadata *pb.KubernetesObjectMetadata) error {
-	if err := sm.streamClient.streamKubernetesResources.Send(&pb.SendKubernetesResourcesRequest{Request: &pb.SendKubernetesResourcesRequest_ResourceMetadata{ResourceMetadata: metadata}}); err != nil {
+func sendObjectMetadata(sm *streamManager, metadata *pb.KubernetesObjectData) error {
```
rlenglet, line 12:
```suggestion
func sendObjectData(sm *streamManager, metadata *pb.KubernetesObjectData) error {
```

rlenglet (changes requested):
Looks good. Only some minor nits.
Thanks!

internal/controller/resource_manager.go
rlenglet: The meta point here was to move the logic into `convertMetaObjectToMetadata` so you don't have copy-and-paste the `if`/`switch` below multiple times.

evanj80-illumio: Ohh yeah I get what you are saying! Will do.

rlenglet (approved):
LGTM thank you!

PR #64, opened 2024-10-17
Fixing proto definitions

evanj80-illumio:
-> Fixing typo located in protofile
-> issue with the connection always being closed when trying to reconnect


aaronIllumio (approved):
LPTM

PR #70, opened 2024-11-05
Adding Falco listener and Falco event streamer to cloud-operator

evanj80-illumio:
Went through multiple iterations of how to implement listening to falco events. Concluded that forcing falco to send http events to cloud-operator makes the most sense and is the easiest way to source all events from multple agents. 

This PR builds support for listening, parsing and streaming those events to CloudSecure.


cmd/main.go
evanj80-illumio, line 1: Adding here to be more consistent with how I create new http listener for falco. I think this looks nicer and makes more sense.

internal/controller/resource_manager.go
evanj80-illumio, line 1: Cleanup that was missed from testing PR

evanj80-illumio, line 1: Clean up of file from accidental print statement from test

internal/controller/flow_falco.go
```diff
@@ -0,0 +1,87 @@
...
+				return
+			}
+			eventChan <- podInfo
+		}
+		w.WriteHeader(http.StatusOK)
+	}
+}
+
+// filterIllumioTraffic filters out events related to Illumio network traffic.
+func filterIllumioTraffic(body string) bool {
```
aaronIllumio, line 85: Why do we have a specific function to do only illumio traffic

internal/controller/streams.go
```diff
@@ -380,15 +442,18 @@ func ConnectStreams(ctx context.Context, logger *zap.SugaredLogger, envMap Envir
...
 
 			go manageStream(logger, connectAndStreamResources, sm, resourceDone)
 			go manageStream(logger, connectAndStreamLogs, sm, logDone)
+			go manageStream(logger, connectAndStreamFalcoNetworkFlows, sm, falcoDone)
 			// Only start network flows stream if not disabled
-			if !sm.streamClient.disableNetworkFlows {
+			if !sm.streamClient.disableNetworkFlowsCilium {
 				ciliumDone = make(chan struct{})
 				go manageStream(logger, connectAndStreamCiliumNetworkFlows, sm, ciliumDone)
+
```
aaronIllumio, line 456: please add the falcoDone channel to the select at line 464 for reconnect logic

aaronIllumio (changes requested):
Looks pretty good left some comments. Can you also test this the reconnect logic by connecting to dev and enabling/disabling the cluster a couple of times thanks

internal/controller/flow_falco.go
evanj80-illumio: Its for filtering out falco rules, This is filtering based on the string of the output value contained within the JSON body of the http request that falco sends us on a per request basis.

evanj80-illumio: Customers will use a custom set of rules we give them and we will make sure they always contain that string so we can filter easily.

internal/controller/streams.go
evanj80-illumio: Good catch

internal/controller/streams.go
```diff
@@ -17,13 +19,14 @@ import (
...
-	client              pb.KubernetesInfoServiceClient
-	disableNetworkFlows bool
-	logStream           pb.KubernetesInfoService_SendLogsClient
-	networkFlowsStream  pb.KubernetesInfoService_SendKubernetesNetworkFlowsClient
-	resourceStream      pb.KubernetesInfoService_SendKubernetesResourcesClient
+	ciliumNamespace           string
+	conn                      *grpc.ClientConn
+	client                    pb.KubernetesInfoServiceClient
+	disableNetworkFlowsCilium bool
+	falcoChan                 chan string
```
aaronIllumio, line 26: can you name this falcoEventChan to be more clear with what it actually is

internal/controller/streams_helper.go
```diff
@@ -29,6 +29,21 @@ func sendObjectData(sm *streamManager, metadata *pb.KubernetesObjectData) error
 	return sendToResourceStream(sm.logger, sm.streamClient.resourceStream, request)
 }
 
+// sendFalcoFlow sends a FalcoFlow to the networkFlowsStream
+func sendFalcoFlow(sm *streamManager, flow *pb.FalcoFlow) error {
```
aaronIllumio, line 33: Maybe if possible make a generalized flows send function since we are using the same stream for all flows. This might be helpful with all of the future cni plugins we will use and not have a lot fo duplicate code. Thanks!

aaronIllumio (changes requested):
Looks pretty good! Just a couple of questions and changes

internal/controller/streams.go
```diff
@@ -236,6 +272,26 @@ func connectAndStreamCiliumNetworkFlows(logger *zap.SugaredLogger, sm *streamMan
 	return nil
 }
 
+// connectAndStreamFalcoNetworkFlows creates networkFlowsStream client and begins the streaming of network flows.
+func connectAndStreamFalcoNetworkFlows(logger *zap.SugaredLogger, sm *streamManager) error {
+	falcoCtx, falcoCancel := context.WithCancel(context.Background())
+	defer falcoCancel()
+	sendFalcoNetworkFlows, err := sm.streamClient.client.SendKubernetesNetworkFlows(falcoCtx)
```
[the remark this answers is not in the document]
evanj80-illumio: Yeah it is possible but I think we are going to have an advisory to not let that happen. If we have a customer who will have both on their cluster it still should be fine because in order to have falco talk to our operator they'll need to specifically configure it with out service endpoint. (which cannot accidentally happen)

aaronIllumio (dismissed):
Can you make sure you cover all edges cases in ur tests

Overall LGTM

internal/controller/flow_falco.go
```diff
@@ -0,0 +1,160 @@
...
+		return &pb.IP{Source: source, Destination: destination, IpVersion: pb.IPVersion_IP_VERSION_IPV4}, nil
+	} else if ipVersion == "ipv6" {
+		return &pb.IP{Source: source, Destination: destination, IpVersion: pb.IPVersion_IP_VERSION_IPV6}, nil
+	}
+	return &pb.IP{Source: source, Destination: destination, IpVersion: pb.IPVersion_IP_VERSION_IP_NOT_USED_UNSPECIFIED}, nil
+}
+
+// CreateLayer4Message converts event protocol and ports to a Layer4 proto message
+func CreateLayer4Message(proto string, srcPort, dstPort uint32, ipVersion string) (*pb.Layer4, error) {
+	switch proto {
```
rlenglet, line 122: Have you confirmed that Falco supports ICMP?
https://github.com/falcosecurity/libs/blob/230ddfb91d075241391697900ab4281aa67fe108/userspace/libsinsp/utils.cpp#L1026-L1052

internal/controller/flow_falco.go
```diff
@@ -0,0 +1,160 @@
...
+			}, nil
+		} else if ipVersion == "ipv6" {
+			return &pb.Layer4{
+				Protocol: &pb.Layer4_Icmpv6{
+					Icmpv6: &pb.ICMPv6{},
+				},
+			}, nil
+		}
+	default:
+		return nil, fmt.Errorf("unknown protocol: %s", proto)
```
rlenglet, line 157: I'd rather silently ignore. Don't return an error for that. Because it seems like we'lll also get UDS flows, RAW socket flows, etc.?
https://github.com/falcosecurity/libs/blob/230ddfb91d075241391697900ab4281aa67fe108/userspace/libsinsp/parsers.cpp#L2564-L2565
We should just ignore those.

internal/controller/flow_falco.go
```diff
@@ -0,0 +1,160 @@
...
+					SourcePort:      srcPort,
+					DestinationPort: dstPort,
+				},
+			},
+		}, nil
+	case "icmp":
+		if ipVersion == "ipv4" {
+			return &pb.Layer4{
+				Protocol: &pb.Layer4_Icmpv4{
+					Icmpv4: &pb.ICMPv4{},
```
rlenglet, line 146: Have you confirmed that we don't get the ICMP type and code as the src port and dst port, by any chance?

internal/controller/streams.go
```diff
@@ -203,15 +206,48 @@ func (sm *streamManager) StreamCiliumNetworkFlows(ctx context.Context, ciliumNam
...
 
-// connectAndStreamCiliumNetworkFlows creates ciliumNetworkFlows client and begins the streaming of network flows.
+// StreamFalcoNetworkFlows handles the falco network flow stream.
+func (sm *streamManager) StreamFalcoNetworkFlows(ctx context.Context) error {
+	var falcoFlow string
+	for {
+		falcoFlow = <-sm.streamClient.falcoEventChan
+		if filterIllumioTraffic(falcoFlow) {
+			// Extract the relevant part of the output string
+			re := regexp.MustCompile(`\((.*?)\)`)
```
rlenglet, line 224: Move this into a global variable, and call the `MustCompile` code into `func init() {...}` so this is compiled only once, as this call is expensive.

internal/controller/streams.go
```diff
@@ -203,15 +206,48 @@ func (sm *streamManager) StreamCiliumNetworkFlows(ctx context.Context, ciliumNam
...
 	}
 	return nil
 }
 
-// connectAndStreamCiliumNetworkFlows creates ciliumNetworkFlows client and begins the streaming of network flows.
+// StreamFalcoNetworkFlows handles the falco network flow stream.
+func (sm *streamManager) StreamFalcoNetworkFlows(ctx context.Context) error {
+	var falcoFlow string
+	for {
+		falcoFlow = <-sm.streamClient.falcoEventChan
```
rlenglet, line 221: This variable doesn't need to be defined out of the loop as it's never accessed outside of the loop, and doesn't need to be preserved across iterations.
```suggestion
	for {
		falcoFlow := <-sm.streamClient.falcoEventChan
```

internal/controller/streams.go
```diff
@@ -203,15 +206,48 @@ func (sm *streamManager) StreamCiliumNetworkFlows(ctx context.Context, ciliumNam
...
+				return nil
+			}
+
+			podInfo, err := parsePodNetworkInfo(match[1])
+			if err != nil {
+				return nil
+			}
+			convertedFalcoFlow, err := convertFalcoEventToFlow(podInfo)
+			if err != nil {
+				sm.logger.Errorw("Failed to convert FalcoEvent to flows", "error", err)
```
rlenglet, line 236:
```suggestion
				sm.logger.Errorw("Failed to parse Falco event into flow", "error", err)
```

internal/controller/streams.go
```diff
@@ -203,15 +206,48 @@ func (sm *streamManager) StreamCiliumNetworkFlows(ctx context.Context, ciliumNam
...
+			// Extract the relevant part of the output string
+			re := regexp.MustCompile(`\((.*?)\)`)
+			match := re.FindStringSubmatch(falcoFlow)
+			if len(match) < 2 {
+				return nil
+			}
+
+			podInfo, err := parsePodNetworkInfo(match[1])
+			if err != nil {
+				return nil
```
rlenglet, line 232: Add a comment, as this is surprising at first glance.
```suggestion
				// If the event can't be parsed, consider that it's not a flow event and just ignore it.
				return nil
```

internal/controller/streams.go
```diff
@@ -236,6 +272,26 @@ func connectAndStreamCiliumNetworkFlows(logger *zap.SugaredLogger, sm *streamMan
...
+// connectAndStreamFalcoNetworkFlows creates networkFlowsStream client and begins the streaming of network flows.
+func connectAndStreamFalcoNetworkFlows(logger *zap.SugaredLogger, sm *streamManager) error {
+	falcoCtx, falcoCancel := context.WithCancel(context.Background())
+	defer falcoCancel()
+	sendFalcoNetworkFlows, err := sm.streamClient.client.SendKubernetesNetworkFlows(falcoCtx)
+	if err != nil {
+		logger.Errorw("Failed to connect to server", "error", err)
+		return err
+	}
+	sm.streamClient.networkFlowsStream = sendFalcoNetworkFlows
```
rlenglet, line 284:
```suggestion

	sm.streamClient.networkFlowsStream = sendFalcoNetworkFlows
```

rlenglet, line 284: How do you make sure that only one of `connectAndStreamCiliumNetworkFlows` or `connectAndStreamFalcoNetworkFlows` gets called and sets this `networkFlowsStream ` field?

internal/controller/streams.go
```diff
@@ -353,6 +409,27 @@ func manageStream(logger *zap.SugaredLogger, connectAndStream func(*zap.SugaredL
 
 // ConnectStreams will continue to reboot and restart the main operations within the operator if any disconnects or errors occur.
 func ConnectStreams(ctx context.Context, logger *zap.SugaredLogger, envMap EnvironmentConfig, bufferedGrpcSyncer *BufferedGrpcWriteSyncer) {
+	// Falco channels communicate news events between http server and our network flows strea,
+	falcoEventChan := make(chan string)
+	http.HandleFunc("/", NewFalcoEventHandler(falcoEventChan))
+	falcoEvent := &http.Server{Addr: ":5000"}
```
rlenglet, line 415: At least define a constant for this port.

internal/controller/streams.go
rlenglet, line 284: More generally, I still don't see the point of defining and setting this `networkFlowsStream` field, as the stream is only used by the one call to `StreamFalcoNetworkFlows` below. So it is not really part of the state. It's only used in the context of this one execution stack.
This should rather be passed as a parameter to `StreamFalcoNetworkFlows` (resp. `StreamCiliumNetworkFlows`).

internal/controller/streams.go
```diff
@@ -353,6 +409,27 @@ func manageStream(logger *zap.SugaredLogger, connectAndStream func(*zap.SugaredL
...
+	go func() {
+		for {
+			func() {
+				defer func() {
+					if r := recover(); r != nil {
+						logger.Infow("Recovered from panic. Restarting server...", "error", r)
+					}
+				}()
+				if err := falcoEvent.ListenAndServe(); err != nil {
+					logger.Fatal("Falco server failed", zap.Error(err))
```
rlenglet, line 426: `logger.Fatal` just exits the process. So what's the point of this loop and the `recover()`, etc.?
https://github.com/uber-go/zap/blob/966925341b9f43d835469b153e907cc9e1003c17/logger.go#L292-L293

rlenglet (changes requested):
Looks pretty good, thanks!
Some relatively minor comments.

cloud-operator/templates/cloud-operator-service.yaml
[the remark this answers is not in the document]
evanj80-illumio: Was not planning on adding a deployment of falco to our helm charts. Was going to create a guide ( which I have a rough version of on confluence) that customers can use to install falco using our config and rules.

internal/controller/flow_falco.go
```diff
@@ -0,0 +1,160 @@
...
+		var body struct {
+			Output string `json:"output"`
+		}
+
+		if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		eventChan <- body.Output
+		w.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
evanj80-illumio: Great idea here.

internal/controller/flow_falco.go
evanj80-illumio: Seems like their docs may be different than implementation you linked.... https://falco.org/docs/reference/rules/supported-fields/ under fd.l4proto it states icmp is supported type.

internal/controller/flow_falco.go
```diff
@@ -0,0 +1,160 @@
...
+	// DstPort is the destination port number involved in the network event.
+	DstPort string `json:"dstport"`
+	// Proto is the protocol used in the network event (e.g., TCP, UDP).
+	Proto string `json:"proto"`
+	// IpVersion is the version used in the network event (e.g. ipv4, ipv6).
+	IpVersion string `json:"prototype"`
+}
+
+// parsePodNetworkInfo parses the input string to extract network information into a FalcoEvent struct.
+func parsePodNetworkInfo(input string) (FalcoEvent, error) {
```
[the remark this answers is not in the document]
evanj80-illumio: Yeah just looked over the logic, made it like that in case that logic needed to be separate down the line, but not needed to be like that now.

internal/controller/streams.go
evanj80-illumio: Yeah looking back (after 2 weeks) ill simplify this

internal/controller/flow_falco.go
evanj80-illumio: Their docs outline that we only get src/dst ports with TCP and UDP

internal/controller/streams.go
evanj80-illumio: my thought process is we will always have falco enabled UNLESS Cilium is used. Then if cilium is enabled that falco server will never be used thus its okay for us to override that networkFlowsStream. Now if we are worried there will be some customer environments that may have both, than Cilium should still take precedence (which is what is happening now). Now if they are both set up by a customer, they will not have the illumio specific rules that our falco events are being filtered on thus those events should never even be sent down stream.

cmd/main.go
```diff
@@ -103,14 +103,18 @@ func main() {
...
 
 	go func() {
-		errChan <- http.ListenAndServe(":8080", nil)
+		errChan <- healthChecker.ListenAndServe()
 		err := <-errChan
 		logger.Fatal("healthz check server failed", zap.Error(err))
 	}()
 
 	ctx := context.Background()
+	controller.InitRegex()
```
rlenglet, line 117: Use an `init` function, cf. https://go.dev/doc/effective_go#init

rlenglet (changes requested):

internal/controller/streams.go
```diff
@@ -380,21 +460,26 @@ func ConnectStreams(ctx context.Context, logger *zap.SugaredLogger, envMap Envir
...
 			// Only start network flows stream if not disabled
-			if !sm.streamClient.disableNetworkFlows {
+			if !sm.streamClient.disableNetworkFlowsCilium {
 				ciliumDone = make(chan struct{})
 				go manageStream(logger, connectAndStreamCiliumNetworkFlows, sm, ciliumDone)
+				falcoDone = nil
 			} else {
 				ciliumDone = nil
+				go manageStream(logger, connectAndStreamFalcoNetworkFlows, sm, falcoDone)
+
```
rlenglet, line 477: Nit.
```suggestion
```

rlenglet (changes requested):
Looks good!
Just a few minor nits.

rlenglet (approved):

PR #94, opened 2024-12-18
Optionally deploy Falco DaemonSet in Helm chart

evanj80-illumio:
Adding ability to have an illumio managed falco daemonset as part of the cloud-operator


cloud-operator/values.yaml
```diff
@@ -32,6 +32,35 @@ onboardingSecret:
...
+      enabled: true
+      keep_alive: true
+      url: "http://illumio-cloud-operator.illumio-cloud.svc.cluster.local:5000/"
+      insecure: true
+    json_output: true
+    json_include_tags_property: false
+    json_include_output_property: true
+    buffered_outputs: true
+  customRules:
+    rules-networking.yaml: |-
```
evanj80-illumio, line 49: This may not be the finalized rule or only rule at the end of the day. But this is good for now.

rlenglet (approved):
That's awesome, thank you!

cloud-operator/values.yaml
rlenglet: LGTM, thanks!

PR #96, opened 2024-12-18
Add Integration and End to End Test for Operator

evanj80-illumio:
Create the foundation of what integrations should look like and setting up how they will run via GHA.


.github/workflows/build_deploy_onboard.yml
```diff
@@ -0,0 +1,99 @@
+name: Build, Deploy and Onboard Cloud Operator To Live Services
```
evanj80-illumio, line 1: Ignore versions of some things in herer, going to go through GHA and fix versioning

cloud-operator/values.yaml
evanj80-illumio, line 1: IMO this should have been **false** to begin with. Making the change in this PR

cmd/main.go
```diff
@@ -112,7 +117,9 @@ func main() {
 		err := <-errChan
 		logger.Fatal("healthz check server failed", zap.Error(err))
 	}()
-
+	if envConfig.RunningGHA {
```
evanj80-illumio, line 120: This is required since we create the cluster first in GHA and the go test -v command needs time to spin up since its installs all of the required go packages. I think there may be a better version of this, maybe pre install go files as part of a GHA step. let me know what you think

fakeserver/fakeserver_integration_test.go
```diff
@@ -0,0 +1,160 @@
...
+	// nosemgrep: jwt.hardcoded-jwt-key
+	signedToken, err := jwtToken.SignedString(mySigningKey)
+	if err != nil {
+		logger.Error("Token could not be signed with fake secret key")
+	}
+	return signedToken
+
+}
+
+// Simulating a client interacting with the gRPC server
```
rlenglet, line 36:
```suggestion
// TestFakeServerConnectionSuccesfulAndRetry tests a client connecting to the gRPC server
```

fakeserver/fakeserver_integration_test.go
```diff
@@ -0,0 +1,160 @@
...
+			if stateChanged {
+				t.Log("Connection Succes is true. Test passed.")
+				assert.Equal(t, stateChanged, true)
+				break mainloop
+			}
+		}
+	}
+	fakeServer.stop()
+
+	time.Sleep(5 * time.Second)
```
rlenglet, line 84: Why this sleep?

fakeserver/fakeserver_integration_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Start the server
+	err := fakeServer.start()
+	assert.NoError(t, err, "Failed to start the FakeServer")
+
+	// Cleanup: Stop the server after the test
+	defer fakeServer.stop()
+
+	// Allow the server a bit of time to start
+	time.Sleep(2 * time.Second)
```
rlenglet, line 56: Why this sleep? This seems brittle. Can't you use a WaitGroup or something like that?

fakeserver/fakeserver_integration_test.go
```diff
@@ -0,0 +1,160 @@
...
```
rlenglet, line 135: Same here. Why this sleep?

integration-testing/onboard_test.go
```diff
@@ -0,0 +1,174 @@
+package integrationtesting
```
rlenglet, line 1: Add the copyright header.

fakeserver/oauth2_token_endpoint.go
```diff
@@ -63,17 +62,8 @@ func jsonResponse(w http.ResponseWriter, status int, data any) {
...
-	authService := &AuthService{
-		logger:       loggerToUse,
-		clientID:     clientID,
-		clientSecret: clientSecret,
-		token:        tokenString,
-	}
-	http.HandleFunc("/api/v1/k8s_cluster/authenticate", authService.authenticateHandler)
-	http.HandleFunc("/api/v1/k8s_cluster/onboard", authService.onboardCluster)
-
+func startOAuth2HTTPServer(address string, cert tls.Certificate, authService *AuthService) (*http.Server, error) {
```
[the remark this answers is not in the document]
evanj80-illumio: Was trying to be more file specific but your right. Also was a mistake to delete comment. Did not realize I had.

integration-testing/onboard_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	return clustersResponse.Clusters, nil
+}
+
+// offboardCluster sends a request to offboard a cluster by its ID.
+func offboardCluster(config Config, clusterIds []string) error {
+	offBoardURL := "https://cloud.illum.io/api/v1/k8s_cluster/offboard"
+
+	body := map[string][]string{
```
rlenglet, line 117: It would be more readable to define a `struct`s with a `json:"..."` tag on each field than to create JSON objects by hand.
Like you did for `ClustersResponse` above.

rlenglet (changes requested):
Looks good. Some more minor comments.

.github/workflows/integration_fakeserver_tests.yml
```diff
@@ -0,0 +1,107 @@
...
+        #     # --set env.tokenEndpoint=$TOKEN_ENDPOINT \
+        #     # --set env.tlsSkipVerify=$TLS_SKIP_VERIFY \
+        #     # --set falco.enabled=false
+      
+      - name:  Run Integration Tests
+        env:
+          ONBOARDING_CLIENT_ID: "client_id_1"
+          ONBOARDING_CLIENT_SECRET: "client_secret_1"
+          TLS_SKIP_VERIFY: true
+          ONBOARDING_ENDPOINT: "https://192.168.49.1:50053/api/v1/k8s_cluster/onboard"
```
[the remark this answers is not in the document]
evanj80-illumio: I looked into it and it seems more trouble then worth it

cloud-operator/values.yaml
```diff
@@ -33,7 +34,7 @@ onboardingSecret:
   clientSecret: ""
 
 falco:
-  enabled: true
+  enabled: false
```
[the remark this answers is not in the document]
evanj80-illumio: https://github.com/illumio/cloud-operator/pull/119

fakeserver/main.go
```diff
@@ -1,355 +1,46 @@
...
-	flag.Parse()
-
-	logger, err := zap.NewDevelopment()
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed to configure logger: %s", err)
-		os.Exit(1)
-	}
-	token = "token1"
+	logger, _ := zap.NewDevelopment()
+	aud := "192.168.65.254:50051"
```
rlenglet, line 12: Is it ok to hardcode this IP address, instead of keeping the command-line flag?
Doesn't that prevent using `fakeserver` for interactive dev testing now?

integration-testing/onboard_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+// getAuthorizationHeader returns the Basic Authorization header using the provided credentials.
+func getAuthorizationHeader(key, secret string) string {
+	credentials := fmt.Sprintf("%s:%s", key, secret)
+	encodedCredentials := base64.StdEncoding.EncodeToString([]byte(credentials))
+	return fmt.Sprintf("Basic %s", encodedCredentials)
+}
+
+// buildRequest creates a new HTTP request with common headers.
+func buildRequest(method, url string, body io.Reader, config Config) (*http.Request, error) {
```
rlenglet, line 60: NIt. I think "new" is more idiomatic.
```suggestion
func newRequest(method, url string, body io.Reader, config Config) (*http.Request, error) {
```

integration-testing/onboard_test.go
```diff
@@ -0,0 +1,191 @@
...
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("X-Tenant-Id", config.TenantId)
+	req.Header.Add("X-User-Id", config.UserId)
+
+	return req, nil
+}
+
+// fetchClusters makes an HTTP request to fetch a list of clusters.
+func fetchClusters(config Config) ([]Cluster, error) {
+	baseURL := "https://cloud.illum.io/api/v1/k8s_cluster" // Replace with actual URL
```
rlenglet, line 77: Is this comment still relevant?
```suggestion
	baseURL := "https://cloud.illum.io/api/v1/k8s_cluster"
```

integration-testing/onboard_test.go
```diff
@@ -0,0 +1,191 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("request failed with status: %v", resp.Status)
+	}
+
+	return nil
+}
+
+// Function to extract UUIDs from the list of clusters
+func extractUUIDs(clusters []Cluster) []string {
+	var uuids []string
```
rlenglet, line 151:
```suggestion
	uuids := make([]string, 0, len(clusters))
```

integration-testing/onboard_test.go
```diff
@@ -0,0 +1,191 @@
...
+	config := Config{
+		TenantId:     viper.GetString("TENANT_ID"),
+		CloudIdKey:   viper.GetString("CLOUD_API_KEY"),
+		CloudIdValue: viper.GetString("CLOUD_SECRET"),
+		UserId:       viper.GetString("USER_ID"),
+	}
+
+	// Fetch clusters
+	clusters, err := fetchClusters(config)
+	fmt.Println(clusters)
```
rlenglet, line 172:
```suggestion
```

integration-testing/onboard_test.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	// Validate the response
+	if len(clusters) == 0 {
+		t.Error("Expected non-empty response, got empty response")
+	}
+
+	// Check the first cluster and attempt to offboard it
+	uuids := extractUUIDs(clusters)
+	fmt.Println(uuids)
```
rlenglet, line 184:
```suggestion
```

integration-testing/onboard_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("request failed with status: %v", resp.Status)
+	}
+
+	return nil
+}
+
+// Function to extract UUIDs from the list of clusters
+func extractUUIDs(clusters []Cluster) []string {
```
rlenglet, line 150:
```suggestion
func getClusterIDs(clusters []Cluster) []string {
```

rlenglet (changes requested):
Looks good!
I just have a few minor nits.

fakeserver/main.go
evanj80-illumio: So when you are running the fakeserver through main.go it injects this IP so that it will work locally

.github/workflows/build_deploy_onboard.yml
evanj80-illumio, line 1: Will need to set env vars for this test @rlenglet

.github/workflows/build_deploy_onboard.yml
```diff
@@ -0,0 +1,99 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+  
+      - name: Build and push Docker image
+        id: build-and-push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          push: true
```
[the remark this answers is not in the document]
evanj80-illumio: Without push: true it wont push the temp image to GHCR

.github/workflows/build_deploy_onboard.yml
```diff
@@ -0,0 +1,108 @@
...
+          platforms: linux/amd64
+          build-args: |
+            VERSION=${{ github.sha }}
+          tags: |
+            ghcr.io/${{ github.repository }}:${{ github.sha }}
```
rlenglet, line 35: It would be better to use the `docker/metadata-action` to determine the tags:
https://github.com/docker/metadata-action
In case this is executed in the context of a PR, it generates a tag like `pr-96`, which would make it much easier to find and to distinguish from production image tags.

OK to do that later if you want.

rlenglet (approved):

.github/workflows/build_deploy_onboard.yml
rlenglet: Feel free to close if you prefer to move to that in another PR.

PR #100, opened 2025-01-02
Revising Falco Rule output

evanj80-illumio:
Changing rule output string to align with new language for timestamps.


aaronIllumio (approved):
LGTM

PR #104, opened 2025-01-11
Add Node.spec.providerID into ObjectMetadata message

evanj80-illumio:
Add the Node.spec.providerID as part of the Node resources sent up from cloud-operator.


api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -23,9 +23,16 @@ message KubernetesObjectData {
   string uid = 9;
   oneof kind_specific {
     KubernetesPodData pod = 100;
+    KubernetesNodeData node = 101;
   }
 }
 
+// KubernetesNodeData contains information specific to a Node Object.
+// We need to grab the providerId from each given Node.
```
rlenglet, line 31: This comment doesn't belong in the API.
```suggestion
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -23,9 +23,16 @@ message KubernetesObjectData {
...
+message KubernetesNodeData {
+  string provider_id = 1;
```
rlenglet, line 33: Clarify where this comes from.
```suggestion
  // Node.spec.providerID
  string provider_id = 1;
```

internal/controller/k8s_resources.go
```diff
@@ -85,9 +85,34 @@ func convertMetaObjectToMetadata(ctx context.Context, logger *zap.SugaredLogger,
 		}
 		objMetadata.KindSpecific = &pb.KubernetesObjectData_Pod{Pod: &pb.KubernetesPodData{IpAddresses: convertHostIPsToStrings(hostIPs)}}
 	}
+	if resource == "nodes" {
```
rlenglet, line 88: Replace the chain of `if`s with a `switch`.
```
	switch resource {
	case "pods":
		// ...
	case "nodes":
		// ...
	}
	return objMetadata, nil
```

internal/controller/k8s_resources.go
```diff
@@ -85,9 +85,34 @@ func convertMetaObjectToMetadata(ctx context.Context, logger *zap.SugaredLogger,
...
+
+		if err != nil {
+			return objMetadata, nil
+		}
+		objMetadata.KindSpecific = &pb.KubernetesObjectData_Node{Node: &pb.KubernetesNodeData{ProviderId: providerId}}
+	}
 	return objMetadata, nil
 }
 
+// getProviderIdNodeSpec uses a node name to grab the providerId within the nodeSpec
```
rlenglet, line 99:
```suggestion
// getProviderIdNodeSpec uses a node name to return the providerID within the node's spec
```

internal/controller/k8s_resources.go
```diff
@@ -85,9 +85,34 @@ func convertMetaObjectToMetadata(ctx context.Context, logger *zap.SugaredLogger,
...
+func getProviderIdNodeSpec(ctx context.Context, logger *zap.SugaredLogger, nodeName string) (string, error) {
+	clientset, err := NewClientSet()
```
rlenglet, line 101: Is it OK to create a new `ClientSet` for every single `Node` and every single `Pod`?
That seems expensive to create at scale.
Create it only once and reuse it.

internal/controller/k8s_resources.go
```diff
@@ -85,9 +85,34 @@ func convertMetaObjectToMetadata(ctx context.Context, logger *zap.SugaredLogger,
...
+		return "", err
+	}
+	node, err := clientset.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		return "", nil
+	}
+	if node.Spec.ProviderID != "" {
+		return node.Spec.ProviderID, nil
+	}
+	return "", nil
```
rlenglet, line 113: The semantics in the caller is that it sends a `KubernetesNodeData` if the returned error is `nil`, and doesn't when the error is not `nil`. Stick to that contract: if you have no provider ID to return, return a non-nil error.
```suggestion
	return "", errors.New("no providerID set")
```
Alternatively, modify the caller to ignore the returned provider ID if it is `""`.

rlenglet (changes requested):
Overall LGTM. Some relatively minor comments.

internal/controller/k8s_resources.go
evanj80-illumio: Did a larger rewrite to make this more efficient as you pointed out.

internal/controller/k8s_resources.go
```diff
@@ -67,7 +68,7 @@ func getMetadatafromResource(logger *zap.SugaredLogger, resource unstructured.Un
 }
 
 // convertMetaObjectToMetadata takes a metav1.ObjectMeta and converts it into a proto message object KubernetesMetadata.
-func convertMetaObjectToMetadata(ctx context.Context, logger *zap.SugaredLogger, obj metav1.ObjectMeta, resource string) (*pb.KubernetesObjectData, error) {
+func convertMetaObjectToMetadata(ctx context.Context, obj metav1.ObjectMeta, clientset *kubernetes.Clientset, resource string) (*pb.KubernetesObjectData, error) {
```
rlenglet, line 71: This starts to feel like this should be a struct with the `ClientSet` as a field, and `convertMetaObjectToMetadata`, `getPodIPAddresses`, `getProviderIdNodeSpec` should be methods of that struct.
OK to do that later. Not critical for this PR.

internal/controller/resource_manager.go
```diff
@@ -25,8 +26,8 @@ type ResourceManager struct {
 }
 
 // DynamicListAndWatchResources lists and watches the specified resource dynamically, managing context cancellation and synchronization with wait groups.
-func (r *ResourceManager) DyanmicListAndWatchResources(ctx context.Context, cancel context.CancelFunc, resource string, apiGroup string, allResourcesSnapshotted *sync.WaitGroup, snapshotCompleted *sync.WaitGroup) {
-	resourceListVersion, err := r.DynamicListResources(ctx, resource, apiGroup)
+func (r *ResourceManager) DyanmicListAndWatchResources(ctx context.Context, cancel context.CancelFunc, resource string, apiGroup string, clientset *kubernetes.Clientset, allResourcesSnapshotted *sync.WaitGroup, snapshotCompleted *sync.WaitGroup) {
```
rlenglet, line 29: This `ClientSet` should be a field of `ResourceManager` instead of new parameters to its methods.

rlenglet (changes requested):
Looks great. Just one nit about adding fields for the shared `ClientSet`, instead of passing it everywhere as a parameter.

internal/controller/k8s_resources.go
evanj80-illumio: Ill label it as TODO for something to clean up

rlenglet (approved):

PR #111, opened 2025-01-22
Send OwnerReferences and fix Kinds in objectMetadata

evanj80-illumio:
- Patch kind field to return proper cased K8s Kind
- Add namespace as a resource to send and watch
- Adding OwnerReference to k8s objectMetadata

### DIFF
--- a/internal/controller/k8s_resources.go
+++ b/internal/controller/k8s_resources.go
@@ -93,14 +93,13 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
 			return objMetadata, nil
 		}
 		objMetadata.KindSpecific = &pb.KubernetesObjectData_Pod{Pod: &pb.KubernetesPodData{IpAddresses: convertHostIPsToStrings(hostIPs)}}
-	case "nodes":
+	case "Node":
 		providerId, err := getProviderIdNodeSpec(ctx, clientset, obj.GetName())
 		if err != nil {
 			return objMetadata, nil
 		}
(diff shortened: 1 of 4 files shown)


internal/controller/resource_manager.go
```diff
@@ -113,7 +114,8 @@ func (r *ResourceManager) watchEvents(ctx context.Context, resource string, apiG
 			r.logger.Errorw("Cannot convert runtime.Object to metav1.ObjectMeta", "error", err)
 			return err
 		}
-		metadataObj, err := convertMetaObjectToMetadata(ctx, *convertedData, r.clientset, resource)
+		resource := event.Object.GetObjectKind().GroupVersionKind().Kind
```
evanj80-illumio, line 117: We pass the old resourse format here "pods" or "nodes" because we need that format to start a Watch as thats what k8s api takes. Then within that watchEvents we get the new Kind per event to give to resource stream.

internal/controller/k8s_resources.go
```diff
@@ -96,6 +103,38 @@ func convertMetaObjectToMetadata(ctx context.Context, obj metav1.ObjectMeta, cli
 	return objMetadata, nil
 }
 
+func convertOwnerReferences(ownerReferences []metav1.OwnerReference) ([]*pb.KubernetesOwnerReference, error) {
+	if len(ownerReferences) == 0 {
+		return nil, nil
+	}
+
+	var result []*pb.KubernetesOwnerReference
```
rlenglet, line 111: Pre-allocate the slice.
```suggestion
	result := make([]*pb.KubernetesOwnerReference, 0, len(ownerReferences))
```

rlenglet (changes requested):
Looks great, thanks!
Just one nit.

rlenglet (approved):

PR #121, opened 2025-02-11
devcontainer: bump go version to 1.23

AriSweedler:
First commit :)


.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG VARIANT=1.22
+ARG VARIANT=1.23-bookworm
```
rlenglet, line 1: This is the default value of the arg if it's not overridden.
In our case, it _is_ overridden, here:
https://github.com/illumio/cloud-operator/blob/9e2aa1d8332797b68f5cd6c4cd16742cad9c591f/.devcontainer/devcontainer.json#L6
Update that line too.

rlenglet (changes requested):
One other line needs to be updated.

PR #124, opened 2025-02-13
Addressing Falco Flows with Incomplete L4 Messages

evanj80-illumio:
We observed Falco returning incomplete L4 network logs. Instead of sending those incomplete flows to CloudSecure we want to just ignore them since we cannot make much sense of them.

### DIFF
--- a/internal/controller/flow_falco.go
+++ b/internal/controller/flow_falco.go
@@ -131,6 +131,7 @@ func createLayer3Message(source string, destination string, ipVersion string) (*
 	} else if ipVersion == "ipv6" {
 		return &pb.IP{Source: source, Destination: destination, IpVersion: pb.IPVersion_IP_VERSION_IPV6}, nil
 	}
+	// If this is IPVersion_IP_VERSION_IP_NOT_USED_UNSPECIFIED do we want to drop this packet?
 	return &pb.IP{Source: source, Destination: destination, IpVersion: pb.IPVersion_IP_VERSION_IP_NOT_USED_UNSPECIFIED}, nil
 }
 
@@ -172,5 +173,5 @@ func createLayer4Message(proto string, srcPort, dstPort uint32, ipVersion string
 		}
 	default:
 	}
-	return &pb.Layer4{}, nil
+	return &pb.Layer4{}, ErrFalcoIncompleteL4Flow
 }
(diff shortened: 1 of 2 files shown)


internal/controller/flow_falco.go
```diff
@@ -70,21 +70,21 @@ func parsePodNetworkInfo(input string) (*pb.FalcoFlow, error) {
 
 	layer3Message, err := createLayer3Message(info.SrcIP, info.DstIP, info.IpVersion)
 	if err != nil {
-		return nil, fmt.Errorf("unable to create Layer3 message falco flows: %v", err)
+		return &pb.FalcoFlow{}, err
```
rlenglet, line 73: Why not just return `nil`?
```suggestion
		return nil, err
```
Same with the other error returns below.

rlenglet (changes requested):

rlenglet (approved):

internal/controller/streams.go
```diff
@@ -85,6 +85,9 @@ var resourceAPIGroupMap = map[string]string{
 var dd = &deadlockDetector{}
 var ErrStopRetries = errors.New("stop retries")
 var ErrFalcoEventIsNotFlow = errors.New("ignoring falco event, not a network flow")
+var ErrFalcoIncompleteL3Flow = errors.New("ignoring incomplete falco l3 network flow")
+var ErrFalcoIncompleteL4Flow = errors.New("ignoring incomplete falco l4 network flow")
+var ErrFalcoInvalidPort = errors.New("ignoring incomplete falco flow due to bad ports")
```
AriSweedler, line 90: Do you have a testcase for `ErrFalcoInvalidPort`? Looks like you got everything else. Just below the `"Incomplete L4 TCP flow"` testcase:
```
		"Invalid port": {
			input:    "time=1987-02-22T00:39:07.267635635+0000\t srcip=192.168.1.1 dstip=192.168.1.2 srcport=invalid dstport=5678 proto= ipversion=ipv4",
			expected: nil,
			err:      ErrFalcoInvalidPort,
		},
```

internal/controller/streams.go
```diff
@@ -240,8 +243,10 @@ func (sm *streamManager) StreamFalcoNetworkFlows(ctx context.Context) error {
 			}
 
 			convertedFalcoFlow, err := parsePodNetworkInfo(match[1])
-			if errors.Is(err, ErrFalcoEventIsNotFlow) {
+			if convertedFalcoFlow == nil {
```
AriSweedler, line 246: The go style is to use `err`, not the value being nil or not. Right? If there's a reason for the departure from the norm, plz explain.

evanj80-illumio: Or maybe you are right, maybe I should just check if there is an error. Since in all cases where we drop a flow we error out anyways. So yeah you are right we probably should have stuck to that norm!

internal/controller/streams.go
evanj80-illumio: I missed this test case.

PR #127, opened 2025-02-13
Ingest Service clusterIps as a new K8sObject specific metadata

evanj80-illumio:
Relatively ~~easy~~ **not easy** implementation that ingests clusterIps that are associated with the already ingested k8s Service metadata.


fakeserver/server.go
```diff
@@ -124,7 +124,6 @@ func (s *server) SendKubernetesResources(stream pb.KubernetesInfoService_SendKub
 			serverState.ConnectionSuccessful = true
 		case *pb.SendKubernetesResourcesRequest_KubernetesResourceMutation:
 			logger.Info("Mutation Detected")
-			logger.Info(req.String())
```
evanj80-illumio, line 127: This is just clean up on fakeserver, we dont need this to log all of these 😄

rlenglet (changes requested):

internal/controller/k8s_resources.go
```diff
@@ -99,10 +99,32 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
 	return objMetadata, nil
 }
 
+// getServiceClusterIp retrieves the Cluster IP of a given service in a specified namespace.
+func getServiceClusterIp(ctx context.Context, serviceName string, clientset *kubernetes.Clientset, namespace string) (string, error) {
+	service, err := clientset.CoreV1().Services(namespace).Get(ctx, serviceName, metav1.GetOptions{})
+	if err != nil {
+		return "", errors.New("failed to get service")
+	}
+	clusterIp := service.Spec.ClusterIP
```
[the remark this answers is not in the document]
rlenglet: While we're at it, I'd also read the `externalIPs` and `loadBalancerIP` fields if they're set.
And after you make a list of all of those, sort the list.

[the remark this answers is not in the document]
rlenglet: And it would be great to also add the IP addresses from `status.loadBalancer.ingress` if it's set, for services of type `LoadBalancer`.
https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer

internal/controller/k8s_resources.go
```diff
@@ -99,10 +99,32 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+	if clusterIp == "" {
+		return "", errors.New("service does not have a cluster IP")
```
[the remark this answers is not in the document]
evanj80-illumio: Then should I just set an empty clusterIp?

internal/controller/k8s_resources.go
```diff
@@ -99,10 +114,109 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+		}
+		if i.Hostname != "" {
+			result = append(result, i.Hostname)
+		}
+	}
+	return result
+}
+
+func convertServicePortsToNodePorts(servicePorts []v1.ServicePort) []*NodePort {
+	var nodePorts []*NodePort
```
rlenglet, line 143: Pre-allocate?
```suggestion
	nodePorts := make([]*NodePort, 0, len(servicePorts))
```

internal/controller/k8s_resources.go
```diff
@@ -99,10 +114,109 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+}
+
+func getServiceAttributes(ctx context.Context, serviceName string, clientset *kubernetes.Clientset, namespace string) (*ServiceAttributes, error) {
+	service, err := clientset.CoreV1().Services(namespace).Get(ctx, serviceName, metav1.GetOptions{})
+	if err != nil {
+		return nil, errors.New("failed to get service")
+	}
+
+	attributes := &ServiceAttributes{}
+	attributes.ClusterIPs = service.Spec.ClusterIPs
```
rlenglet, line 164: What if just `spec.clusterIP` is set?
Shouldn't you add it if it's set?

internal/controller/k8s_resources.go
```diff
@@ -99,10 +114,109 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+		attributes.ExternalName = service.Spec.ExternalName
+
+	case v1.ServiceTypeClusterIP:
+		// ClusterIP-specific processing is not required as ClusterIPs and ExternalIPs are common for other types as well.
+
+	case v1.ServiceTypeNodePort:
+		attributes.NodePorts = convertServicePortsToNodePorts(service.Spec.Ports)
+
+	case v1.ServiceTypeLoadBalancer:
+		attributes.NodePorts = convertServicePortsToNodePorts(service.Spec.Ports)
```
rlenglet, line 178: I wouldn't mix those with NodePorts. It's a completely different beast.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -24,9 +24,24 @@ message KubernetesObjectData {
...
+message NodePorts {
+  // The port on each node on which this service is exposed
+  int32 node_port = 1;
+  // The port that will be exposed by this service
+  int32 port = 2;
+  // The IP protocol for this port, supports "TCP", "UDP", and "SCTP"
+  string protocol = 3;
+}
+
+message KubernetesServiceData {
```
rlenglet, line 40: Add the service type. That's important information.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -24,9 +24,24 @@ message KubernetesObjectData {
...
+  repeated string ip_addresses = 1;
+  repeated NodePorts node_ports = 2;
```
rlenglet, line 42: Add another field for `LoadBalancer` ports? I think we shouldn't mix them with `NodePort`s.
Or better, just keep this named `ports` as it's named `ports` in the k8s `spec`. That would be more consistent.
```suggestion
  repeated ServicePorts ports = 2;
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -24,9 +24,24 @@ message KubernetesObjectData {
   oneof kind_specific {
     KubernetesPodData pod = 100;
     KubernetesNodeData node = 101;
+    KubernetesServiceData service = 102;
   }
 }
 
+message NodePorts {
```
rlenglet, line 31: 1. Move this message definition into the `KubernetesServiceData` definition below. I don't think this would ever need to be reused in another context.
1. Make this singular: ports -> port
1. As I suggested below, let's name this `Port` since it's containing info from the `spec.ports` field.

rlenglet (changes requested):

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -24,6 +24,27 @@ message KubernetesObjectData {
...
+  message ServicePorts {
+    // The port on each node on which this service is exposed
+    int32 node_port = 1;
+    // The port that will be exposed by this service
+    int32 port = 2;
+    // The IP protocol for this port, supports "TCP", "UDP", and "SCTP"
+    string protocol = 3;
+    // List of ports exposing the service via a load balancer
+    repeated string load_balencer_ports = 4;
   }
```
rlenglet, line 48: 1. Naming nits.
1. `node_port` should be optional.
```suggestion
  message ServicePort {
    // The port on each node on which this service is exposed
    optional uint32 node_port = 1;
    // The port that will be exposed by this service
    uint32 port = 2;
    // The IP protocol for this port, supports "TCP", "UDP", and "SCTP"
    string protocol = 3;
    // List of ports exposing the service via a load balancer
    repeated string load_balancer_ports = 4;
  }
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -24,6 +24,27 @@ message KubernetesObjectData {
...
+  }
+}
+
+message KubernetesServiceData {
+  // List of IP addresses associated with the Kubernetes service
+  repeated string ip_addresses = 1;
+  // List of service ports details
+  repeated ServicePorts ports = 2;
+  // Type of Kubernetes service (e.g., ClusterIP, NodePort, LoadBalancer, etc.)
+  string service_type = 3;
```
rlenglet, line 37: Remove the `service_` prefix, it's redundant with the message name, and just using `type` matches the name of the field in the k8s `Service` spec.
I don't think `type` is a reserved keyword??
```suggestion
  string type = 3;
```

internal/controller/k8s_resources.go
```diff
@@ -20,6 +20,25 @@ import (
...
+type ServiceAttributes struct {
+	ClusterIPs          []string
+	ExternalIPs         []string
+	LoadBalancerIP      string
+	LoadBalancerIngress []string
+	LoadBalancerClass   string
+	NodePorts           []*Ports
+	ExternalName        string
+	ServiceType         string // make sure you have this field to capture the service type.
+}
```
rlenglet, line 33: 1. Define the non-slice optional fields as pointers.
1. Is `LoadBalancerIngress` really a list? In which case it should be plural.
```suggestion
type ServiceAttributes struct {
	ClusterIPs            []string
	ExternalIPs           []string
	LoadBalancerIP        *string
	LoadBalancerIngresses []string
	LoadBalancerClass     *string
	Ports                 []*Ports
	ExternalName          *string
	Type                  string
}
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -24,6 +24,27 @@ message KubernetesObjectData {
   oneof kind_specific {
     KubernetesPodData pod = 100;
     KubernetesNodeData node = 101;
+    KubernetesServiceData service = 102;
+  }
+}
+
+message KubernetesServiceData {
```
rlenglet, line 31: Add the missing fields from your `ServiceAttributes` struct.
Most of them should be `optional`.

internal/controller/k8s_resources.go
```diff
@@ -20,6 +20,25 @@ import (
...
+
+type Ports struct {
+	NodePort          int32
```
rlenglet, line 36: Optional.
```suggestion
	NodePort          *int32
```

internal/controller/k8s_resources.go
```diff
@@ -20,6 +20,25 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+// Assuming ServiceAttributes has appropriate fields
+type ServiceAttributes struct {
```
rlenglet, line 24: Why do you need those new structs in the first place? Can't you have `getServiceAttributes` directly return a protobuf?

internal/controller/k8s_resources.go
```diff
@@ -99,10 +118,116 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+	nodePorts := make([]*Ports, 0, len(servicePorts))
+	for _, sp := range servicePorts {
+		if sp.NodePort != 0 {
+			nodePort := &Ports{
+				NodePort: sp.NodePort,
+				Port:     sp.Port,
+				Protocol: string(sp.Protocol),
+			}
+			nodePorts = append(nodePorts, nodePort)
+		}
```
rlenglet, line 156:
```suggestion
		port := &Ports{
			Port:     sp.Port,
			Protocol: string(sp.Protocol),
		}
		if sp.NodePort != 0 {
			port.NodePort = sp.NodePort
		}
		ports = append(ports, port)
```

internal/controller/k8s_resources.go
```diff
@@ -99,10 +118,116 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+}
+
+func convertServicePortsToNodePorts(servicePorts []v1.ServicePort) []*Ports {
+	nodePorts := make([]*Ports, 0, len(servicePorts))
+	for _, sp := range servicePorts {
+		if sp.NodePort != 0 {
+			nodePort := &Ports{
+				NodePort: sp.NodePort,
+				Port:     sp.Port,
+				Protocol: string(sp.Protocol),
```
rlenglet, line 153: Beware that `protocol` is optional and defaults to `"TCP"`. Handle the case when it's not set.

internal/controller/k8s_resources.go
```diff
@@ -99,10 +118,116 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+		return nil, errors.New("failed to get service")
+	}
+	loadBalancerIngress := []string{}
+	if len(service.Status.LoadBalancer.Ingress) > 0 {
+		loadBalancerIngress = convertIngressToStringList(service.Status.LoadBalancer.Ingress)
+	}
+	loadBalancerClass := ""
+	if service.Spec.LoadBalancerClass != nil {
+		loadBalancerClass = *service.Spec.LoadBalancerClass
+	}
```
rlenglet, line 173: Make the field `optional` in the protobuf message, so the field becomes a pointer and you can keep everything a pointer.
```suggestion
```

internal/controller/k8s_resources.go
```diff
@@ -99,10 +118,116 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+	attributes := &ServiceAttributes{
+		ClusterIPs:          service.Spec.ClusterIPs,
+		ExternalIPs:         service.Spec.ExternalIPs,
+		LoadBalancerIngress: loadBalancerIngress,
+		LoadBalancerIP:      service.Spec.LoadBalancerIP,
+		LoadBalancerClass:   loadBalancerClass,
```
rlenglet, line 179:
```suggestion
		LoadBalancerClass:   service.Spec.LoadBalancerClass,
```

internal/controller/k8s_resources.go
```diff
@@ -99,10 +118,116 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+	}
+
+	if service.Spec.Type == v1.ServiceTypeNodePort || service.Spec.Type == v1.ServiceTypeLoadBalancer {
+		attributes.NodePorts = convertServicePortsToNodePorts(service.Spec.Ports)
+	}
```
rlenglet, line 184: Convert ports unconditionally regardless of the service type.
```suggestion
	attributes.Ports = convertServicePortsToNodePorts(service.Spec.Ports)
```

internal/controller/k8s_resources.go
```diff
@@ -99,10 +118,116 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+
+	// Only set ExternalName if the service type is ExternalName
+	if service.Spec.Type == v1.ServiceTypeExternalName {
+		attributes.ExternalName = service.Spec.ExternalName
+	}
```
rlenglet, line 189: Same thing here, simplify and just capture the external name regardless of the service type.
```suggestion
```

internal/controller/k8s_resources.go
```diff
@@ -99,10 +118,116 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+	if len(service.Status.LoadBalancer.Ingress) > 0 {
+		loadBalancerIngress = convertIngressToStringList(service.Status.LoadBalancer.Ingress)
+	}
+	loadBalancerClass := ""
+	if service.Spec.LoadBalancerClass != nil {
+		loadBalancerClass = *service.Spec.LoadBalancerClass
+	}
+	attributes := &ServiceAttributes{
+		ClusterIPs:          service.Spec.ClusterIPs,
+		ExternalIPs:         service.Spec.ExternalIPs,
```
rlenglet, line 176:
```suggestion
		ExternalIPs:         service.Spec.ExternalIPs,
		ExternalName:        service.Spec.ExternalName,
```

internal/controller/k8s_resources.go
```diff
@@ -99,10 +118,116 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+// Convert ServiceAttributes to KubernetesServiceData
+func convertToKubernetesServiceData(attributes *ServiceAttributes) *pb.KubernetesServiceData {
+	// Combine all IPs
+	combinedIPs := combineIPAddresses(attributes)
+
+	// Convert NodePorts to ServicePorts
+	servicePorts := make([]*pb.KubernetesServiceData_ServicePorts, len(attributes.NodePorts))
+	for i, np := range attributes.NodePorts {
+		servicePorts[i] = &pb.KubernetesServiceData_ServicePorts{
+			NodePort:          int32(np.NodePort),
```
rlenglet, line 215: This should be optional.

internal/controller/k8s_resources.go
evanj80-illumio: Yeah it is a list.

internal/controller/k8s_resources.go
evanj80-illumio: Nice catch on this one.

internal/controller/k8s_resources.go
evanj80-illumio: Im not quite sure what you mean?

internal/controller/k8s_resources.go
```diff
@@ -99,10 +107,101 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
+			port.NodePort = &sp.NodePort
+		}
+		ports = append(ports, port)
+	}
+	return ports
+}
+
+// Combine all IPs into a single list
+func combineIPAddresses(clusterIps []string, externalIps []string, loadBalancerIngresses []string, loadBalancerIp string) []string {
+	combinedIPs := []string{}
```
rlenglet, line 154: Really minor nit. Not a PR blocker. Feel free to resolve this conversation without fixing.
```suggestion
	combinedIPs := make([]string, 0, len(clusterIps)+len(externalIps)+len(loadBalancerIngresses)+1)
```

internal/controller/k8s_resources.go
```diff
@@ -20,6 +20,14 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+// Assuming ServiceAttributes has appropriate fields
+type Ports struct {
```
rlenglet, line 24: You could replace this with `pb.KubernetesServiceData_ServicePort` directly. There's really no difference between those two structs.
If you want to clean this up in another PR, I'm OK with that.

internal/controller/k8s_resources_test.go
```diff
@@ -616,3 +635,249 @@ func (suite *ControllerTestSuite) TestExtractObjectMetas() {
...
+		suite.Run(name, func() {
+			result := convertIngressToStringList(tt.ingress)
+			assert.Equal(suite.T(), tt.expectedResult, result)
+		})
+	}
+}
+
+func (suite *ControllerTestSuite) TestConvertServicePortsToPorts() {
+
+	tests := map[string]struct {
```
rlenglet, line 684: I'd like to see another test case for the case where `Protocol` is `""`, which should default to `"TCP"`.

internal/controller/k8s_resources.go
```diff
@@ -99,10 +107,101 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
 	return objMetadata, nil
 }
 
+func convertIngressToStringList(ingresses []v1.LoadBalancerIngress) []string {
+	result := []string{}
+	for _, ingress := range ingresses {
+		if ingress.IP != "" {
+			result = append(result, ingress.IP)
+		} else if ingress.Hostname != "" {
+			result = append(result, ingress.Hostname)
```
rlenglet, line 127: OK with capturing non-IP addresses in `ip_addresses` for now.
Those will have to be filtered out and ignored in the backend until we decide what to do with those.

rlenglet (dismissed):
Looks great!
I have only relatively minor comments about test cases, more cleanups, etc. but all of those can be done in a future PR. Nothing blocking.
So feel free to close those conversations, merge the PR, and address those in another PR.

internal/controller/k8s_resources.go
```diff
@@ -99,10 +107,101 @@ func convertMetaObjectToMetadata(logger *zap.SugaredLogger, ctx context.Context,
...
 	}
 	return objMetadata, nil
 }
 
+func convertIngressToStringList(ingresses []v1.LoadBalancerIngress) []string {
+	result := []string{}
+	for _, ingress := range ingresses {
+		if ingress.IP != "" {
+			result = append(result, ingress.IP)
+		} else if ingress.Hostname != "" {
```
rlenglet, line 126: This will fix your failing test case in `TestConvertIngressToStringList`:
```suggestion
		}
		if ingress.Hostname != "" {
```

internal/controller/k8s_resources.go
evanj80-illumio: 👍

PR #128, opened 2025-02-14
Cleanup

AriSweedler:
- Update comment on line 94 in resource_manager.go
- remove `master` in the github action
- Use "helm install action" in helm_lint_and_test instead of hand-rolling it (it works now! Didn't 6 months ago)


.devcontainer/Dockerfile
```diff
@@ -34,6 +34,7 @@ RUN curl -fsSL -o /tmp/helm.tar.gz https://get.helm.sh/helm-${HELM_VERSION}-linu
 USER vscode
 
 RUN go install github.com/bufbuild/buf/cmd/buf@${BUF_VERSION}
+RUN go install sigs.k8s.io/kind@v0.26.0
```
evanj80-illumio, line 37: Whats the reason for putting this in the Dockerfile?

evanj80-illumio (commented):
Tiny little comment

.devcontainer/Dockerfile
AriSweedler: Spoke in person - recording this here because Raman is goated and he does that.

This is only for the `.devcontainer`, and the integration tests use `kind`. (Devcontainer ITs still fail but this is at least a step in the right direction)

evanj80-illumio (approved):
Thank you for this!

.devcontainer/Dockerfile
rlenglet, line 37: I like to define all installed versions in one place as arguments at the top, so add this at the top after `HELM_VERSION`:
```
ARG KIND_VERSION=v0.26.0
```
And use it here:
```suggestion
RUN go install sigs.k8s.io/kind@${KIND_VERSION}
```

rlenglet (changes requested):
LGTM. Just one nit about specifying the Kind version in `Dockerfile`.

.devcontainer/Dockerfile
AriSweedler: Done

Nice, that's a strictly superior way of doing things. I'll remember this 👍

AriSweedler:
I just missed this notification & forgot about this branch. I just wrote a filter for `reason:author` and pinned it to my sidebar - that will make it easier for me to not make this mistake in the future

evanj80-illumio (approved):

PR #133, opened 2025-02-17
Fix Falco rules with new ipversion field + updating cloud-operator reboot logic to cope with long disconnect periods

evanj80-illumio:
- Add a field to Falco rule output in order to capture ip version.
- Fix a race condition when checking for hubble relay vs deciding which CNI to use.
- Fix the reboot timer logic for when a cluster is disconnected to disabled for a long period of time. After 10 severeErrors (or 10 failed connection attempts) the operator would return and stop attempting to connect. Modify that case to reset everything on the reset timer. Long term we may need to rewrite this reboot logic to distinguish between a cluster that has been disabled vs one that is actually having issues.


rlenglet (approved):
LGTM, thanks!

PR #134, opened 2025-02-20
Revising context timeout logic when requesting a new token

evanj80-illumio:
The context we had previously attached to our token refresh call had a timeout of 10 seconds. This means that our token would only be valid for that long. We are adjusting our context to only wrap the requests that retrieve a new token with a timeout to still failover if the request does not complete, but not to nullify the token itself.

Also adding consistency with how we are reading and writing k8s secrets.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"net/http"
-	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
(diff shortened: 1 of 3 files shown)


AriSweedler (approved):
Image from iOS

Instead of letting the `parent context` timeout, we now let the `child context` timeout. That way when we are in `SetUpOAuthConnection` we only have 10 seconds to get the token source, but we can use it for >10 seconds.

The `parent context` is the `authContext`, which should live for a long time - as long as the returned token is valid (1800 seconds, 30 minutes). When `authContext` gets canceled, it should only be because we need to get a new token.

rlenglet (approved):
Great, thanks!

PR #137, opened 2025-02-22
Fix gops binary

evanj80-illumio:
Build `gops` as a static binary.
Install `gops` into `/sbin` in the Docker image.
Set `GOPS_CONFIG_DIR` to `/var/run/gops` to contain the gops Unix Domain Sockets.
Mount an empty dir at `/var/run/gops` to make it writeable, as the rest of the filesystem is read-only.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,7 @@ ARG TARGETARCH
 WORKDIR /workspace
 
 # Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
+COPY go.mod go.sum ./
 
 # Cache dependencies before copying the source to avoid re-downloading
 RUN go mod download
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -23,17 +23,20 @@ COPY api/ api/
...
-# Install debugging tools (including bash) and gops for troubleshooting
+# Install gops for troubleshooting
 RUN go install github.com/google/gops@latest
 
 # Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 
+
+ENV GOPS_CONFIG_DIR="/var/run/gops"
```
AriSweedler, line 33: This `ENV` line shouldn't be it's own thunk. Either add it as part of the `FROM` thunk or as part of the `COPY` thunk.

I think I prefer adding it to the `COPY` thunk. I think `FROM + WORKDIR` is one thunk, then `COPY --from=builder /workspace/manager .` is another. And finally, `#set up gops` is the 3rd.

cmd/main.go
```diff
@@ -97,9 +97,10 @@ func main() {
 		"token_endpoint", envConfig.TokenEndpoint,
 		"tls_skip_verify", envConfig.TlsSkipVerify,
 	)
+	gopsDirectory := os.Getenv("GOPS_CONFIG_DIR")
```
AriSweedler, line 100: What do you want to do if this is unset. Silence? Log a warning?

AriSweedler (dismissed):

Dockerfile
evanj80-illumio: Yeah agreed that make sense

cmd/main.go
evanj80-illumio: I think we can log a warning

AriSweedler (approved):

rlenglet (approved):

AriSweedler (commented):
Nice. Thanks for adding me to this review so I could see 👍

PR #141, opened 2025-02-28
Fix gRPC logger to log structured fields

rlenglet:
- Reimplement gRPC logger as a Zap Core wrapper instead of a Core hook, so it can log the structured log fields.
- Replace `SugaredLogger` with `Logger` in the whole codebase to improve efficiency and eliminate formatting bugs.
- Encode each entry into a string immediately when logged, before buffering, and never refer to the input Entry and Fields afterwards, as it is reduces the buffer's memory usage and it allows the entry and fields to be garbage-collected while a log entry is buffered.
- Fix log buffering to stop buffering when reaching the buffer capacity.
- Eliminate a channel leak by not replacing the `done` channel without closing it when updating the gRPC connection and stream client.

### DIFF
--- a/internal/controller/grpc_logger.go
+++ b/internal/controller/grpc_logger.go
@@ -302,17 +302,15 @@ func NewGRPCLogger(grpcSyncer *BufferedGrpcWriteSyncer, addCaller bool, clock za
 	consoleCore := zapcore.NewCore(encoder, consoleSyncer, atomicLevel)
 
 	// Create zap logger with the console core
-	logger := zap.New(consoleCore,
+	logger := zap.New(
+		&zapCoreWrapper{
+			core:       consoleCore,
+			encoder:    encoder,
+			grpcSyncer: grpcSyncer,
+		},
 		zap.WithCaller(addCaller),
 		zap.AddStacktrace(zapcore.ErrorLevel),
 		zap.WithClock(clock),
-		zap.WrapCore(func(core zapcore.Core) zapcore.Core {
-			return &zapCoreWrapper{
-				core:       core,
-				encoder:    encoder,
-				grpcSyncer: grpcSyncer,
-			}
-		}),
 	)
 
 	grpcSyncer.logger = logger
--- a/internal/controller/grpc_logger.go
+++ b/internal/controller/grpc_logger.go
@@ -182,7 +182,6 @@ func (b *BufferedGrpcWriteSyncer) UpdateClient(client pb.KubernetesInfoService_S
 	b.mutex.Lock()
 	b.client = client
 	b.conn = conn
-	b.done = make(chan struct{})
 	b.flush()
 	b.mutex.Unlock()
 }
--- a/internal/controller/grpc_logger.go
+++ b/internal/controller/grpc_logger.go
@@ -3,13 +3,8 @@
 package controller
 
 import (
-	"encoding/json"
-	"errors"
-	"fmt"
 	"io"
-	"maps"
 	"os"
-	"slices"
 	"sync"
 	"time"
 
@@ -37,7 +32,7 @@ type BufferedGrpcWriteSyncer struct {
 	buffer              []*zapcore.Entry
 	mutex               sync.Mutex
 	done                chan struct{}
-	logger              *zap.SugaredLogger
+	logger              *zap.Logger
 	logLevel            zap.AtomicLevel
 	encoder             zapcore.Encoder
 	lostLogEntriesCount int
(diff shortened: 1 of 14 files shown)


internal/controller/grpc_logger.go
```diff
@@ -78,7 +83,7 @@ func (b *BufferedGrpcWriteSyncer) flush() {
 		lostLogEntry := &zapcore.Entry{
 			Level:   zap.ErrorLevel,
 			Time:    time.Now(),
-			Message: "Lost logs due to buffer overflow",
+			Message: `{"msg":"Lost logs due to buffer overflow"}`,
```
AriSweedler, line 86: The purpose of this PR is confusing to me because I am missing context.

But given that the purpose is correct, can we expose an easier to use API? Keep the data (log message) separate from the metadata. Like so:
```suggestion
			Message: iloLogMessage("Lost logs due to buffer overflow").AsString(),
```

Or like so:
```
iloLogMessage{
    Level:   zap.ErrorLevel,
    Message: "Lost logs due to buffer overflow",
}.AsZapEntry()
```

----

Maybe this is not the type of thing that needs to get written a lot tho,,, and it's fine to leave as-is

internal/controller/grpc_logger.go
```diff
@@ -122,10 +127,50 @@ func (b *BufferedGrpcWriteSyncer) run() {
 
 // sendLog sends the log as a string to the server.
 func (b *BufferedGrpcWriteSyncer) sendLog(logEntry *zapcore.Entry, extraFields []zap.Field) error {
-	buf, err := b.encoder.EncodeEntry(*logEntry, extraFields)
+	var decodedMessage map[string]any
+
+	// Assume that the message has been serialized into a JSON object
```
AriSweedler, line 132: This is necessary given what we have, but it should also be the type of thing that can be caught at compile-time. I cannot think of how to accomplish this other than changing the parameter type from `*zapcore.Entry` to some local type.

internal/controller/grpc_logger.go
rlenglet: This is not the final solution, as I don't think it's working.
This PR is not yet ready for review.
Please hold off for a bit...

internal/controller/grpc_logger.go
rlenglet: This is not the final solution.
This PR is not yet ready for review.
Please hold off for a bit...

internal/controller/grpc_logger.go
```diff
@@ -120,23 +126,31 @@ func (b *BufferedGrpcWriteSyncer) run() {
...
+// encodeLogEntry encodes a log Entry and fields into a string using the given Encoder.
+func encodeLogEntry(encoder zapcore.Encoder, logEntry zapcore.Entry, fields []zap.Field) (string, error) {
+	buf, err := encoder.EncodeEntry(logEntry, fields)
 	if err != nil {
-		return err
+		return "", fmt.Errorf("failed to encode log entry: %w", err)
 	}
-	err = b.client.Send(&pb.SendLogsRequest{
+
+	// Remove any newline added by Zap's encoder
```
evanj80-illumio, line 136: Is there a reason that this is happening? Or just something that the zap encoder does?

internal/controller/grpc_logger.go
```diff
@@ -169,11 +183,11 @@ func (b *BufferedGrpcWriteSyncer) ListenToLogStream() error {
 }
 
 // bufferLog adds the log entry to in-memory buffer
-func (b *BufferedGrpcWriteSyncer) bufferLogEntry(entry *zapcore.Entry) {
+func (b *BufferedGrpcWriteSyncer) bufferLogEntry(jsonMessage string) {
 	if len(b.buffer) >= maxBufferSize {
```
evanj80-illumio, line 187: I think this may be problematic no? We are still attempting to buffer messages even when the buffer is full.

evanj80-illumio (dismissed):
just a few comments!

internal/controller/grpc_logger.go
rlenglet: Yes, I noticed while testing that the Zap encoder adds a newline.

internal/controller/grpc_logger.go
rlenglet: Wow, good catch!

rlenglet: Fix that and added a unit test. Thanks!

internal/controller/grpc_logger.go
```diff
@@ -120,44 +126,79 @@ func (b *BufferedGrpcWriteSyncer) run() {
...
-	return err
 }
 
-// UpdateClient will update BufferedGrpcWriteSyncer with new client stream and GRPC connection
+// UpdateClient updates the gRPC connection and connection in the BufferedGrpcWriteSyncer.
 func (b *BufferedGrpcWriteSyncer) UpdateClient(client pb.KubernetesInfoService_SendLogsClient, conn ClientConnInterface) {
 	b.mutex.Lock()
 	b.client = client
 	b.conn = conn
-	b.done = make(chan struct{})
```
evanj80-illumio, line 144: Does removing this not have any side effects?

internal/controller/grpc_logger.go
```diff
@@ -120,44 +126,79 @@ func (b *BufferedGrpcWriteSyncer) run() {
...
-			b.logger.Info("Server has closed the SendLogs stream")
+			b.logger.Info("Server closed the SendLogs stream")
 			return nil
 		}
 		if err != nil {
-			b.logger.Errorw("Stream terminated", "error", err)
+			b.logger.Error("Stream terminated", zap.Error(err))
 			return err
 		}
 		switch res.Response.(type) {
```
evanj80-illumio, line 204: @pavankumarinnamuri I think for your work this should be removed.

evanj80-illumio (approved):

internal/controller/grpc_logger.go
rlenglet: Nope, I checked, and the `done` channel is supposed to be closed once and only once by `Stop`, which is in practice never called.
So here, every time we reconnect, we were leaking one channel for no reason.

internal/controller/grpc_logger.go
rlenglet: Yep, but to avoid breaking API backward-compatibility, we will do that in phases.
First, just ignore the responses on the operator side, and remove the logic that sends those responses on the server side. Deploy the server to prod and release the operator.
Then, when we can remove that response from the API spec.

PR #143, opened 2025-03-04
Add IP addresses into Node inventory data

evanj80-illumio:


internal/controller/k8s_resources.go
```diff
@@ -110,6 +114,23 @@ func convertMetaObjectToMetadata(logger *zap.Logger, ctx context.Context, obj me
 	return objMetadata, nil
 }
 
+// getNodeIpAddresses fetches the IP addresses of a node
+func getNodeIpAddresses(ctx context.Context, clientset *kubernetes.Clientset, nodeName string) ([]string, error) {
+	node, err := clientset.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
+	if err != nil {
+		return nil, errors.New("failed to get node")
+	}
+	ipAddresses := []string{}
```
rlenglet, line 123: No need to instantiate an empty list. Returning `nil` is fine if there is no IP address.
```suggestion
	var ipAddresses []string
```

rlenglet: Same in quite a few other places in this file, where we create empty slices statically:
https://github.com/illumio/cloud-operator/blob/377cca810a95608e095a0f580c1e167581ed2ece/internal/controller/k8s_resources.go#L114
https://github.com/illumio/cloud-operator/blob/377cca810a95608e095a0f580c1e167581ed2ece/internal/controller/k8s_resources.go#L180
https://github.com/illumio/cloud-operator/blob/377cca810a95608e095a0f580c1e167581ed2ece/internal/controller/k8s_resources.go#L269

And in other cases, we create slices based on the length of some other collection, which may be zero:
https://github.com/illumio/cloud-operator/blob/377cca810a95608e095a0f580c1e167581ed2ece/internal/controller/k8s_resources.go#L135
https://github.com/illumio/cloud-operator/blob/377cca810a95608e095a0f580c1e167581ed2ece/internal/controller/k8s_resources.go#L164
https://github.com/illumio/cloud-operator/blob/main/internal/controller/k8s_resources.go#L189
https://github.com/illumio/cloud-operator/blob/377cca810a95608e095a0f580c1e167581ed2ece/internal/controller/k8s_resources.go#L274

We should try to return `nil` instead of an empty slice as much as possible, since both have the same behavior, and `nil` is obviously more compact.

rlenglet (approved):
LGTM.

One nit about allocating empty slices where `nil` would do. I'm OK with merging this PR without fixing that...

internal/controller/k8s_resources.go
evanj80-illumio: Sure that makes sense Ill make a note to fix this stuff in another PR

PR #154, opened 2025-03-24
Run and build operator without Docker Hub

evanj80-illumio:
Adding the ability to build the cloud-operator to a local registry and pull from that local registry into a local minikube instance for testing with fakeserver.

Also subbing out IP addresses for internal docker address.


rlenglet (dismissed):
LGTM, but I'd like to have a better mechanism to override Helm values for development instead of copying-and-pasting the Helm chart's file.

cloud-operator/values-for-local-dev.yaml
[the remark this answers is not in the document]
evanj80-illumio: Yeah thats a good point. I think my logic was to have this file laid out so that people can see what is different between the 2 for local development. But I think you may be right. I think ill change it to instead employ helm --set and add to the README

Makefile
```diff
@@ -40,10 +42,32 @@ docker-push:
...
 	docker push $(DOCKER_IMAGE):latest
 
-# Deploy target (build and push Docker image)
+# Deploy target (build and push Docker image to Docker Hub)
 .PHONY: deploy
 deploy: docker-build docker-push
 
+# Create and run a local Docker registry
+.PHONY: local-registry
+local-registry:
```
AriSweedler, line 51: nit: namespace this to `docker`. As in:
```suggestion
docker-registry-local:
```

That way, all docker commands start with `docker`. Also not entirely sure if the `||` is correct. It could fail for ANY reason (docker isn't running, image can't be pulled (no internet), container with that name is already running).

It is better to have no unhappy-path output than incorrect unhappy-path output.

README.md
```diff
@@ -61,7 +61,35 @@ docker login
...
+`
+make deploy-local
+`
+
+To deploy using helm and to test the operator using fakeserver here is an example of the command with `--set` args
+`
+helm install illumio --namespace illumio-cloud --create-namespace \
+  --set image.repository=host.docker.internal:5000/operator \
+  --set image.tag=latest \
+  --set onboardingSecret.clientId=client_id_1 \
```
AriSweedler, line 86: opinion: Instead of the 5 sets:
```
  --set onboardingSecret.clientId=client_id_1 \
  --set onboardingSecret.clientSecret=client_secret_1 \
  --set env.tlsSkipVerify=true \
  --set env.onboardingEndpoint=https://host.docker.internal:50053/api/v1/k8s_cluster/onboard \
  --set env.tokenEndpoint=https://host.docker.internal:50053/api/v1/k8s_cluster/authenticate \
```

I recommend `--values ./fakeserver/cloud-operator.fakeserver.yaml`: https://github.com/illumio/cloud-operator/blob/main/fakeserver/cloud-operator.fakeserver.yaml

----

THEN I recommend setting the `image.repository` according to this strategy as well. Include a "`*.yaml` plugin" to get `cloud-operator` to pull from a local registry.

Eventually I can imagine a wrapper utility where we can just add flags `./bin/local-install --target=fakeserver --registry=local` to run a local install of k8s. When you let people state their intention like this, you start to get more freedom to implement it. Maybe one day the `./bin/XXX build` would create the local registry transparently without a user having to worry about it. And trying to RUN form a local registry would check that the image is available there before running the `helm install` command. Advanced error checking with nice diagnostic messages instead of a shell one-liner in a Makefile that's harder to build up.

----

Action item: Shorten this `helm install` command by using `--value`. Create a `*.yaml` file for setting `image.*` to get a local image.

README.md
```diff
@@ -61,7 +61,35 @@ docker login
 make docker-build docker-push DOCKER_USERNAME=arisweedler386
 ```
 
-TODO: ENABLE LOCAL BUILD
```
AriSweedler, line 64: You have some structured content here. I don't think it's a bad idea to hoist this to a whole new document. Or at least it's own header. The summary is as follows:

1. Run a local registry
2. Push a build to the local registry
3. Get a k8s cluster to pull from a local registry

The `making a private build` is a good header for this, but that's a pre-req to the set of 3 things. Keep the information like a tree and it's easier to navigate

AriSweedler (dismissed):

rlenglet (dismissed):

AriSweedler (approved):

PR #162, opened 2025-03-27
Implement auditing and logging for cloud-operator

pavankumarinnamuri:
- Added CniPluginStatus enumeration and used it within KubernetesClusterMetadata message in k8s_info.proto.
- Implemented logic in streams.go to determine and log CNI plugin status.
- Added log entries for the following scenarios:
  - cloud-operator has restarted
  - cloud-operator finally connected but only after many retries
  - cloud-operator found no CNI plugin in the cluster
  - cloud-operator found no sources of traffic flows
- Updated streams_helper.go with helper functions to send various types of data to the resource stream.
- Ensured that all relevant log entries are present and correctly implemented for auditing purposes.


internal/controller/streams_helper.go
```diff
@@ -89,7 +89,7 @@ func streamMutationObjectData(sm *streamManager, metadata *pb.KubernetesObjectDa
 }
 
 // sendClusterMetadata sends a message to indicate current cluster metadata
-func sendClusterMetadata(ctx context.Context, sm *streamManager) error {
+func sendClusterMetadata(ctx context.Context, sm *streamManager, cniStatus pb.CniPluginStatus) error {
```
evanj80-illumio, line 92:
```suggestion
func sendClusterMetadata(ctx context.Context, sm *streamManager) error {
```
based on comments above

internal/controller/streams_helper.go
```diff
@@ -105,12 +105,25 @@ func sendClusterMetadata(ctx context.Context, sm *streamManager) error {
...
 		return err
 	}
+	if cniStatus == pb.CniPluginStatus_CNI_PLUGIN_STATUS_NONE_FOUND {
+		sm.logger.Warn("No CNI plugin found: emitting audit log")
+		// Optionally send an audit event or log to CloudSecure
+		if sm.streamClient.logStream != nil {
+			logEntry := &pb.LogEntry{
+				JsonMessage: "no CNI plugin found in cluster (sent via ClusterMetadata)",
+			}
+			_ = sm.streamClient.logStream.Send(&pb.SendLogsRequest{
```
evanj80-illumio, line 115: It may be useful here to log if there is an error in this stream. This could become a race condition, line 111 evaluates as stream is open and then before it gets to line 115 the stream has died. Now we try to send this message and it doesnt work.

evanj80-illumio: Or even return here since if the stream is dead we want to propagate that back up to the restart logic.

internal/controller/streams.go
```diff
@@ -505,7 +525,25 @@ func ConnectStreams(ctx context.Context, logger *zap.SugaredLogger, envMap Envir
...
+				streamClient.cniStatus = pb.CniPluginStatus_CNI_PLUGIN_STATUS_CILIUM_PRESENT
+			case falcoPresent:
+				streamClient.cniStatus = pb.CniPluginStatus_CNI_PLUGIN_STATUS_FALCO_PRESENT
+			default:
+				streamClient.cniStatus = pb.CniPluginStatus_CNI_PLUGIN_STATUS_NONE_FOUND
+				// Log cloudOperatorNoCNIPlugin
+				logEntry := &pb.LogEntry{
+					JsonMessage: "cloud-operator no CNI plugin found",
+				}
+				_ = bufferedGrpcSyncer.client.Send(&pb.SendLogsRequest{
```
evanj80-illumio, line 543: Have a comment below that covers an issue with this type of sending a message downstream without catching an error

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -89,6 +89,16 @@ message KubernetesClusterMetadata {
   // Version of the operator. This version is following the Semver 2.0.0 format, e.g. "1.2.0".
   // https://semver.org/spec/v2.0.0.html
   string operator_version = 3;
+
+  CniPluginStatus cni_plugin_status = 4;
```
evanj80-illumio, line 93: Please add comments on each of these proto entries

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -89,6 +89,16 @@ message KubernetesClusterMetadata {
...
+}
+
+enum CniPluginStatus {
```
evanj80-illumio, line 96: Add comments

evanj80-illumio (changes requested):

internal/controller/streams.go
```diff
@@ -433,6 +436,7 @@ func manageStream(logger *zap.SugaredLogger, connectAndStream func(*zap.SugaredL
 // ConnectStreams will continue to reboot and restart the main operations within
 // the operator if any disconnects or errors occur.
 func ConnectStreams(ctx context.Context, logger *zap.SugaredLogger, envMap EnvironmentConfig, bufferedGrpcSyncer *BufferedGrpcWriteSyncer) {
+	//logger.Info("cloud-operator restarting")
```
evanj80-illumio, line 439: Get rid of this

evanj80-illumio (changes requested):
I would like to see some unit tests for some of these changes, otherwise changes look good!

internal/controller/streams.go
rlenglet: Unresolving. This comment has not been addressed.

rlenglet (changes requested):

internal/controller/streams_helper.go
```diff
@@ -105,12 +105,29 @@ func sendClusterMetadata(ctx context.Context, sm *streamManager) error {
...
+			}
+			err = sm.streamClient.logStream.Send(&pb.SendLogsRequest{
+				Request: &pb.SendLogsRequest_LogEntry{LogEntry: logEntry},
+			})
+			if err != nil {
+				sm.logger.Errorw("Failed to send no CNI plugin found log in sendClusterMetadata", "error", err)
+				return err
+			}
+		}
+	}
```
rlenglet, line 123: Remove. Reporting this in `KubernetesClusterMetadata` is sufficient.
```suggestion
```

internal/controller/streams.go
```diff
@@ -543,13 +570,47 @@ func ConnectStreams(ctx context.Context, logger *zap.SugaredLogger, envMap Envir
...
+	}
+
+	_, err = clientset.CoreV1().Namespaces().Get(ctx, "falco", metav1.GetOptions{})
+	if err != nil {
+		sm.logger.Warn("Falco namespace not found, assuming Falco is not installed")
+		return false
+	}
+
+	return true
 }
```
rlenglet, line 614: Remove. We already know whether we enabled Falco network flow collection.
```suggestion
```

internal/controller/streams.go
```diff
@@ -26,6 +28,7 @@ type streamClient struct {
 	disableNetworkFlowsCilium bool
 	networkFlowWg             sync.WaitGroup
 	falcoEventChan            chan string
+	cniStatus                 pb.CniPluginStatus
```
rlenglet, line 31:
```suggestion
	flowCollector             pb.FlowCollector
```

internal/controller/streams.go
```diff
@@ -505,7 +509,30 @@ func ConnectStreams(ctx context.Context, logger *zap.SugaredLogger, envMap Envir
 				logger:             logger,
 				bufferedGrpcSyncer: bufferedGrpcSyncer,
 			}
-
+			ciliumFlowCollector, err := newCiliumFlowCollector(ctx, logger, envMap.CiliumNamespace)
```
rlenglet, line 512: This is creating a Cilium flow collector just to check whether this succeeds?
This creates a gRPC client and doesn't close it, which basically may leak connections and goroutines.
We shouldn't have to do that.

internal/controller/streams.go
```diff
@@ -505,7 +509,30 @@ func ConnectStreams(ctx context.Context, logger *zap.SugaredLogger, envMap Envir
...
+					JsonMessage: "cloud-operator no CNI plugin found",
+				}
+				err := bufferedGrpcSyncer.client.Send(&pb.SendLogsRequest{
+					Request: &pb.SendLogsRequest_LogEntry{LogEntry: logEntry},
+				})
+				if err != nil {
+					logger.Warnw("Failed to send cloudOperatorNoCNIPlugin log entry", "error", err)
+				} else {
+					logger.Infow("Emitted cloudOperatorNoCNIPlugin log entry")
+				}
```
rlenglet, line 534: There is no need to log.
Reporting this in `KubernetesClusterMetadata` is sufficient.
```suggestion
```

internal/controller/streams.go
```diff
@@ -543,13 +570,47 @@ func ConnectStreams(ctx context.Context, logger *zap.SugaredLogger, envMap Envir
 				failureReason = "Log stream closed"
 			}
 			authConContextCancel()
+			logger.Warn("One or more streams have been closed; restarting all streams", zap.String("reason", failureReason))
```
rlenglet, line 573: Why was it needed to duplicate the log below?
https://github.com/illumio/cloud-operator/pull/162/files#diff-65d4066e2022c168a97a2083474d348fa6d5d1fffb6e8d7378282935ad187940R590
```suggestion
```

internal/controller/streams.go
```diff
@@ -543,13 +570,47 @@ func ConnectStreams(ctx context.Context, logger *zap.SugaredLogger, envMap Envir
...
+							JsonMessage: fmt.Sprintf("cloud-operator connected successfully after %d retries", attempt-1),
+						},
+					},
+				})
+				if err != nil {
+					logger.Warn("Failed to send cloudOperatorConnectedAfterRetries log entry", zap.Error(err))
+				} else {
+					logger.Infow("Emitted cloudOperatorConnectedAfterRetries audit log entry", "retry_count", attempt-1)
+				}
+			}
```
rlenglet, line 587: 1. I don't see why this is needed?
1. Why directly call the SendLogs RPC instead of just relying on the logger to do its job?
1. This code will change significantly with #164. I recommend that you don't modify anything in this loop to prevent painful merge conflicts. Limit the scope of this PR to adding the `FlowCollector` into the cluster metadata.
```suggestion
```

rlenglet (changes requested):

internal/controller/streams.go
evanj80-illumio: The only way we know that falco is enabled is if 
1) we get flows from the falco agent
2) the config in the credentials file is changed to include falco

1 is a valid way to know that falco is enabled, but it more work than is needed. 2 is problematic because someone could have enabled falco but when they helm install a number of things could have happened without them knowing or understanding (namespace did not exist, no permissions to deploy a daemonset etc). I think looking for the namespace is a pretty good way of knowing if falco is enabled. However, to truly know if falco is enabled maybe we can build on this function and look specifically for the falco agent in the falco namespace?

evanj80-illumio: @rlenglet What are your thoughts on this?

rlenglet: How will you know that the Falco rules are setup correctly to export flows to cloud-operator? How will you know that there are no NetworkPolicies preventing us from receiving those flows?

My proposal is to just stop pretending that we're detecting CNI plugins or flow exporters.
It's too difficult, and it's inaccurate.

My proposal is to just reporting what flow _**collector**_ the cloud-operator has setup.
Hence the proposed renaming of the message to "Flow Collector". This is about reporting how cloud-operator has decided that we're going to collect flows, for which cloud-operator is 100% in control.

internal/controller/streams.go
pavankumarinnamuri: i have a new PR opend for flowCollector into the cluster metadata.

PR #164, opened 2025-03-28
Simplify the sleep timer/reset timer in manageStreams & fix ConnectStreams channel leak

AriSweedler:
There are 2 clear cut-points: the exponential backoff (sleepTimer) and the reset backoff (resetTimer). These can both be modeled as exponential backoffs. Here's the behavior that we get when using this double backoff. The blue line is how much we are sleeping. The light pink line is how many times the resetCount (severe error) has triggered. As you can see, it resets the backoff to baseline.

image


internal/controller/streams.go
```diff
@@ -363,76 +367,50 @@ func connectAndStreamLogs(logger *zap.Logger, sm *streamManager) error {
...
-			} else {
-				consecutiveFailures = 0
-				backoff = initialBackoff
-				// Reset sleep timer back to initial backoff interval
-				sleepTimer.Reset(initialBackoff)
-				resetTimer.Reset(resetPeriod)
-			}
-		}
+
+	lambda := func() error {
```
rlenglet, line 378: I don't see the term "lambda" used in Golang. The term "anonymous function" is used instead. Let's stick to that.
```suggestion
	f := func() error {
```

internal/controller/streams.go
```diff
@@ -363,76 +367,50 @@ func connectAndStreamLogs(logger *zap.Logger, sm *streamManager) error {
...
+		return connectAndStream(logger, sm)
+	}
+
+	// If a stream goes down, try to reconnect. With exponential backoff
+	lambdaWithBackoff := func() error {
```
rlenglet, line 383:
```suggestion
	funcWithBackoff := func() error {
```

internal/controller/streams.go
```diff
@@ -363,76 +367,50 @@ func connectAndStreamLogs(logger *zap.Logger, sm *streamManager) error {
...
+			Name:                 "connectToStreamWithBackoff",
+			logger:               logger,
+		}, lambda)
+	}
+
+	// If reapated attempts to connect fail, that is, the "SevereErrorThreshold"
+	// in the above backoff is triggered, then wait and try again. By setting
+	// ExponentialFactor to 1, we will wait the same amount of time between every
+	// attempt. This is desirable
+	lambdaWithBackoffAndReset := func() error {
```
rlenglet, line 399:
```suggestion
	funcWithBackoffAndReset := func() error {
```

internal/controller/streams.go
```diff
@@ -363,76 +367,50 @@ func connectAndStreamLogs(logger *zap.Logger, sm *streamManager) error {
...
+		return exponentialBackoff(backoffOpts{
+			InitialBackoff:       10 * time.Minute,
+			MaxBackoff:           10 * time.Second,
+			MaxJitterPct:         0.10,
+			SevereErrorThreshold: math.MaxInt,
+			// Setting ExponentialFactor 1 will cause the backoff timer to stay
+			// constant.
+			ExponentialFactor: 1,
+			Name:              "connectToStreamWithBackoffAndReset",
+			logger:            logger,
```
rlenglet, line 409: Instead of passing in the name as a parameter and then adding it explicitly in every log statement, just set it here in the logger's context only once:
```suggestion
			logger:            logger.With(zap.String("name", "reset_and_retry")),
```

internal/controller/streams.go
```diff
@@ -363,76 +367,50 @@ func connectAndStreamLogs(logger *zap.Logger, sm *streamManager) error {
...
+	// If a stream goes down, try to reconnect. With exponential backoff
+	lambdaWithBackoff := func() error {
+		return exponentialBackoff(backoffOpts{
+			InitialBackoff:       1 * time.Second,
+			MaxBackoff:           1 * time.Minute,
+			MaxJitterPct:         0.20,
+			SevereErrorThreshold: 10,
+			ExponentialFactor:    2.0,
+			Name:                 "connectToStreamWithBackoff",
+			logger:               logger,
```
rlenglet, line 391: 1. Why not give the name of the function that is called?
1. Instead of passing in the name as a parameter and then adding it explicitly in every log statement, just set it here in the logger's context only once.
```suggestion
			logger:               logger.With(zap.String("name", "connect_and_stream")),
```

internal/controller/streams_retry.go
```diff
@@ -0,0 +1,110 @@
...
+	timer               *time.Timer
+	consecutiveFailures int
+	opts                backoffOpts
+}
+
+func exponentialBackoff(opts backoffOpts, action Action) error {
+	s := state{
+		backoff:             opts.InitialBackoff,
+		consecutiveFailures: 0,
+		timer:               time.NewTimer(0),
```
rlenglet, line 44: Add a comment. This is important and subtle.
```suggestion
		timer:               time.NewTimer(0), // Don't wait before the first attempt to execute the action.
```

internal/controller/streams_retry.go
```diff
@@ -0,0 +1,110 @@
...
+		opts:                opts,
+	}
+	defer s.timer.Stop()
+	opts.logger.Info("Starting backoff", zap.String("name", opts.Name), zap.Any("opts", opts))
```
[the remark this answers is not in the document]
AriSweedler: > Keep it structured by using either zap.Object or zap.Inline (preferred here IMHO).

TIL about `zap.Inline`. I had to implement `MarshalLogObject` which was very good to do (as we don't want to log the `opts.logger` object, so we just leave it out)

I will keep this in mind for future complex logs like this.

> Isn't this too noisy at Info level? I'd make this Debug, like the log in AddBackoff.

You're right - this makes a lot more sense as a debug log. I will downgrade it. We only start an exponential backoff sequence during interesting events (channel error --> try to reconnect; 10 minute reset timer goes off). But the rising edge of an interesting event should always have an error log. 

> We're not backing off at this point, so I wouldn't mention "backoff" until we're actually retrying.

🫡

internal/controller/streams_retry.go
```diff
@@ -0,0 +1,110 @@
...
+
+type backoffOpts struct {
+	Name                 string
+	InitialBackoff       time.Duration
+	MaxBackoff           time.Duration
+	MaxJitterPct         float64
+	SevereErrorThreshold int
+	ExponentialFactor    float64
+	logger               *zap.Logger
+}
```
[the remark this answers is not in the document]
AriSweedler: TIL about the pattern of `var _ interface = &obj`. Makes sense - gets the compiler to do this interface c'tor to verify that it works. Even if it is not used anywhere in the real world. It really sets intention that you can use `obj` as `interface`.

AriSweedler:
Had to add a commit to fix a lint error. 13520fc2c19a6e7217483a7156d9a6bd3c380f08

I wanted to add a context to make backoffs cancelable instead of changing the `for { select {` into a `for range`. But that would require some logic upstream. It didn't quite make sense to me to do that in this PR.

Also,,, the parent never really wants to cancel this task. Only this task cancels itself (when it fails). I will say though, that if the parent could cancel this task, then that actually might help with the keepalive issue - it would bring us from `red` to `blue` in this diagram. Explaining the diagram a bit more - `message 1` (M1) goes through, then the server disconnects. We don't notice that until `message 2` (M2). We force M2 with the keepalives, so it's fast enough for us to never care. But if we noticed a stream disconnecting and turned that into a ctx cancel, then we could be event-driven instead of poll-driven.

keepalives

The only problem with canceling contexts of streams is... we don't wanna overdo it. How can we tell if a stream goes down from no connection to the server (they ALL should go down) vs. a blip & JUST this stream losing connection vs. any other error. There's probably a way to do that right (and if we want to proceed like that, a new PR will implement it). But,,, why complicate things. It's actually quite fine to just wait the configurable keepalive time to get the task to realize it has no connection and cancel itself.

rlenglet (approved):
LGTM, thanks!

PR #165, opened 2025-03-31
Add flow_collector to KubernetesClusterMetadata

pavankumarinnamuri:
Add a `flow_collector` field into the `KubernetesClusterMetadata` message.


api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -91,6 +91,20 @@ message KubernetesClusterMetadata {
...
   // https://semver.org/spec/v2.0.0.html
   string operator_version = 3;
+  // The network flow collection mechanism configured in the operator's cluster
+  FlowCollector flow_collector = 4;
+}
+
+// Enum for CniPluginStatus that defines possible states of the CNI plugin.
+enum FlowCollector {
+  // Default value. This value is unused and should not be set.
+  FLOW_COLLECTOR_UNSPECIFIED = 0;
```
AriSweedler, line 101: nit: Any time you wanna refer to this value in code you'll need to prefix with the enum type. Thus, the `FLOW_COLLECTOR_` prefix is redundant.
```suggestion
  UNSPECIFIED = 0;
```

AriSweedler: When you make this change, it will be very text-heavy unless you use LSP renaming. Which your IDE should support!

internal/controller/streams.go
```diff
@@ -25,6 +25,7 @@ type streamClient struct {
 	client                    pb.KubernetesInfoServiceClient
 	disableNetworkFlowsCilium bool
```
AriSweedler, line 26: Get rid of `disableNetworkFlowsCilium` if you can! Keep it DRY

internal/controller/streams.go
```diff
@@ -513,10 +514,16 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
...
 			} else {
 				sm.streamClient.disableNetworkFlowsCilium = false
+				sm.streamClient.flowCollector = pb.FlowCollector_FLOW_COLLECTOR_CILIUM
+			}
+			// If neither is available
+			if sm.streamClient.flowCollector != pb.FlowCollector_FLOW_COLLECTOR_CILIUM &&
+				sm.streamClient.flowCollector != pb.FlowCollector_FLOW_COLLECTOR_FALCO {
+				sm.streamClient.flowCollector = pb.FlowCollector_FLOW_COLLECTOR_DISABLED
 			}
-
```
AriSweedler, line 519: nit: pref keeping these `thunk`s apart.

this is a style thing, but I find it helpful to think of code (and navigate it) in terms of `thunks`. Thunks should be indivisible. As an example of a thunk: It doesn’t make sense to say `err = tx.Commit(ctx)` without checking `if err != nil {`, so you would never put a whitespace between them. That's a thunk. those lines of code that do work are inseparable from the lines of code that error check and that log and that record metrics. It's basically like a mini function - the thunk is inseparable. A thunk can have an output or a side effect, but it should not be broken up.

So for this newline - by removing it, you're stating that determining what the `flowCollector` is is inseparable from creating these channels. Which is not quite true.

I would even recommend hoisting the `detectFlowCollector` logic into its own function.

AriSweedler (dismissed):

internal/controller/streams.go
AriSweedler: Do a global search on this variable - I don't think it shows up very often.

internal/controller/streams.go
AriSweedler: It's always a great refactor to realize that you are taking an overly simple piece of code (one thunk) into multiple thunks. Then group those thunks together with a function name. It makes the calling code read more easily.

And it makes the callee unit testable. Well - maybe it's hard to unit test this, because you would have to mock the inputs. But hey - the input which is `findHubbleRelay` is already hoisted into its own function so it's really nice to not have to worry about that set of thunks that implements it. If everything was inlined it might be harder to thread new logic through it.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: The PR follows the Protocol Buffer Style Guide:
https://protobuf.dev/programming-guides/style/#enums

rlenglet (approved):
LGTM, thanks!

rlenglet:
@pavankumarinnamuri rebase on main and re-generate the code to resolve conflicts in `k8s_info.pb.go`.

rlenglet (approved):

internal/controller/streams.go
```diff
@@ -36,6 +36,7 @@ type streamClient struct {
 	client                    pb.KubernetesInfoServiceClient
 	disableNetworkFlowsCilium bool
 	falcoEventChan            chan string
+	flowCollector             pb.FlowCollector
```
rlenglet, line 39: It might happen that the cluster medatata is sent to the server before the flow collector has been decided.
In this case, we’ll send “UNSPECIFIED”.

Solution 1:

You should initialize this field to “DISABLED”.
Also, we need to send a new cluster metadata whenever this field’s value changes.
That would be quite complicated because that would require changing the state machine on the server side.

Solution 2:

As a better alternative, I would recommend blocking sending the cluster metadata until the flow collector has been decided (it’s not UNDEFINED anymore).

pavankumarinnamuri: @rlenglet I'll go with the sync.WaitGroup approach that will ensure it waits until the flow collector decision is made before sending the cluster metadata.

rlenglet: Good idea!

internal/controller/streams.go
```diff
@@ -570,8 +575,24 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
...
 				sm.streamClient.disableNetworkFlowsCilium = false
+				sm.streamClient.flowCollector = pb.FlowCollector_FLOW_COLLECTOR_CILIUM
+			}
+			// If neither is available
+			if sm.streamClient.flowCollector != pb.FlowCollector_FLOW_COLLECTOR_CILIUM &&
+				sm.streamClient.flowCollector != pb.FlowCollector_FLOW_COLLECTOR_FALCO {
+				sm.streamClient.flowCollector = pb.FlowCollector_FLOW_COLLECTOR_DISABLED
+			}
+			// Wait for the flow collector decision
+			sm.streamClient.flowCollectorWG.Done()
```
rlenglet, line 589: I'm sorry, I gave you the wrong feedback.
Re-reading this, it turns out we don't need any synchronization, since we decide the flow collector before we even start the resources stream that will send the cluster metadata, cf. this line below:
https://github.com/illumio/cloud-operator/pull/165/files#diff-65d4066e2022c168a97a2083474d348fa6d5d1fffb6e8d7378282935ad187940R603
Therefore, we don't need to add this WaitGroup, we're already good without it.
```suggestion
```

internal/controller/streams.go
```diff
@@ -560,7 +562,10 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
 				ciliumNamespace:           envMap.CiliumNamespace,
 				disableNetworkFlowsCilium: false,
 				falcoEventChan:            falcoEventChan,
+				flowCollector:             pb.FlowCollector_FLOW_COLLECTOR_UNSPECIFIED,
 			}
+			// Initialize the WaitGroup with a count of 1.
+			streamClient.flowCollectorWG.Add(1)
```
rlenglet, line 568: I'm sorry, I gave you the wrong feedback.
We don't need synchronization / a WaitGroup. We're already doing things in the right order.
```suggestion
```

internal/controller/streams.go
```diff
@@ -36,6 +36,8 @@ type streamClient struct {
 	client                    pb.KubernetesInfoServiceClient
 	disableNetworkFlowsCilium bool
 	falcoEventChan            chan string
+	flowCollector             pb.FlowCollector
+	flowCollectorWG           sync.WaitGroup
```
rlenglet, line 40: I'm sorry, I gave you the wrong feedback.
We don't need synchronization / a WaitGroup. We're already doing things in the right order.
```suggestion
```

internal/controller/streams_helper.go
```diff
@@ -94,6 +94,9 @@ func streamMutationObjectData(sm *streamManager, metadata *pb.KubernetesObjectDa
 
 // sendClusterMetadata sends a message to indicate current cluster metadata
 func sendClusterMetadata(ctx context.Context, sm *streamManager) error {
+	// Wait until the flow collector decision is complete.
+	sm.streamClient.flowCollectorWG.Wait()
+
```
rlenglet, line 99: I'm sorry, I gave you the wrong feedback.
We don't need synchronization / a WaitGroup. We're already doing things in the right order.
```suggestion
```

rlenglet (changes requested):
I'm sorry, I gave you the wrong feedback.
We don't synchronization / a WaitGroup. We're already doing things in the right order.
Please remove the new code that adds a WaitGroup.
Sorry for the churn.

rlenglet (dismissed):
Neat, thanks!

evanj80-illumio (approved):

evanj80-illumio (approved):
Approved

rlenglet (approved):

PR #171, opened 2025-04-09
Correctly appending Pod IP addresses to object metadata

evanj80-illumio:
Found that the pod ip addresses were resolving as the node ip address. Looked closer and found that we were sending up the wrong `status` attribute as the pod ip addresses.


aaronIllumio (approved):
LPTM

PR #174, opened 2025-04-14
Annotating clustercreds to protect in Helm Upgrade

evanj80-illumio:


internal/controller/authenticator.go
```diff
@@ -86,28 +85,31 @@ func (authn *Authenticator) DoesK8sSecretExist(ctx context.Context, secretName s
...
+// WriteK8sSecret updates the data in an existing Kubernetes Secret without overwriting annotations or labels.
 func (authn *Authenticator) WriteK8sSecret(ctx context.Context, keyData OnboardResponse, ClusterCreds string) error {
 	clientset, err := NewClientSet()
 	if err != nil {
 		authn.Logger.Error("Failed to create clientSet", zap.Error(err))
 		return err
 	}
 
-	secretData := map[string][]byte{
+	namespace := "illumio-cloud" // TODO: Make this configurable
```
rlenglet, line 96: The Pod can discover its own namespace.
In the pod spec's `env`, add:
```
            - name: MY_POD_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
```
Per https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables

internal/controller/authenticator.go
```diff
@@ -86,28 +85,31 @@ func (authn *Authenticator) DoesK8sSecretExist(ctx context.Context, secretName s
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      ClusterCreds,
-			Namespace: namespace,
-		},
-		Data: secretData,
-	}
 
-	_, err = clientset.CoreV1().Secrets(namespace).Update(ctx, secret, metav1.UpdateOptions{})
+	// Apply the update
+	_, err = clientset.CoreV1().Secrets(namespace).Update(ctx, existingSecret, metav1.UpdateOptions{})
```
rlenglet, line 112: You could reuse the same `clientset.CoreV1().Secrets(namespace)` client as above?

rlenglet (changes requested):
LGTM, great solution!
However, you should not hardcode the Pod's namespace.

rlenglet (approved):

PR #177, opened 2025-04-15
Flowstream fails reconnect

AriSweedler:
`StreamFalcoNetworkFlows` was refusing to reconnect when it was disabled. Because it wouldn't listen to the `ctx` getting cancelled.

Similarly, `StreamResources` needs to cancel the context when it exits, so that the `exponentialBackoff` function can notice the failures and back off.


internal/controller/streams_retry.go
```diff
@@ -62,7 +62,7 @@ func exponentialBackoff(opts backoffOpts, action Action) error {
 	for range s.timer.C {
 		err := action()
 
-		if err != nil {
+		if err == nil {
```
evanj80-illumio, line 65: Not quite following this change! Can you explain it a little bit

internal/controller/streams.go
```diff
@@ -323,8 +324,14 @@ func (sm *streamManager) StreamCiliumNetworkFlows(ctx context.Context, logger *z
 // StreamFalcoNetworkFlows handles the falco network flow stream.
 func (sm *streamManager) StreamFalcoNetworkFlows(ctx context.Context, logger *zap.Logger) error {
 	for {
-		falcoFlow := <-sm.streamClient.falcoEventChan
-		if filterIllumioTraffic(falcoFlow) {
+		select {
+		case <-ctx.Done():
```
evanj80-illumio, line 328: This is listening to the parent context in the manageStreams to see if its been cancelled right?

AriSweedler: Yes, that's correct.

The keepalive fails to send & cancels the context. The whole task-tree should go down. However, this function stayed alive. And thus the `connectAndStreamFlows` function in `manageStreams` was never exiting.

AriSweedler: Right now we have a task tree with a synchronous child (stream flows) and an asynchronous child (keepalive). That is more complicated than necessary.

We should let let the type system keep track of them. That is, the parent recv's a slice of callbacks (Each callback is like: `func (ctx, lg) error`, runs indefinitely). *TANGENT1*. The parent would then wrap & dispatch all children in the goroutines. Wrap how? It would check context then do a unit of work in the callback. If any of them failed, the wrapper would cancel the parent context (thus terminating all siblings, and returning "control" back to the parent, which can then report the failure & get restarted by `expBackoff`)

I don't think that the wrapper + compatible children would be particularly easy to write, tho... It's easier to just have the children do more work:
* cancel the context if they fail
* read the context

But instead what we have is:
* One synchronous function that forgot to read the context because it had no siblings until I added one and forgot to add this
* One asynchronous function that relies on its parent to cancel the context

example:
```
	go func() {
		err := sm.StreamKeepalives(ciliumCtx, logger, keepalivePeriod, STREAM_NETWORK_FLOWS)
		if err != nil {
			logger.Error("Failed to send keepalives; canceling stream", zap.Error(err))
		}
		ciliumCancel()
	}()
```

*TANGENT1* Unfortunately the go type system doesn't have a way to spell "runs indefinitely", so I actually think we should come up with a naming scheme. Perhaps adding `indefinitely` to the end of the function name, or maybe something less intrusive, like an `_` or an `88` for infinity & also good luck.

evanj80-illumio: Yeah I like the `indefinitely ` suffix for functions that are meant to run forever. I agree the keepalives and flow stream make this complicated to managed and also catch issues like this. We should for sure have some time allocated to take a look how this is structured and maybe rewrite and simplify this so we dont miss things like this.

evanj80-illumio (approved):
Make sure you test the enable/disable now and as long as this works LGTM

internal/controller/streams.go
```diff
@@ -136,6 +136,7 @@ func ServerIsHealthy() bool {
 
 // StreamResources handles the resource stream.
 func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger, cancel context.CancelFunc) error {
+	defer cancel()
```
rlenglet, line 139: This is unnecessary. The caller already does that here:
https://github.com/illumio/cloud-operator/blob/d0075cbd712ad3808e64be909e9fc28040ea6d8c/internal/controller/streams.go#L455

PR #180, opened 2025-04-16
Serialize list and watch logic

evanj80-illumio:
Avoid issues with k8s API servers rate limiting us by performing the list in listAndWatch in serial instead of parallel.
When cloud-operator starts up, it invokes a list and watch. That is, it gets an initial snapshot of all k8s resources we care about (list), then watches them for any changes (watch). We currently list all resources in parallel. This causes us to get rate limited. So we will switch to list in serial.

Also patching go net vulnerability


internal/controller/resource_manager.go
```diff
@@ -31,38 +30,39 @@ type ResourceManager struct {
...
+		return "", err
 	}
-	allResourcesSnapshotted.Done()
-	snapshotCompleted.Wait()
+	return resourceVersion, nil
+}
+
+// WatchK8sWorkloads initiates a watch stream for the specified Kubernetes resource starting from the given resourceVersion.
+// This function blocks until the watch ends or the context is canceled.
+func (r *ResourceManager) WatchK8sWorkloads(ctx context.Context, cancel context.CancelFunc, resource string, apiGroup string, resourceVersion string) {
```
rlenglet, line 46:
```suggestion
func (r *ResourceManager) WatchK8sResources(ctx context.Context, cancel context.CancelFunc, resource string, apiGroup string, resourceVersion string) {
```

internal/controller/resource_manager.go
```diff
@@ -31,38 +30,39 @@ type ResourceManager struct {
...
-// dynamically, managing context cancellation and synchronization with wait
-// groups. As long as we are able to watch and stream, we will never return from
-// this function
-func (r *ResourceManager) DynamicListAndWatchResources(ctx context.Context, cancel context.CancelFunc, resource string, apiGroup string, allResourcesSnapshotted *sync.WaitGroup, snapshotCompleted *sync.WaitGroup) {
-	defer cancel()
-
-	resourceListVersion, err := r.DynamicListResources(ctx, r.logger, resource, apiGroup)
+// ListCurrentK8sWorkloads lists the current state of a specified Kubernetes resource
+// using the dynamic client, and returns its latest resourceVersion.
+func (r *ResourceManager) ListCurrentK8sWorkloads(ctx context.Context, resource string, apiGroup string) (string, error) {
```
rlenglet, line 35:
```suggestion
func (r *ResourceManager) ListK8sResources(ctx context.Context, resource string, apiGroup string) (string, error) {
```

internal/controller/resource_manager.go
```diff
@@ -31,38 +30,39 @@ type ResourceManager struct {
...
-	resourceListVersion, err := r.DynamicListResources(ctx, r.logger, resource, apiGroup)
+// ListK8sResources lists the current state of a specified Kubernetes resource
+// using the dynamic client, and returns its latest resourceVersion.
+func (r *ResourceManager) ListK8sResources(ctx context.Context, resource string, apiGroup string) (string, error) {
+	resourceVersion, err := r.DynamicListResources(ctx, r.logger, resource, apiGroup)
 	if err != nil {
-		allResourcesSnapshotted.Done()
-		r.logger.Error("Unable to list resources", zap.Error(err))
-		return
+		r.logger.Error("Failed to list resource", zap.String("resource", resource), zap.Error(err))
```
rlenglet, line 38: All this method does is wrap `DynamicListResources` to log the error if it's not `nil`.
That error is logged by the caller with the exact same log message here: https://github.com/illumio/cloud-operator/pull/180/files#diff-65d4066e2022c168a97a2083474d348fa6d5d1fffb6e8d7378282935ad187940R204

So this `ListK8sResources` function is unnecessary. Just have the caller call `DynamicListResources` directly.

internal/controller/resource_manager.go
```diff
@@ -31,38 +30,39 @@ type ResourceManager struct {
...
-	err = limiter.Wait(ctx)
+	err := limiter.Wait(ctx)
 	if err != nil {
 		r.logger.Error("Cannot wait using rate limiter", zap.Error(err))
 		return
 	}
 
 	err = r.watchEvents(ctx, resource, apiGroup, watchOptions)
 	if err != nil {
+		r.logger.Error("Watch failed", zap.String("resource", resource), zap.Error(err))
```
rlenglet, line 65: Instead of adding this `resource` field almost everywhere, pass in the `Logger` as a parameter to this method, and set this `resource` field in the caller before passing it in.
??

internal/controller/resource_manager.go
```diff
@@ -31,38 +30,39 @@ type ResourceManager struct {
...
 	watchOptions := metav1.ListOptions{
 		Watch:           true,
-		ResourceVersion: resourceListVersion,
+		ResourceVersion: resourceVersion,
 	}
 
 	// Prevent us from overwhelming K8 api
 	limiter := rate.NewLimiter(1, 5)
-	err = limiter.Wait(ctx)
+	err := limiter.Wait(ctx)
```
rlenglet, line 57: What is the point of this???
We create a rate limiter to just get exactly _one_ token from it, which in practice should never ever block?
So what is this rate limiter limiting?

internal/controller/streams.go
```diff
@@ -196,12 +193,38 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
 		return err
 	}
+
+	resourceVersions := make(map[string]string)
 	for resource, apiGroup := range resourceAPIGroupMap {
-		allResourcesSnapshotted.Add(1)
-		go resourceLister.DynamicListAndWatchResources(ctx, cancel, resource, apiGroup, &allResourcesSnapshotted, &snapshotCompleted)
+		// Add jitter between list resources
+		jitter := time.Duration(rand.Intn(1000)+1000) * time.Millisecond // 1-2 s
+		time.Sleep(jitter)
```
rlenglet, line 201: 1. This sleep is too long.
1. I think this sleep is unnecessary.
1. You shouldn't sleep while ignoring `ctx`. If we really needed to sleep here, this should be using a `Timer` and a select with another case to check `ctx.Done()`.

Remove this and check that it's alright:
```suggestion
```

internal/controller/streams.go
```diff
@@ -196,12 +193,38 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
+		version, ok := resourceVersions[resource]
+		if !ok {
+			logger.Warn("Skipping watch: missing resourceVersion", zap.String("resource", resource))
+			return err
+		}
+
+		go func(resource, apiGroup, version string) {
+			// Add jitter between watcher startups
+			jitter := time.Duration(rand.Intn(1000)+1000) * time.Millisecond // 1-2 s
+			time.Sleep(jitter)
```
rlenglet, line 223: Same as above.
```suggestion
```
If you are worried about doing too many concurrent requests, you could use other mechanisms to serialize the watcher creations that would be better than sleeping for random durations.

internal/controller/streams.go
```diff
@@ -196,12 +193,38 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
 		logger.Error("Failed to send cluster metadata", zap.Error(err))
 		return err
 	}
+
+	resourceVersions := make(map[string]string)
 	for resource, apiGroup := range resourceAPIGroupMap {
```
rlenglet, line 198: In each loop, you should check `ctx.Done()` for every iteration.
```suggestion
	for resource, apiGroup := range resourceAPIGroupMap {
		select {
		case <-ctx.Done():
			return ctx.Err()
		default:
		}
```

internal/controller/streams.go
```diff
@@ -196,12 +193,38 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
+			return err
+		}
+		resourceVersions[resource] = resourceVersion
 	}
-	allResourcesSnapshotted.Wait()
+
 	err = sm.sendResourceSnapshotComplete(logger)
+	logger.Info("cloud-operator has succesfully ingested and sent a snapshot of k8s cluster")
+
+	for resource, apiGroup := range resourceAPIGroupMap {
```
rlenglet, line 213: In each loop, you should check `ctx.Done()` for every iteration.
```suggestion
	for resource, apiGroup := range resourceAPIGroupMap {
		select {
		case <-ctx.Done():
			return ctx.Err()
		default:
		}
```

rlenglet (changes requested):
I think the random sleeps are unnecessary for the list operations, since they are now always serialized.
The watcher creations are now still concurrent. But if we're worried about that, I think the random sleeps there should be replaced either by:
1) by creating the watchers serially in the loop before creating each goroutine to read from it, or
2) enforcing that only one watcher is created at a time, e.g. by using a shared `Mutex`.

internal/controller/resource_manager.go
AriSweedler: I believe we have a single resource manager that gets used for ALL resources. The resource manager contains a logger.

Based on the name, I think it may make sense to create a resource manager per resource (even if they share k8s handles like `clientset` - that field is a ptr anyway so that's fine). `NewResourceManager(...)` can require the `resource` field (as well as ptrs to clientset and dynamicClient and ctx and anything else necessary). The "c'tor" automatically add the `zap.String("resource", resource)` field onto the resource manager.

internal/controller/resource_manager.go
AriSweedler: I believe we have a single resource manager that gets used for ALL resources. The resource manager contains a logger.

Based on the name, I think it may make sense to create a resource manager per resource (even if they share k8s handles like clientset - that field is a ptr anyway so that's fine). NewResourceManager(...) can require the resource field (as well as ptrs to clientset and dynamicClient and ctx and anything else necessary). The "c'tor" automatically add the zap.String("resource", resource) field onto the resource manager.

AriSweedler (dismissed):

internal/controller/resource_manager.go
evanj80-illumio: The point is that since this was happening concurrently we could have 10 go routines all trying to grab a limiter. So yeah this is not useful anymore. Will take it out good catch.

rlenglet: This won't work until a single limiter is shared among all the goroutines.
Right now, this creates one limiter per goroutine, which defeats its purpose.

internal/controller/resource_manager.go
rlenglet: > The resource manager contains a logger.

That's a problem. Each `Logger` has a specific context. I consider them like `Context`s, they should be passed as function parameters like `Context`s.

evanj80-illumio: Will update this !

internal/controller/resource_manager.go
evanj80-illumio: Okay have created a shared one which should work as it was intended

internal/controller/streams.go
```diff
@@ -196,21 +198,48 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
 		logger.Error("Failed to send cluster metadata", zap.Error(err))
 		return err
 	}
+	var allWatchInfos []watcherInfo
```
rlenglet, line 203: Pre-allocate the slice.
```suggestion

	allWatchInfos := make([]watcherInfo, 0, len(resourceAPIGroupMap))
```

internal/controller/streams.go
```diff
@@ -196,21 +198,48 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
+
+	// PHASE 1: List all resources
 	for resource, apiGroup := range resourceAPIGroupMap {
```
rlenglet, line 206: Iterate over the resources in a deterministic order? This may make troubleshooting easier?
```suggestion
	for _, resource := slices.Sorted(maps.Keys(resourceAPIGroupMap)) {
		apiGroup := resourceAPIGroupMap[resource]
```

internal/controller/streams.go
```diff
@@ -196,21 +198,48 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
 	}
-	allResourcesSnapshotted.Wait()
+
+	// PHASE 2: Send snapshot complete
 	err = sm.sendResourceSnapshotComplete(logger)
+	if err != nil {
+		resourceLister.logger.Error("Failed to send snapshot complete", zap.Error(err))
+		return err
+	}
+	logger.Info("cloud-operator has successfully ingested and sent a snapshot of k8s cluster")
```
rlenglet, line 232: We know that this is cloud-operator. No need to repeat it.
```suggestion
	logger.Info("Successfully sent resource snapshot")
```

internal/controller/streams.go
```diff
@@ -196,21 +198,48 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
+
+	// PHASE 3: Start watchers concurrently
+	for _, info := range allWatchInfos {
+		go func(info watcherInfo) {
+			resourceLister.WatchK8sResources(ctx, cancel, info.resource, info.apiGroup, info.resourceVersion)
+		}(info)
+	}
+
 	dd.timeStarted = time.Now()
```
rlenglet, line 241: I don't understand why we're setting `timeStarted` twice (the first time is here: https://github.com/illumio/cloud-operator/pull/180/files#diff-65d4066e2022c168a97a2083474d348fa6d5d1fffb6e8d7378282935ad187940R187) and why we're not locking before setting it the second time here??

I think setting it a second time is unnecessary given this is the only place where we're reading it:
https://github.com/illumio/cloud-operator/blob/3d70473f350d4245e7e7efa5031c89c4560ded24/internal/controller/streams.go#L131 (it's ignored if `processingResources` is `false`).

rlenglet (changes requested):
Looks great, more relatively minor comments.

AriSweedler (approved):

internal/controller/resource_manager.go
evanj80-illumio: Okay looking back at this I will note down this as a separate PR to make a resourceManager per resource in order to clean up this logic. Think its out of scope for this PR.

internal/controller/streams.go
evanj80-illumio: Yeah this should not be getting set here.

evanj80-illumio: Seems like the diff is off, one was added but does not seem to show that

internal/controller/resource_manager.go
rlenglet: SGTM.

evanj80-illumio:
> Doc comments:
> 
> * rename PR to `serialize list and watch logic` instead of `streamline`
> * Add a description, here's what I'd say (feel free to use verbatim or edit or completely rewrite)
> 
> > Avoid issues with k8s API servers rate limiting us by performing the `list` in `listAndWatch` in serial instead of parallel.
> > When `cloud-operator` starts up, it invokes a `list and watch`. That is, it gets an initial snapshot of all k8s resources we care about (`list`), then watches them for any changes (`watch`). We currently `list` all resources in parallel. This causes us to get rate limited. So we will switch to `list` in serial.

Im going to use this!

rlenglet (approved):
LGTM, thanks!

PR #183, opened 2025-04-17
Test coverage for cloud operator

pavankumarinnamuri:


evanj80-illumio (changes requested):

evanj80-illumio:
Also new tests that were created are failing so please fix that

internal/controller/onboarding_test.go
```diff
@@ -75,6 +106,60 @@ func (suite *ControllerTestSuite) TestOnboard() {
...
+			serverHandler: func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+				_, err := w.Write([]byte("invalid json"))
+				if err != nil {
+					suite.T().Fatal("Failed to write response in invalid-json-response test: " + err.Error())
+				}
+			},
+			requestURL:     "http://example.com",
+			expectedError:  true,
+			expectedErrMsg: "invalid character 'i' looking for beginning of value",
```
evanj80-illumio, line 151: I think this is NOT a good thing to test. I think maybe its better to fix this to be a more graceful error. Can you create a ticket to fix this test case in the future?

internal/controller/onboarding_test.go
```diff
@@ -167,22 +252,15 @@ func (suite *ControllerTestSuite) TestGetClusterID() {
...
 	clusterID, err := GetClusterID(ctx, logger)
 	assert.NoError(suite.T(), err)
 	assert.NotEmpty(suite.T(), clusterID)
 
-	// Check if the returned UID matches the expected UID
-	assert.Equal(suite.T(), expectedUID, clusterID)
-
-	// Log the cluster ID for verification
-	suite.T().Logf("Cluster ID: %s", clusterID)
+	// Test error case - invalid context
```
evanj80-illumio, line 260: I think leaving the successful test case how it was is the right thing here. We want to not only check that getClusterID can be called without error but we also want to make sure its returning the correct ID. Thus, thats why we had the expectedUID value

internal/controller/streams_retry_test.go
evanj80-illumio, line 1: @AriSweedler please look this over as well since your are the EXPERT 😎

evanj80-illumio (changes requested):
Looking great so far. A few comments

evanj80-illumio:
Please check the go test GHA and make sure all tests here are passing. Then once that is green please fill in what test coverage has been achieved. I think no need to add anymore tests into this PR.

internal/controller/authenticator_test.go
```diff
@@ -312,3 +313,135 @@ func (suite *ControllerTestSuite) TestParseToken() {
...
+			expectedError: "Required field not found in k8s Secret | field='client_secret'",
+			isTargetError: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := NewCredentialNotFoundInK8sSecretError(tt.field)
+
+			// Test error message
```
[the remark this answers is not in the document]
evanj80-illumio: Agree!

evanj80-illumio (approved):
LGTM just address @AriSweedler comment

evanj80-illumio:
Please merge when conflicts and all tests are passing

evanj80-illumio (approved):

PR #185, opened 2025-04-18
Aggregate collected flows before exporting

evanj80-illumio:


internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,255 @@
...
+	"container/list"
+	"context"
+	"fmt"
+	"time"
+
+	pb "github.com/illumio/cloud-operator/api/illumio/cloud/k8sclustersync/v1"
+	"go.uber.org/zap"
+)
+
+type FlowKey struct {
```
AriSweedler, line 13: `FlowKey` is different per CNI. I don't think that a single type `FlowKey` is correct. I *think* we may want this to be an interface, implemented by `FlowKeyFalco`, `FlowKeyCilium`, etc. Perhaps the flow key will be similar enough for all the CNIs that that is not worthwhile.

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,255 @@
...
+			now := time.Now().UTC()
+			expiredCutoff := now.Add(-activeTimeout)
+
+			for sm.FlowCache.queue.Len() > 0 {
+				frontElem := sm.FlowCache.queue.Front()
+				var flowTimestamp time.Time
+				var flowKey FlowKey
+				var err error
+
+				switch f := frontElem.Value.(type) {
```
AriSweedler, line 66: Building off of https://github.com/illumio/cloud-operator/pull/185/files#r2051125014, this `switch` could be replaced with dynamic dispatch (method calls).

To make this easy, the output of `sm.FlowCache.queue.Front()` would simply need to get piped into a factory.

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,255 @@
...
+				}
+
+				sm.FlowCache.queue.Remove(frontElem)
+				delete(sm.FlowCache.cache, flowKey)
+				sm.sendNetworkFlowRequest(logger, frontElem.Value)
+			}
+
+			// Reset timer for next expiration
+			if front := sm.FlowCache.queue.Front(); front != nil {
+				switch f := front.Value.(type) {
```
AriSweedler, line 94: https://github.com/illumio/cloud-operator/pull/185/files#r2051125014 Same with this `switch`. Just a single line when u have an interface

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,255 @@
...
+
+func resetTimer(timer *time.Timer, timestamp time.Time, timeout time.Duration) {
+	delay := timestamp.Add(timeout).Sub(time.Now())
+	if delay <= 0 {
+		delay = time.Second // fallback to a short delay
+	}
+	timer.Reset(delay)
+}
+
+func (sm *streamManager) createFlowKey(flow interface{}) (FlowKey, error) {
```
AriSweedler, line 159: dynamic dispatch --> use interfaces

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,255 @@
...
+	DestinationEndpoint string
+}
+
+type FlowCache struct {
+	cache      map[FlowKey]*list.Element
+	queue      *list.List
+	bufferSize int
+}
+
+func (sm *streamManager) cacheManagerIndefinitely(ctx context.Context, logger *zap.Logger) error {
```
AriSweedler, line 42: Add a nice doc comment here:

```
cacheManagerIndefinitely aggregates flows by `flowKey` in order to decrease the burden of processing flows downstream
```

AriSweedler (changes requested):

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+	bufferSize int
+}
+
+// cacheManagerIndefinitely manages the flow cache by evicting expired flows based on the active timeout,
+// processing new flows, and resetting the timer for the next expiration.
+func (sm *streamManager) cacheManagerIndefinitely(ctx context.Context, logger *zap.Logger) error {
+	const activeTimeout = 20 * time.Second // Define the active timeout period for flows. TODO: make configurable
+
+	// Create a new timer to trigger every activeTimeout duration.
+	timer := time.NewTimer(activeTimeout)
```
rlenglet, line 46:
```suggestion
	timer := time.NewTimer(c.activeTimeout)
```

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-ctx.Done(): // If the context is canceled, exit the loop and return the error.
+			return ctx.Err()
+
+		case <-timer.C:
+			now := time.Now().UTC()
+			expiredCutoff := now.Add(-activeTimeout) // Determine the cutoff time for expired flows.
```
rlenglet, line 56:
```suggestion
			activeTimeoutCutoff := now.Sub(activeTimeout) // Determine the cutoff time for expired flows.
```

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+				frontElem := sm.FlowCache.queue.Front()
+				var flowTimestamp time.Time
+				var flowKey FlowKey
+				var err error
+
+				// Check the type of flow and get the flow timestamp and flow key.
+				switch f := frontElem.Value.(type) {
+				case *pb.FalcoFlow:
+					flowTimestamp = f.GetTimestamp().AsTime()
+					flowKey, err = sm.createFlowKey(f)
```
rlenglet, line 69: The flow key should be computed once and for all.
You shouldn't have to re-compute it every time.

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+				// If the flow is not expired (timestamp is newer than the cutoff), stop evicting.
+				if flowTimestamp.After(expiredCutoff) {
+					break
+				}
+
+				// Evict the expired flow by removing it from the cache and queue.
+				sm.FlowCache.queue.Remove(frontElem)
+				delete(sm.FlowCache.cache, flowKey)
+
+				sm.sendNetworkFlowRequest(logger, frontElem.Value)
```
rlenglet, line 89: Whatever is blocking sending a flow out must always take the Context to return immediately when it's canceled.
```suggestion
				select {
				case <-ctx.Done():
					return ctx.Err()
				case c.outFlows <- frontElem.Value.(Flow):
				}
```

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+				// Check the type of flow and get the flow timestamp and flow key.
+				switch f := frontElem.Value.(type) {
+				case *pb.FalcoFlow:
+					flowTimestamp = f.GetTimestamp().AsTime()
+					flowKey, err = sm.createFlowKey(f)
+				case *pb.CiliumFlow:
+					flowTimestamp = f.GetTime().AsTime()
+					flowKey, err = sm.createFlowKey(f)
+				default:
+					break
```
rlenglet, line 74: That's not right. You should panic in this case.
Or better, define a common struct or interface type to avoid this situation.

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+
+			// Reset the timer to expire based on the next flow's timestamp.
+			if front := sm.FlowCache.queue.Front(); front != nil {
+				switch f := front.Value.(type) {
+				case *pb.FalcoFlow:
+					resetTimer(timer, f.GetTimestamp().AsTime(), activeTimeout)
+				case *pb.CiliumFlow:
+					resetTimer(timer, f.GetTime().AsTime(), activeTimeout)
+				default:
+					timer.Reset(activeTimeout)
```
rlenglet, line 100: That's not right. You should panic in this case.
Or better, define a common struct or interface type to avoid this situation.

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+				}
+			} else {
+				// If there are no flows left, reset the timer to the default activeTimeout period.
+				timer.Reset(activeTimeout)
```
rlenglet, line 104: If there are no flows cached, there should be no timer set.
```suggestion
```

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+			}
+
+		case flow := <-sm.flowChannel:
```
rlenglet, line 107:
```suggestion
		case flow := <-c.inFlows:
```

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+			flowKey, err := sm.createFlowKey(flow)
+			if err != nil {
+				return err
+			}
+
+			// If the flow already exists in the cache, skip adding it again.
+			if _, found := sm.FlowCache.cache[flowKey]; found {
+				continue
```
rlenglet, line 115: You need to increment the flow count, and potentially the byte count (which we'll get in OVN IPFIX flows).
If you don't want to do that in this PR, at the minimum add a big fat TODO comment to remember to do that later.

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+			}
+
+			// If the cache is full, evict the oldest flow to make room for the new flow.
+			// I think bufferSize should be renamed to maxBufferSize or something.
```
rlenglet, line 119: I proposed `maxFlows`.
```suggestion
```

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+				// Process the evicted flow.
+				switch old := oldestElem.Value.(type) {
+				case *pb.FalcoFlow, *pb.CiliumFlow:
+					oldKey, err := sm.createFlowKey(old)
+					if err == nil {
+						delete(sm.FlowCache.cache, oldKey)
+						sm.sendNetworkFlowRequest(logger, old)
+					}
+				}
+			}
```
rlenglet, line 133:
```suggestion
			if len(c.cache) >= c.maxFlows {
				oldestFlow := c.queue.Front()
				c.queue.Remove(oldestFlow)
				select {
				case <-ctx.Done():
					return ctx.Err()
				case c.outFlows <- oldestFlow.Value.(Flow):
				}
			}
```

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+			flowElem := sm.FlowCache.queue.PushBack(flow)
+			sm.FlowCache.cache[flowKey] = flowElem
+
+			// Reset the timer based on the new flow's timestamp.
+			switch f := flow.(type) {
+			case *pb.FalcoFlow:
+				resetTimer(timer, f.GetTimestamp().AsTime(), activeTimeout)
+			case *pb.CiliumFlow:
+				resetTimer(timer, f.GetTime().AsTime(), activeTimeout)
+			}
```
rlenglet, line 145: That's incorrect.
The timer must always be set based on the front/oldest flow's timestamp, not the most recent like you do here.

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+	}
+}
+
+func resetTimer(timer *time.Timer, timestamp time.Time, timeout time.Duration) {
+	delay := timestamp.Add(timeout).Sub(time.Now())
+	// Not sure if I need this
+	// if delay <= 0 {
+	// 	delay = time.Second
+	// }
```
rlenglet, line 155: No, you don't need this. Because if the new duration is negative it means that the flow has already expired, e.g. because we took too much time sending flows, etc.
That's ok, the timer will just fire immediately in this case, which is what we want. I quickly tested that.
```suggestion
```

internal/controller/streams.go
```diff
@@ -389,11 +394,7 @@ func (sm *streamManager) StreamFalcoNetworkFlows(ctx context.Context, logger *za
 				logger.Error("Failed to parse Falco event into flow", zap.Error(err))
 				return err
 			}
-			err = sm.sendNetworkFlowRequest(logger, convertedFalcoFlow)
-			if err != nil {
-				logger.Error("Failed to send Falco flow", zap.Error(err))
-				return err
-			}
+			sm.flowChannel <- convertedFalcoFlow
```
rlenglet, line 397: Never do a blocking write into a channel ignoring `ctx`.
Just wrap that into a method of `FlowCache` as I suggested above.
```suggestion
			sm.flowCache.CacheFlow(ctx, convertedFalcoFlow)
```

internal/controller/streams.go
```diff
@@ -707,6 +707,12 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
 			sm := &streamManager{
 				streamClient:       streamClient,
 				bufferedGrpcSyncer: bufferedGrpcSyncer,
+				FlowCache: FlowCache{
+					cache:      make(map[FlowKey]*list.Element),
+					queue:      list.New(),
+					bufferSize: 100,
+				},
+				flowChannel: make(chan interface{}, int(100)),
```
rlenglet, line 715: 100 is too large a channel IMHO.
```suggestion
				FlowCache: NewFlowCache(
					20*time.Second, // TODO: Make the active timeout configurable.
					1000, // TODO: Make the maxFlows capacity configurable.
					make(chan Flow, 10),
				)
```

internal/controller/streams.go
```diff
@@ -389,11 +394,7 @@ func (sm *streamManager) StreamFalcoNetworkFlows(ctx context.Context, logger *za
 				logger.Error("Failed to parse Falco event into flow", zap.Error(err))
 				return err
 			}
-			err = sm.sendNetworkFlowRequest(logger, convertedFalcoFlow)
-			if err != nil {
-				logger.Error("Failed to send Falco flow", zap.Error(err))
-				return err
-			}
+			sm.flowChannel <- convertedFalcoFlow
 		}
```
rlenglet, line 398: With my proposal, this now must read from the cache's `outFlows` and actually send those flows.

One way to do it is here, in the same `select`:
```suggestion
		case falcoFlow <-c.flowCache.outFlows:
			err = sm.sendNetworkFlowRequest(logger, falcoFlow)
			if err != nil {
				logger.Error("Failed to send Falco flow", zap.Error(err))
				return err
			}
		}
```
But that won't work for Cilium, since the body of the `for` loop in `exportCiliumFlows` is not one `select` like here.

So what I would suggest instead is that `streamManager` starts a separate goroutine to read flows from `outFlows` and send them. That would work for both Cilium and Falco.

internal/controller/flow_cilium.go
```diff
@@ -187,11 +187,7 @@ func (fm *CiliumFlowCollector) exportCiliumFlows(ctx context.Context, sm *stream
 		if ciliumFlow == nil {
 			continue
 		}
-		err = sm.sendNetworkFlowRequest(fm.logger, ciliumFlow)
-		if err != nil {
-			fm.logger.Error("Cannot send cilium flow", zap.Error(err))
-			return err
-		}
+		sm.flowChannel <- ciliumFlow
```
rlenglet, line 190: Never do a blocking write into a channel ignoring `ctx`.
Just wrap that into a method of `FlowCache` as I suggested above.
```suggestion
		sm.flowCache.CacheFlow(ctx, ciliumFlow)
```

rlenglet (changes requested):

internal/controller/flow_aggregation.go
```diff
@@ -0,0 +1,270 @@
...
+	SourceEndpoint string
+	// DestinationEndpoint contains k8s metadata for destination endpoint
+	DestinationEndpoint string
+}
+
+type FlowCache struct {
+	cache      map[FlowKey]*list.Element
+	queue      *list.List
+	bufferSize int
+}
```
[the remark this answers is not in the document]
AriSweedler: My intuition states that setting `inFlows: make(chan Flow, 1)` to have a buffer size of `1` will create a bottleneck... Let's say the producer is marginally faster than the consumer. Even if they are both active, when trying to produce we will continually block and context switch out. The producer will immediately context switch back in when it's available, but we still pay the price of sleep and wake. Perhaps even changing the CPU we run on (invalidating cache).

We could make this less common by setting this to 10 or even 100. I don't believe it would cost too much memory.

internal/controller/flow_aggregation.go
AriSweedler: Please add a comment deliberately stating something like: `Don't override the flowCache's activeTimeout when newer flows of the same key come in. It is an acceptable tradeoff to lose some amount of compression to prevent starvation`

internal/controller/flow_aggregation.go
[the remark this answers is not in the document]
AriSweedler: Just saw your comment: https://github.com/illumio/cloud-operator/pull/185#discussion_r2051213950 `100 is too large for a channel` & wanted to include it in this conversation. Looks like my suggestion of `100` is bad. Is `1` really ok?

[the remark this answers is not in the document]
rlenglet: Why does it matter? All flows must be serialized anyway? What would be the point of queuing them at the input to the aggregator?

internal/controller/streams.go
evanj80-illumio: In order for streamManager to have access to that chan shall it also now include the flowcache object?

internal/controller/flow_aggregation.go
evanj80-illumio: Okay moved the flow key logic around to make this better

internal/controller/flow_cache.go
```diff
@@ -0,0 +1,300 @@
...
+					resetTimer(timer, f.GetTimestamp().AsTime(), c.activeTimeout)
+				case *pb.CiliumFlow:
+					resetTimer(timer, f.GetTime().AsTime(), c.activeTimeout)
+				default:
+					return errors.New("Non supported flow.")
+				}
+			}
+		case flow := <-c.inFlows:
+			// If the flow already exists in the cache, skip adding it again.
+			if _, found := c.cache[flow.Key]; found {
```
rlenglet, line 155:
```suggestion
			if _, found := c.cache[flow.Key()]; found {
```

internal/controller/flow_cache.go
```diff
@@ -0,0 +1,300 @@
...
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case c.outFlows <- oldestFlow.Value.(Flow):
+				}
+			}
+
+			// Add the new flow to the cache and queue.
+			c.queue.PushBack(flow)
+			c.cache[flow.Key] = flow
```
rlenglet, line 172:
```suggestion
			c.cache[flow.Key()] = flow
```

internal/controller/flow_cache.go
```diff
@@ -0,0 +1,300 @@
...
+
+			oldestFlow := c.queue.Front().Value.(Flow)
+			// Reset the timer based on the oldest flow's timestamp (queue is ordered so the front element is oldest).
+			switch f := oldestFlow.rawFlow.(type) {
+			case *pb.FalcoFlow:
+				resetTimer(timer, f.GetTimestamp().AsTime(), c.activeTimeout)
+			case *pb.CiliumFlow:
+				resetTimer(timer, f.GetTime().AsTime(), c.activeTimeout)
+			}
```
rlenglet, line 181:
```suggestion
			// Reset the timer based on the oldest flow's timestamp (queue is ordered so the front element is oldest).
			resetTimer(timer, c.queue.Front().Value.(Flow).StartTimestamp(), c.activeTimeout)
```

internal/controller/flow_cache.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+	}
+}
+
+func resetTimer(timer *time.Timer, timestamp time.Time, timeout time.Duration) {
+	delay := timestamp.Add(timeout).Sub(time.Now())
```
rlenglet, line 187: GitHub suggests:
```suggestion
	delay := time.Until(timestamp.Add(timeout))
```

internal/controller/flow_cache.go
```diff
@@ -0,0 +1,300 @@
...
+
+	return FlowKeyFalco{
+		Ts:              timestamp,
+		SourceIP:        srcIP,
+		DestinationIP:   dstIP,
+		SourcePort:      srcPort,
+		DestinationPort: dstPort,
+		Protocol:        proto,
+	}
+}
```
rlenglet, line 300: Remove.
```suggestion
```

Just have both `pb.FalcoFlow` and `pb.CiliumFlow` implement the `Flow` interface.
Do that by adding new files **_into `api/illumio/cloud/k8sclustersync/v1`_**:

`flow_falco.go`:
```
//...

var _ Flow = &FalcoFlow{}

func (f *FalcoFlow) StartTimestamp() time.Time {
	return f.GetTimestamp().AsTime()
}

type FalcoFlowKey struct {
	SourceIP           string
	DestinationIP      string
	SourcePort         int
	DestinationPort    int
	Protocol           string
}

func (f *FalcoFlow) Key() any {
	if flow == nil {
		return nil
	}

	srcIP := flow.GetLayer3().GetSource()
	dstIP := flow.GetLayer3().GetDestination()

	var (
		srcPort int
		dstPort int
		proto   string
	)

	switch l4 := flow.GetLayer4().GetProtocol().(type) {
	case *pb.Layer4_Tcp:
		srcPort = int(l4.Tcp.GetSourcePort())
		dstPort = int(l4.Tcp.GetDestinationPort())
		proto = "TCP"
	case *pb.Layer4_Udp:
		srcPort = int(l4.Udp.GetSourcePort())
		dstPort = int(l4.Udp.GetDestinationPort())
		proto = "UDP"
	case *pb.Layer4_Sctp:
		srcPort = int(l4.Sctp.GetSourcePort())
		dstPort = int(l4.Sctp.GetDestinationPort())
		proto = "SCTP"
	case *pb.Layer4_Icmpv4:
		proto = "ICMPv4"
	case *pb.Layer4_Icmpv6:
		proto = "ICMPv6"
	default:
		proto = "UNKNOWN"
	}

	return &FalcoFlowKey{
		SourceIP:        srcIP,
		DestinationIP:   dstIP,
		SourcePort:      srcPort,
		DestinationPort: dstPort,
		Protocol:        proto,
	}
}
```

`flow_cilium.go`:
```
//...

var _ Flow = &CiliumFlow{}

func (f *CiliumFlow) StartTimestamp() time.Time {
	return f.GetTime().AsTime()
}

type CiliumFlowKey struct {
	SourceIP           string
	DestinationIP      string
	SourcePort         int
	DestinationPort    int
	Protocol           string
	SourceK8sMeta      string
	DestinationK8sMeta string
}

func (f *CiliumFlow) Key() any {
	if flow == nil {
		return nil
	}

	var (
		srcPort   int
		dstPort   int
		proto     string
	)

	// Ports + Protocol
	switch l4 := flow.GetLayer4().GetProtocol().(type) {
	case *pb.Layer4_Tcp:
		srcPort = int(l4.Tcp.GetSourcePort())
		dstPort = int(l4.Tcp.GetDestinationPort())
		proto = "TCP"
	case *pb.Layer4_Udp:
		srcPort = int(l4.Udp.GetSourcePort())
		dstPort = int(l4.Udp.GetDestinationPort())
		proto = "UDP"
	case *pb.Layer4_Sctp:
		srcPort = int(l4.Sctp.GetSourcePort())
		dstPort = int(l4.Sctp.GetDestinationPort())
		proto = "SCTP"
	case *pb.Layer4_Icmpv4:
		proto = "ICMPv4"
	case *pb.Layer4_Icmpv6:
		proto = "ICMPv6"
	default:
		proto = "UNKNOWN"
	}

	return &CiliumFlowKey{
		SourceIP:           flow.Layer3.GetSource(),
		DestinationIP:      flow.Layer3.GetDestination(),
		SourcePort:         srcPort,
		DestinationPort:    dstPort,
		Protocol:           proto,
		SourceK8sMeta:      flow.SourceEndpoint.GetPodName(),
		DestinationK8sMeta: flow.DestinationEndpoint.GetPodName(),
	}
}

```

internal/controller/flow_cilium.go
```diff
@@ -23,6 +23,17 @@ type CiliumFlowCollector struct {
...
+	Ts                 int64
+	SourceIP           string
+	DestinationIP      string
+	SourcePort         int
+	DestinationPort    int
+	Protocol           string
+	SourceK8sMeta      string
+	DestinationK8sMeta string
+}
+
```
rlenglet, line 36: Move this into `api/.../v1`.
```suggestion
```

internal/controller/flow_cilium.go
```diff
@@ -33,6 +44,15 @@ var (
...
 
+func (f FlowKeyCilium) Timestamp() int64            { return f.Ts }
+func (f FlowKeyCilium) SrcIP() string               { return f.SourceIP }
+func (f FlowKeyCilium) DstIP() string               { return f.DestinationIP }
+func (f FlowKeyCilium) SrcPort() int                { return f.SourcePort }
+func (f FlowKeyCilium) DstPort() int                { return f.DestinationPort }
+func (f FlowKeyCilium) Proto() string               { return f.Protocol }
+func (f FlowKeyCilium) SourceEndpoint() string      { return f.SourceK8sMeta }
+func (f FlowKeyCilium) DestinationEndpoint() string { return f.DestinationK8sMeta }
+
```
rlenglet, line 55: Remove.
```suggestion
```

internal/controller/flow_cilium.go
```diff
@@ -187,11 +207,11 @@ func (fm *CiliumFlowCollector) exportCiliumFlows(ctx context.Context, sm *stream
 		if ciliumFlow == nil {
 			continue
 		}
-		err = sm.sendNetworkFlowRequest(fm.logger, ciliumFlow)
+		flowKey, err := sm.FlowCache.createFlowKey(ciliumFlow)
 		if err != nil {
-			fm.logger.Error("Cannot send cilium flow", zap.Error(err))
 			return err
 		}
+		sm.FlowCache.CacheFlow(ctx, Flow{Timestamp: ciliumFlow.Time.AsTime(), Key: flowKey, rawFlow: ciliumFlow})
```
rlenglet, line 214:
```suggestion
		sm.FlowCache.CacheFlow(ctx, ciliumFlow)
```

internal/controller/flow_falco.go
```diff
@@ -30,6 +30,27 @@ type FalcoEvent struct {
...
+// Methods required by FlowKey interface
+func (f FlowKeyFalco) Timestamp() int64            { return f.Ts }
+func (f FlowKeyFalco) SrcIP() string               { return f.SourceIP }
+func (f FlowKeyFalco) DstIP() string               { return f.DestinationIP }
+func (f FlowKeyFalco) SrcPort() int                { return f.SourcePort }
+func (f FlowKeyFalco) DstPort() int                { return f.DestinationPort }
+func (f FlowKeyFalco) Proto() string               { return f.Protocol }
+func (f FlowKeyFalco) SourceEndpoint() string      { return "" }
+func (f FlowKeyFalco) DestinationEndpoint() string { return "" }
+
```
rlenglet, line 53:
```suggestion
```

internal/controller/streams.go
```diff
@@ -389,11 +402,11 @@ func (sm *streamManager) StreamFalcoNetworkFlows(ctx context.Context, logger *za
 				logger.Error("Failed to parse Falco event into flow", zap.Error(err))
 				return err
 			}
-			err = sm.sendNetworkFlowRequest(logger, convertedFalcoFlow)
+			flowKey, err := sm.FlowCache.createFlowKey(convertedFalcoFlow)
 			if err != nil {
-				logger.Error("Failed to send Falco flow", zap.Error(err))
 				return err
 			}
+			sm.FlowCache.CacheFlow(ctx, Flow{Timestamp: convertedFalcoFlow.GetTimestamp().AsTime(), Key: flowKey, rawFlow: convertedFalcoFlow})
```
rlenglet, line 409:
```suggestion
			sm.FlowCache.CacheFlow(ctx, convertedFalcoFlow)
```

internal/controller/streams.go
```diff
@@ -707,6 +719,11 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
 			sm := &streamManager{
 				streamClient:       streamClient,
 				bufferedGrpcSyncer: bufferedGrpcSyncer,
+				FlowCache: NewFlowCache(
+					20*time.Second, // TODO: Make the active timeout configurable.
+					1000,           // TODO: Make the maxFlows capacity configurable.
+					make(chan Flow, 10),
```
rlenglet, line 725: I changed my mind.
Please increase the output flow buffer to 100.
That will reduce the blocking for input flows that hit the cache.
```suggestion
					make(chan Flow, 100),
```

rlenglet (changes requested):
Some comments to clean up FlowCache further to make it oblivious to the types of flows. It should have no direct dependencies to specific Protobuf message types.

internal/controller/flow_cache.go
evanj80-illumio: Ah makes much more sense this way

internal/controller/flow_aggregation.go
rlenglet: Actually, we don't need to define a specific type for the flow key. `FlowCache` doesn't care what's in the flow key, it's opaque to it. So `any` would do just fine. As long as the values are comparable so they can be used as keys in a `map`.

api/illumio/cloud/k8sclustersync/v1/cilium_flow.go
```diff
@@ -0,0 +1,65 @@
...
+
+	return CiliumFlowKey{
+		SourceIP:           flow.Layer3.GetSource(),
+		DestinationIP:      flow.Layer3.GetDestination(),
+		SourcePort:         srcPort,
+		DestinationPort:    dstPort,
+		Protocol:           proto,
+		SourceK8sMeta:      flow.SourceEndpoint.GetPodName(),
+		DestinationK8sMeta: flow.DestinationEndpoint.GetPodName(),
+	}
```
rlenglet: Nit. Instead of defining local variables and then copying them into the struct, you could directly set the fields of the struct.
```suggestion
	key := CiliumFlowKey{
		SourceIP:           flow.Layer3.GetSource(),
		DestinationIP:      flow.Layer3.GetDestination(),
		SourceK8sMeta:      flow.SourceEndpoint.GetPodName(),
		DestinationK8sMeta: flow.DestinationEndpoint.GetPodName(),
	}

	// Ports + Protocol
	switch l4 := flow.GetLayer4().GetProtocol().(type) {
	case *Layer4_Tcp:
		key.SourcePort = int(l4.Tcp.GetSourcePort())
		key.DestinationPort = int(l4.Tcp.GetDestinationPort())
		key.Protocol = "TCP"
	case *Layer4_Udp:
		key.SourcePort = int(l4.Udp.GetSourcePort())
		key.DestinationPort = int(l4.Udp.GetDestinationPort())
		key.Protocol = "UDP"
	case *Layer4_Sctp:
		key.SourcePort = int(l4.Sctp.GetSourcePort())
		key.DestinationPort = int(l4.Sctp.GetDestinationPort())
		key.Protocol = "SCTP"
	case *Layer4_Icmpv4:
		key.Protocol = "ICMPv4"
	case *Layer4_Icmpv6:
		key.Protocol = "ICMPv6"
	default:
		key.Protocol = "UNKNOWN"
	}

	return key
```

api/illumio/cloud/k8sclustersync/v1/falco_flow.go
```diff
@@ -0,0 +1,64 @@
...
+		proto = "UNKNOWN"
+	}
+
+	return FalcoFlowKey{
+		SourceIP:        srcIP,
+		DestinationIP:   dstIP,
+		SourcePort:      srcPort,
+		DestinationPort: dstPort,
+		Protocol:        proto,
+	}
```
rlenglet: Nit. Same here.
```suggestion
	key := FalcoFlowKey{
		SourceIP:        flow.GetLayer3().GetSource(),
		DestinationIP:   flow.GetLayer3().GetDestination(),
	}

	switch l4 := flow.GetLayer4().GetProtocol().(type) {
	case *Layer4_Tcp:
		key.SourcePort = int(l4.Tcp.GetSourcePort())
		key.DestinationPort = int(l4.Tcp.GetDestinationPort())
		key.Protocol = "TCP"
	case *Layer4_Udp:
		key.SourcePort = int(l4.Udp.GetSourcePort())
		key.DestinationPort = int(l4.Udp.GetDestinationPort())
		key.Protocol = "UDP"
	case *Layer4_Sctp:
		key.SourcePort = int(l4.Sctp.GetSourcePort())
		key.DestinationPort = int(l4.Sctp.GetDestinationPort())
		key.Protocol = "SCTP"
	case *Layer4_Icmpv4:
		key.Protocol = "ICMPv4"
	case *Layer4_Icmpv6:
		key.Protocol = "ICMPv6"
	default:
		key.Protocol = "UNKNOWN"
	}

	return key
```

rlenglet (changes requested):

internal/controller/streams.go
```diff
@@ -333,6 +334,20 @@ func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger
 	return nil
 }
 
+func (sm *streamManager) startFlowCacheOutReader(ctx context.Context, logger *zap.Logger) error {
+	logger.Info("START FLOW CACHEOUT READER")
```
rlenglet, line 338:
```suggestion
```

rlenglet (changes requested):
Last batch of nits. Looks great!

internal/controller/cilium_flow_collector.go
```diff
@@ -187,7 +187,11 @@ func (fm *CiliumFlowCollector) exportCiliumFlows(ctx context.Context, sm *stream
 		if ciliumFlow == nil {
 			continue
 		}
-		sm.FlowCache.CacheFlow(ctx, ciliumFlow)
+		err = sm.FlowCache.CacheFlow(ctx, ciliumFlow)
+		if err != nil {
+			fm.logger.Info("Failed to send flow to cache", zap.Error(err))
```
rlenglet, line 192:
```suggestion
			fm.logger.Error("Failed to cache flow", zap.Error(err))
```

internal/controller/streams.go
```diff
@@ -402,7 +406,11 @@ func (sm *streamManager) StreamFalcoNetworkFlows(ctx context.Context, logger *za
 				logger.Error("Failed to parse Falco event into flow", zap.Error(err))
 				return err
 			}
-			sm.FlowCache.CacheFlow(ctx, convertedFalcoFlow)
+			err = sm.FlowCache.CacheFlow(ctx, convertedFalcoFlow)
+			if err != nil {
+				logger.Info("Failed to send flow to cache", zap.Error(err))
```
rlenglet, line 411:
```suggestion
				logger.Error("Failed to cache flow", zap.Error(err))
```

rlenglet (changes requested):
2 nits.

api/illumio/cloud/k8sclustersync/v1/cilium_flow.go
```diff
@@ -0,0 +1,68 @@
...
+	if flow.GetIsReply().GetValue() {
+		return CiliumFlowKey{
+			SourceIP:           flow.Layer3.GetDestination(),
+			DestinationIP:      flow.Layer3.GetSource(),
+			SourcePort:         key.DestinationPort,
+			DestinationPort:    key.SourcePort,
+			Protocol:           key.Protocol,
+			SourceK8sMeta:      flow.DestinationEndpoint.GetPodName(),
+			DestinationK8sMeta: flow.SourceEndpoint.GetPodName(),
+		}
```
rlenglet, line 65: You could simplify this a bit:
```suggestion
		key = CiliumFlowKey{
			SourceIP:           key.DestinationIP,
			DestinationIP:      key.SourceIP,
			SourcePort:         key.DestinationPort,
			DestinationPort:    key.SourcePort,
			Protocol:           key.Protocol,
			SourceK8sMeta:      key.DestinationK8sMeta,
			DestinationK8sMeta: key.SourceK8sMeta,
		}
```

rlenglet (approved):
LGTM, thanks!!

api/illumio/cloud/k8sclustersync/v1/cilium_flow.go
```diff
@@ -0,0 +1,68 @@
...
+func (flow *CiliumFlow) Key() any {
+	if flow == nil {
+		return nil
+	}
+
+	key := CiliumFlowKey{
+		SourceIP:           flow.Layer3.GetSource(),
+		DestinationIP:      flow.Layer3.GetDestination(),
+		SourceK8sMeta:      flow.SourceEndpoint.GetPodName(),
+		DestinationK8sMeta: flow.DestinationEndpoint.GetPodName(),
```
rlenglet, line 32: The UID would have been a better choice. Because the name is not unique: multiple Pods can have the same name in different namespaces at the same time, and when a Pod in a StatefulSet restarts it gets the same name. But in all cases they all get different UIDs, so using UID is better.

rlenglet (changes requested):
You should actually use the UID instead of the name for the k8s workload. Unlike the name, the UID is really unique.

api/illumio/cloud/k8sclustersync/v1/cilium_flow.go
rlenglet: And that will also eliminate the problem with using the resource name instead of the UID, cf. https://github.com/illumio/cloud-operator/pull/185#discussion_r2057105538

rlenglet (approved):
Thanks!

PR #187, opened 2025-04-22
Integrate kubernetes logs

aaronIllumio:
The Golang k8s client library is generating logs that are useful for troubleshooting.
Those logs are currently not captured and streamed to CloudSecure.
The k8s.io/client-go and k8s.io/apimachinery libraries both use github.com/go-logr/logr and k8s.io/klog/v2 for logging. klog is a thin wrapper on top of logr, cf. https://github.com/kubernetes/klog/blob/main/imports.go.

Redirect those logs into cloud-operator's Zap Core logger so that they get captured and streamed along with cloud-operator's logs:

Configure the global logr.Logger that is backing klog by calling [klog.SetLogger](https://github.com/kubernetes/klog/blob/e7125f792ea66a85818cfb45261c9e1acc585344/contextual.go#L42-L61) or [klog.SetLoggerWithOptions](https://github.com/kubernetes/klog/blob/e7125f792ea66a85818cfb45261c9e1acc585344/contextual.go#L63-L81) using Zapr (https://pkg.go.dev/github.com/go-logr/zapr), an implementation of logr.Logger backed by a Zap zap.Logger.


cmd/main.go
```diff
@@ -73,6 +78,24 @@ func main() {
 	logger := controller.NewProductionGRPCLogger(bufferedGrpcSyncer)
 	defer logger.Sync() //nolint:errcheck
 
+	// Convert zap.Logger to logr.Logger
```
AriSweedler, line 81: can we hoist this to a function & have it return `clientset`? or is it necessary to bring all these vars in scope. I could even see this function working better as in a new file.

cmd/main.go
```diff
@@ -142,7 +165,18 @@ func main() {
 		zap.Duration("stream_success_period_auth", envConfig.StreamSuccessPeriod.Auth),
 	)
 
-	// Start the gops agent
+	// List pods from default namespace
```
AriSweedler, line 168: same w/ this - hoist this to a function. this fxn is long already. And this is just a helper thunk to show that its working, no need to inline it

AriSweedler (dismissed):
LGTM - nicely done

cmd/main.go
```diff
@@ -142,7 +165,18 @@ func main() {
...
+	if err != nil {
+		klog.Error(err, "Failed to list pods")
+		return
+	}
+
+	for _, pod := range pods.Items {
+		klog.Info("Pod found ", "name ", pod.Name)
+	}
+
+	// Start the gops agent and listen on a specific address and port
```
rlenglet, line 179: Revert the change to the comment.
```suggestion
	// Start the gops agent
```

cmd/main.go
```diff
@@ -142,7 +165,18 @@ func main() {
...
-	// Start the gops agent
+	// List pods from default namespace
+	pods, err := clientset.CoreV1().Pods("illumio-cloud").List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Error(err, "Failed to list pods")
+		return
+	}
+
+	for _, pod := range pods.Items {
+		klog.Info("Pod found ", "name ", pod.Name)
```
rlenglet, line 176: 1. Info level is too verbose. How is this information useful? What is the point of all this?
1. Why use `klog` instead of `logger`? Was that for testing? If so, remove it before merging.

cmd/main.go
```diff
@@ -142,7 +165,18 @@ func main() {
 		zap.Duration("stream_success_period_auth", envConfig.StreamSuccessPeriod.Auth),
 	)
 
-	// Start the gops agent
+	// List pods from default namespace
+	pods, err := clientset.CoreV1().Pods("illumio-cloud").List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		klog.Error(err, "Failed to list pods")
```
rlenglet, line 171: Why use `klog` instead of `logger`?

cmd/main.go
```diff
@@ -142,7 +165,18 @@ func main() {
 		zap.Duration("stream_success_period_auth", envConfig.StreamSuccessPeriod.Auth),
 	)
 
-	// Start the gops agent
+	// List pods from default namespace
+	pods, err := clientset.CoreV1().Pods("illumio-cloud").List(context.TODO(), metav1.ListOptions{})
```
rlenglet, line 169: The namespace is not hardcoded anymore.
Use `envConfig.PodNamespace` like the rest of the code.

cmd/main.go
```diff
@@ -73,6 +78,24 @@ func main() {
...
+	// Set logrLogger as the global logger for klog
+	klog.SetLoggerWithOptions(logrLogger)
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		klog.Error(err, "Failed to load in-cluster config")
+		return
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
```
rlenglet, line 93: Why duplicate the logic from `streams.go`?
Why would you need another client set?
https://github.com/illumio/cloud-operator/blob/9d48abbc5f565a9be2a1d8d9787ad291b9d3b0dd/internal/controller/streams.go#L172

rlenglet (changes requested):

cmd/main.go
```diff
@@ -73,6 +78,24 @@ func main() {
...
+	if err != nil {
+		klog.Error(err, "Failed to create clientset")
+		return
+	}
+
```
rlenglet, line 98: I suspect that you only added this code to test.
That's OK.
Please remove this code now.
```suggestion
```

cmd/main.go
rlenglet, line 179: Same here, please remove this test code now.
```suggestion
	// Start the gops agent
```

rlenglet (changes requested):

PR #193, opened 2025-04-24
Build buildResourceApiGroupMap dynamically instead of allocating statically

ganesh-talla:
**Dynamic Resource-to-API Group Mapping:**
- Replaced the static, manually defined resourceAPIGroupMap with a dynamic mapping function (buildResourceApiGroupMap).
- The mapping is now generated by querying the Kubernetes API server using the discovery client.
- handled automatic Gateway API Exclusion(gateway.networking.k8s.io) this exclusion was done manually and statically in the code

**Previous Behavior**
- A static map (resourceAPIGroupMap) was used to map resources to API groups.
- Gateway API resources (gateways, gatewayclasses, httproutes) were manually excluded if the Gateway API group (gateway.networking.k8s.io) was not present.

### DIFF
--- a/.github/workflows/integration_fakeserver_tests_latest_version.yml
+++ b/.github/workflows/integration_fakeserver_tests_latest_version.yml
@@ -77,17 +77,17 @@ jobs:
         run: echo "Skipping Kubernetes version not supported by Minikube."
 
       - name: Start Minikube with pre-release Kubernetes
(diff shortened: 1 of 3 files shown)


rlenglet (dismissed):

evanj80-illumio (dismissed):
🥳

evanj80-illumio (approved):

PR #195, opened 2025-04-29
http proxy support

aaronIllumio:
cloud-operator should support redirecting all HTTP and gRPC requests via an HTTP proxy, by using the HTTPS_PROXY environment variable.

The gRPC library is supposed to use the HTTPS_PROXY env variable per [the Golang gRPC library's documentation](https://github.com/grpc/grpc-go/blob/master/Documentation/proxy.md).

The Golang oauth2 library doesn't seem to support it out of the box, and requires configuring a custom http.Transport with http.ProxyFromEnvironment, cf. https://github.com/concourse/concourse/issues/1337.


internal/controller/authenticator.go
```diff
@@ -181,11 +184,21 @@ func SetUpOAuthConnection(
 	}
 	tokenSource = GetTokenSource(ctx, oauthConfig, tlsConfig)
 	creds := credentials.NewTLS(tlsConfig)
+
+	proxyDialer := func(ctx context.Context, addr string) (net.Conn, error) {
+		proxyURL, err := http.ProxyFromEnvironment(&http.Request{URL: &url.URL{Host: addr}})
```
AriSweedler, line 189: Please have logs for the 3 cases:
- proxy set success
- proxy set failure
- proxy unset

aaronIllumio: done

rlenglet (changes requested):

internal/controller/authenticator.go
```diff
@@ -181,11 +192,23 @@ func SetUpOAuthConnection(
...
+		if err != nil || proxyURL == nil {
+			logger.Warn("[Proxy Failure] No valid proxy found or error occurred", zap.Error(err))
+			return net.Dial("tcp", addr)
+		}
```
rlenglet, line 201: 1. In the common case, where no HTTPS_PROXY environment variable is set, this will always print this scary Warn level message for no reason. You should only print a Warn level message if the variable is set and the proxy can't be found.
1. Remove the fancy formatting and keep it simple. When you want to add structure to log entries, use fields.
```suggestion
		proxyURL, err := http.ProxyFromEnvironment(&http.Request{URL: &url.URL{Host: addr}})
		if err != nil {
			logger.Warn("Invalid HTTPS proxy configured; ignoring proxy settings", zap.Error(err))
			return net.Dial("tcp", addr)
		}
		if proxyURL == nil { // No proxy configured
			return net.Dial("tcp", addr)
		}
```

internal/controller/authenticator.go
```diff
@@ -181,11 +192,23 @@ func SetUpOAuthConnection(
...
+		logger.Info("[Proxy Success] Using proxy", zap.String("proxy", proxyURL.String()))
```
rlenglet, line 202: 1. This is redundant with the logging about the HTTPS_PROXY environment variable above.
1. Remove the fancy formatting and keep it simple. When you want to add structure to log entries, use fields.
```suggestion
```

internal/controller/authenticator.go
```diff
@@ -162,6 +165,14 @@ func SetUpOAuthConnection(
 	}
 	tokenSource := GetTokenSource(contextWithTimeout, oauthConfig, tlsConfig)
 
+	// Log proxy settings
+	proxyEnv := os.Getenv("HTTPS_PROXY")
+	if proxyEnv != "" {
+		logger.Info("[Proxy Set] HTTPS_PROXY is set", zap.String("proxy", proxyEnv))
+	} else {
+		logger.Info("[Proxy Unset] HTTPS_PROXY is not set")
+	}
```
rlenglet, line 174: 1. Remove the fancy formatting and keep it simple. When you want to add structure to log entries, use fields.
1. `http.ProxyFromEnvironment`, which you use below, is documented to use multiple environment variables. Log all of them.
```suggestion
	logger.Info(
		"Configuring proxy from environment variables",
		zap.String("http_proxy", os.Getenv("HTTP_PROXY")),
		zap.String("https_proxy", os.Getenv("HTTPS_PROXY")),
		zap.String("no_proxy", os.Getenv("NO_PROXY")),
	)
```

internal/controller/authenticator_test.go
```diff
@@ -312,3 +318,92 @@ func (suite *ControllerTestSuite) TestParseToken() {
...
+	proxyServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		if _, err := w.Write([]byte("Proxy used")); err != nil {
+			suite.T().Errorf("Failed to write response: %v", err)
+		}
+	}))
+	defer proxyServer.Close()
+
+	// Set the HTTPS_PROXY environment variable to the mock proxy
+	os.Setenv("HTTPS_PROXY", proxyServer.URL)
```
rlenglet, line 333: Never modify or rely on environment variables in unit tests, because this prevents parallelizing unit tests.

internal/controller/authenticator_test.go
```diff
@@ -312,3 +318,92 @@ func (suite *ControllerTestSuite) TestParseToken() {
...
+	proxyURL, _ := url.Parse(proxyServer.URL)
+	transport := &http.Transport{
+		TLSClientConfig: tlsConfig,
+		Proxy:           http.ProxyURL(proxyURL), // Explicitly set the proxy
+	}
+	client := &http.Client{Transport: transport}
+
+	// Log the proxy URL and environment variable for debugging
+	fmt.Println("Proxy URL:", proxyServer.URL)
+	fmt.Println("HTTPS_PROXY Env:", os.Getenv("HTTPS_PROXY"))
```
rlenglet, line 355:
```suggestion
	// Log the proxy URL and environment variable for debugging
	suite.T().Logf("Proxy URL: %s", proxyServer.URL)
	suite.T().Logf("HTTPS_PROXY Env: %s", os.Getenv("HTTPS_PROXY"))
```
Same for all the other usages of `fmt.Print*` in tests in this PR.

internal/controller/authenticator.go
```diff
@@ -181,11 +192,23 @@ func SetUpOAuthConnection(
 	}
 	tokenSource = GetTokenSource(ctx, oauthConfig, tlsConfig)
 	creds := credentials.NewTLS(tlsConfig)
+
+	proxyDialer := func(ctx context.Context, addr string) (net.Conn, error) {
+		proxyURL, err := http.ProxyFromEnvironment(&http.Request{URL: &url.URL{Host: addr}})
```
rlenglet, line 197: You must define the HTTPS_PROXY environment variable in the Deployment's pod spec's `env` here:
https://github.com/illumio/cloud-operator/blob/df44b593931b44dd34b00bfa7b25eb248c9f90a5/cloud-operator/templates/cloud-operator-deployment.yaml#L56
Set it from a new Helm value, and define that value with `""` as default in `cloud-operator/values.yaml`.

rlenglet (changes requested):

internal/controller/authenticator.go
rlenglet: And once that's done, you should extend the `k8s_cluster` Terraform module to add a new `https_proxy` input variable here:
https://github.com/illumio/terraform-illumio-cloudsecure/blob/fc0e11de97687e8d74c4f0eb576027f5e898e0a1/modules/k8s_cluster/variables.tf
And use that variable to set the Helm value in the `helm_release` resource here:
https://github.com/illumio/terraform-illumio-cloudsecure/blob/fc0e11de97687e8d74c4f0eb576027f5e898e0a1/modules/k8s_cluster/main.tf#L5
Follow what we did for the `enable_falco` variable.

rlenglet: Closing. The Terraform module work will be done in that repo.

rlenglet (approved):
LGTM, thank you!

cloud-operator/values.yaml
```diff
@@ -73,3 +73,6 @@ env:
   streamSuccessPeriods:
     connect: "1h"
     auth: "2h"
+
+# Proxy configuration
+httpsProxy: ""
```
rlenglet, line 78: Moving this under `env` just above would be more consistent with the other env variables?

cloud-operator/templates/cloud-operator-deployment.yaml
```diff
@@ -88,6 +88,8 @@ spec:
               value: "{{ .Values.env.streamSuccessPeriods.connect }}"
             - name: STREAM_SUCCESS_PERIOD_AUTH
               value: "{{ .Values.env.streamSuccessPeriods.auth }}"
+            - name: HTTPS_PROXY
+              value: "{{ .Values.httpsProxy }}"
```
rlenglet, line 92: Move this under `env` to be consistent?
```suggestion
              value: "{{ .Values.env.httpsProxy }}"
```

rlenglet, line 92: Add a test like https://github.com/illumio/cloud-operator/blob/df44b593931b44dd34b00bfa7b25eb248c9f90a5/cloud-operator/tests/cloud-operator-deployment_test.yaml#L81-L86
To check that this is properly set.

rlenglet (changes requested):
A few more nits.

rlenglet (dismissed):
Thanks!!

rlenglet (dismissed):

internal/controller/authenticator.go
rlenglet, line 1: I think you need to fix `NewClientSet`. It currently creates every k8s Client by calling `kubernetes.NewForConfig`:
https://github.com/illumio/cloud-operator/blob/df44b593931b44dd34b00bfa7b25eb248c9f90a5/internal/controller/authenticator.go#L136

`kubernetes.NewForConfig` calls `rest.HTTPClientFor` to create the HTTP client:
https://github.com/kubernetes/client-go/blob/3aa3c77983ce87f8702329ea44d366cb16a04fd1/kubernetes/clientset.go#L500

`rest.HTTPClientFor` defaults to `http.DefaultClient` as the HTTP client since your config likely doesn't have any transport configured:
https://github.com/kubernetes/client-go/blob/17ad09eb27d15d66be61e55729f6da1ef5986faf/rest/transport.go#L44

And `http.DefaultClient` defaults to `http.DefaultTransport`, which is configured with `ProxyFromEnvironment`:
https://pkg.go.dev/net/http#DefaultTransport

So as-is in this PR, I think that every HTTP client used by k8s Clients are going through the proxy configured by `HTTPS_PROXY`, which is very likely to fail.

You need to fix `NewClientSet` to pass in a new HTTP Transport which isn't configured with `ProxyFromEnvironment`.

And please test this PR with an actual HTTPS proxy.

rlenglet (changes requested):

cloud-operator/templates/cloud-operator-deployment.yaml
```diff
@@ -88,6 +88,8 @@ spec:
               value: "{{ .Values.env.streamSuccessPeriods.connect }}"
             - name: STREAM_SUCCESS_PERIOD_AUTH
               value: "{{ .Values.env.streamSuccessPeriods.auth }}"
+            - name: HTTPS_PROXY
+              value: "{{ .Values.env.httpsProxy }}"
```
rlenglet, line 92: Why did you remove it from `values.yaml`???
Add it back. Every single Helm value should be listed in `values.yaml`, to serve as documentation and to make it clear what the default value is.

rlenglet (changes requested):

evanj80-illumio (approved):

.github/workflows/integration_fakeserver_tests.yml
```diff
@@ -99,5 +99,6 @@ jobs:
           kubectl wait --for=condition=ready pod -l app=cloud-operator -n illumio-cloud --timeout=300s && go test -v ./fakeserver/...
 
       - name: Fetch Kubernetes Pod logs
+        if: success() || failure()
```
rlenglet, line 102: What does this do? Why is this needed?

fakeserver/fakeserver_integration_test.go
```diff
@@ -37,6 +37,7 @@ func createSignedToken() string {
 
 // TestFakeServerConnectionSuccesfulAndRetry tests a client connecting to the gRPC server
 func TestFakeServerConnectionSuccesfulAndRetry(t *testing.T) {
+	t.Skip("Skipping integration test")
```
rlenglet, line 40: Is that ok? Why are you disabling this test? Was it already failing in main?

fakeserver/fakeserver_integration_test.go
```diff
@@ -108,6 +109,7 @@ mainloop:
 }
 
 func TestFailureDuringIntialCommit(t *testing.T) {
+	t.Skip("Skipping integration test")
```
rlenglet, line 112: Is that ok? Why are you disabling this test? Was it already failing in main?

fakeserver/oauth2_token_endpoint.go
```diff
@@ -110,24 +109,27 @@ func (a *AuthService) authenticateHandler(w http.ResponseWriter, r *http.Request
...
-
 		return
 	}
 
 	req.GrantType = r.FormValue("grant_type")
 	req.ClientID = r.FormValue("client_id")
 	req.ClientSecret = r.FormValue("client_secret")
 
+	a.logger.Info("Received credentials", zap.String("client_id", req.ClientID), zap.String("client_secret", req.ClientSecret))
+
```
rlenglet, line 120: Remove. NEVER log credentials.
```suggestion
```

rlenglet (changes requested):

fakeserver/oauth2_token_endpoint.go
rlenglet: Ah this is fakeserver, my bad.

fakeserver/server.go
```diff
@@ -287,3 +345,33 @@ func generateSelfSignedCert() (tls.Certificate, error) {
...
+	logger.Info("GetConfigurationUpdates stream started")
+	for {
+		// Simulate receiving updates (placeholder logic)
+		select {
+		case <-stream.Context().Done():
+			logger.Error("GetConfigurationUpdates stream context canceled", zap.Error(stream.Context().Err()))
+			return stream.Context().Err()
+		case <-time.After(10 * time.Second):
+			logger.Info("Sending keepalive on GetConfigurationUpdates stream")
+			// Placeholder: Send a keepalive or response
```
rlenglet, line 360: We don't send responses to keepalives.
```suggestion
```

fakeserver/server.go
```diff
@@ -287,3 +345,33 @@ func generateSelfSignedCert() (tls.Certificate, error) {
...
+func (s *server) SendKubernetesNetworkFlows(stream pb.KubernetesInfoService_SendKubernetesNetworkFlowsServer) error {
+	logger.Info("SendKubernetesNetworkFlows stream started")
+	for {
+		select {
+		case <-stream.Context().Done():
+			logger.Error("SendKubernetesNetworkFlows stream context canceled", zap.Error(stream.Context().Err()))
+			return stream.Context().Err()
+		case <-time.After(2 * time.Minute):
+			logger.Info("Sending keepalive on SendKubernetesNetworkFlows stream")
+			// Placeholder: Send a keepalive or response
```
rlenglet, line 374: We don't send responses to keepalives.
```suggestion
```

fakeserver/server.go
```diff
@@ -287,3 +345,33 @@ func generateSelfSignedCert() (tls.Certificate, error) {
 
 	return cert, nil
 }
+
+// Stub implementation for GetConfigurationUpdates
+func (s *server) GetConfigurationUpdates(stream pb.KubernetesInfoService_GetConfigurationUpdatesServer) error {
+	logger.Info("GetConfigurationUpdates stream started")
+	for {
```
rlenglet, line 352: You have to read the messages received from the stream.
Right now, it's only keepalives on this stream.
https://github.com/illumio/cloud-operator/blob/df44b593931b44dd34b00bfa7b25eb248c9f90a5/api/illumio/cloud/k8sclustersync/v1/k8s_info.proto#L408-L412

fakeserver/server.go
```diff
@@ -287,3 +345,33 @@ func generateSelfSignedCert() (tls.Certificate, error) {
...
+		case <-time.After(10 * time.Second):
+			logger.Info("Sending keepalive on GetConfigurationUpdates stream")
+			// Placeholder: Send a keepalive or response
+		}
+	}
+}
+
+func (s *server) SendKubernetesNetworkFlows(stream pb.KubernetesInfoService_SendKubernetesNetworkFlowsServer) error {
+	logger.Info("SendKubernetesNetworkFlows stream started")
+	for {
```
rlenglet, line 367: You have to read the messages received from the stream.
Right now, it's both network flows and keepalives.
https://github.com/illumio/cloud-operator/blob/df44b593931b44dd34b00bfa7b25eb248c9f90a5/api/illumio/cloud/k8sclustersync/v1/k8s_info.proto#L362-L369

internal/controller/authenticator.go
```diff
@@ -115,7 +118,7 @@ func (authn *Authenticator) WriteK8sSecret(ctx context.Context, keyData OnboardR
 	return nil
 }
 
-// NewClientSet returns a new Kubernetes clientset based on the execution environment.
+// Update NewClientSet to use the helper function
```
rlenglet, line 121: What does this comment mean?
Why replace a comment that follows the Golang style guide with one that doesn't?

internal/controller/authenticator.go
```diff
@@ -142,7 +147,7 @@ func IsRunningInCluster() bool {
 	return os.Getenv("KUBERNETES_SERVICE_HOST") != ""
 }
 
-// SetUpOAuthConnection establishes a gRPC connection using OAuth credentials and logging the process.
+// Update SetUpOAuthConnection to use the helper function
```
rlenglet, line 150: What does this comment mean?
Why replace a comment that follows the Golang style guide with one that doesn't?

internal/controller/authenticator.go
```diff
@@ -181,11 +194,34 @@ func SetUpOAuthConnection(
...
+		if err != nil {
+			logger.Warn("Invalid HTTPS proxy configured; ignoring proxy settings", zap.Error(err))
+			return net.Dial("tcp", addr)
+		}
+		if proxyURL == nil { // No proxy configured
+			return net.Dial("tcp", addr)
+		}
+		conn, err := proxy.Dial(ctx, "tcp", addr)
+		if err != nil {
+			logger.Error("Failed to connect via proxy; falling back to direct connection", zap.Error(err))
```
rlenglet, line 213: Info -> Warn to be consistent with the log above?
```suggestion
			logger.Warn("Failed to connect via proxy; falling back to direct connection", zap.Error(err))
```

internal/controller/proxy.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"net/http"
+	"net/url"
+	"os"
+
+	"go.uber.org/zap"
+)
+
+func configureProxy(logger *zap.Logger) func(*http.Request) (*url.URL, error) {
```
rlenglet, line 11: Why did you need to reimplement the standard `http.ProxyFromEnvironment`?
Explain that in a comment for this function.

rlenglet (changes requested):

.github/workflows/integration_fakeserver_tests.yml
aaronIllumio: This outputs the operator logs even if the test fails or passes

rlenglet: OK, thx!

rlenglet:
> Sorry this pr is still a work in progress everything you have commented on has been fixed or changed on my local branch. I only pushed the commit to save my progress. Sorry I should put this back in draft mode.

No problem at all!
I would recommend that you make this a "Draft PR" for the time being?

fakeserver/fakeserver_integration_test.go
```diff
@@ -84,6 +87,9 @@ mainloop:
 	// Start the server
 	err = fakeServer.start()
 	assert.NoError(t, err, "Failed to start the FakeServer")
+
+	logger.Info("Server started")
```
AriSweedler, line 91: I feel like this log belongs in the callee - `fakeServer.start()` - instead of in the caller

internal/controller/suite_setup_test.go
```diff
@@ -44,7 +46,24 @@ func (suite *ControllerTestSuite) SetupSuite() {
 func (suite *ControllerTestSuite) TearDownSuite() {
 	err := testhelper.TearDownTestCluster()
 	if err != nil {
-		suite.T().Fatal("Failed to delete test cluster " + err.Error())
+		suite.T().Log("Failed to delete test cluster on first attempt: " + err.Error())
+
+		// Retry deletion
+		err = testhelper.TearDownTestCluster()
+		if err != nil {
```
AriSweedler, line 53: This pattern of `err != nil` --> `Fatal` can be done by https://pkg.go.dev/github.com/stretchr/testify/require#NoError - don't worry about that change for this PR but in the future I think that library is great.

If a `require` fails then the test immediately ends. If an `assert` fails (https://pkg.go.dev/github.com/stretchr/testify/assert#NoError) then it just gives that great testify log message but continues executing the rest of the test (I think the naming is horrible, but thus is the way).

fakeserver/server.go
```diff
@@ -153,23 +156,72 @@ func (s *server) SendLogs(stream pb.KubernetesInfoService_SendLogsServer) error
...
+		req, err := stream.Recv()
+		if err == io.EOF {
+			return nil
```
AriSweedler, line 210: nit: log on healthy stream closure?

fakeserver/server.go
```diff
@@ -153,23 +156,72 @@ func (s *server) SendLogs(stream pb.KubernetesInfoService_SendLogsServer) error
...
 	return nil
 }
 
+// Stub implementation for GetConfigurationUpdates
+func (s *server) GetConfigurationUpdates(stream pb.KubernetesInfoService_GetConfigurationUpdatesServer) error {
+	logger.Info("GetConfigurationUpdates stream started")
+	for {
+		req, err := stream.Recv()
+		if err == io.EOF {
+			return nil
```
AriSweedler, line 192: nit: log on healthy stream closure?

fakeserver/server.go
```diff
@@ -153,23 +156,72 @@ func (s *server) SendLogs(stream pb.KubernetesInfoService_SendLogsServer) error
...
+		}
+		if err != nil {
+			return err
```
AriSweedler, line 195: if the caller doesn't log then plz log here (and in the other new fxn too)

AriSweedler (dismissed):

internal/controller/suite_setup_test.go
rlenglet: +1

fakeserver/fakeserver_integration_test.go
rlenglet: +1

fakeserver/server.go
```diff
@@ -153,23 +156,72 @@ func (s *server) SendLogs(stream pb.KubernetesInfoService_SendLogsServer) error
 }
 
 func logReceivedLogEntry(log *pb.LogEntry, logger *zap.Logger) error {
+	// Decode the JSON-encoded string into a LogEntry struct
 	var logEntry LogEntry = make(map[string]any)
 	if err := json.Unmarshal([]byte(log.JsonMessage), &logEntry); err != nil {
 		logger.Error("Error decoding JSON log message", zap.Error(err))
+		return status.Errorf(codes.Internal, "Internal error")
```
rlenglet, line 163:
```suggestion
		return status.Error(codes.InvalidArgument, "Invalid JSON message")
```

fakeserver/server.go
```diff
@@ -153,23 +156,72 @@ func (s *server) SendLogs(stream pb.KubernetesInfoService_SendLogsServer) error
...
+	}
+
+	// Check for an empty log entry
+	if len(logEntry) == 0 {
+		logger.Warn("Received empty log entry")
+		return status.Errorf(codes.InvalidArgument, "Empty log entry")
```
rlenglet, line 169:
```suggestion
		return status.Error(codes.InvalidArgument, "Empty log entry")
```

fakeserver/server.go
```diff
@@ -153,23 +156,72 @@ func (s *server) SendLogs(stream pb.KubernetesInfoService_SendLogsServer) error
...
 	}
 
 	level, err := logEntry.Level()
 	if err != nil {
 		logger.Error("Error converting log level", zap.Error(err))
-		return err
+		return status.Errorf(codes.Internal, "Internal error")
```
rlenglet, line 175:
```suggestion
		return status.Error(codes.InvalidArgument, "Invalid log level")
```

internal/controller/authenticator.go
```diff
@@ -133,6 +136,8 @@ func NewClientSet() (*kubernetes.Clientset, error) {
 		return nil, err
 	}
 
+	clusterConfig.Proxy = http.ProxyFromEnvironment
```
rlenglet, line 139: Why did you revert this?
Cf. my previous comment: https://github.com/illumio/cloud-operator/pull/195#discussion_r2067943714
```suggestion
	// Do not use the configured HTTP proxy for connections to apiserver within this cluster,
	// since it is only meant to be used for connections to CloudSecure
	clusterConfig.Proxy = http.ProxyURL(nil)
```
The `Proxy` field already defaults to `http.ProxyFromEnvironment`.
The problem with this default setting is that the HTTPS proxy configured for cloud-operator is meant to be used only for the egress connections from cloud-operator to Cloud Secure. It doesn't make sense to use it for communication between this pod and the apiserver within the same cluster. This is surely going to fail.

internal/controller/authenticator.go
```diff
@@ -181,11 +194,27 @@ func SetUpOAuthConnection(
...
 	tokenSource = GetTokenSource(ctx, oauthConfig, tlsConfig)
 	creds := credentials.NewTLS(tlsConfig)
+
+	proxyDialer := func(ctx context.Context, addr string) (net.Conn, error) {
+		proxyFunc := http.ProxyFromEnvironment
+		proxyURL, _ := proxyFunc(&http.Request{URL: &url.URL{Host: addr}})
+		if proxyURL == nil { // No proxy configured
+			return net.Dial("tcp", addr)
+		}
+		conn, err := proxy.Dial(ctx, "tcp", addr)
```
rlenglet, line 204: Why are you ignoring `proxyURL` and connect to the CloudSecure address anyway?
Have you really tested this?
```suggestion
		conn, err := proxy.Dial(ctx, "tcp", proxyURL.Host)
```

rlenglet (changes requested):

internal/controller/authenticator.go
```diff
@@ -181,11 +194,27 @@ func SetUpOAuthConnection(
...
+		}
+		return conn, nil
+	}
+
 	conn, err := grpc.NewClient(
 		aud,
 		grpc.WithTransportCredentials(creds),
 		grpc.WithPerRPCCredentials(oauth.TokenSource{TokenSource: tokenSource}),
 		grpc.WithKeepaliveParams(kacp),
+		grpc.WithContextDialer(proxyDialer),
```
rlenglet, line 217: Why did you need to replace the default dialer?
The default dialer is documented to already support HTTPS_PROXY out of the box:
https://github.com/grpc/grpc-go/blob/master/Documentation/proxy.md
Have you tested that it works out of the box?
If it doesn't work, add a big fat comment explaining how it is failing.

internal/controller/authenticator.go
```diff
@@ -162,6 +164,14 @@ func SetUpOAuthConnection(
 	}
 	tokenSource := GetTokenSource(contextWithTimeout, oauthConfig, tlsConfig)
 
+	// Log proxy settings
+	logger.Info(
```
rlenglet, line 168: Remove this?? It's redundant with the logging of `HTTPS_PROXY` already done in `main.go`:
https://github.com/illumio/cloud-operator/pull/195/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R154

rlenglet (changes requested):
LGTM, thanks!
Some relatively minor comments re: logging.

rlenglet (approved):
Looks good, thanks!

PR #196, opened 2025-04-29
Fix log message buffering

rlenglet:
Preserve remaining buffered messages during flush if any message failed to be sent.
Properly count dropped messages under all failure conditions.
Add unit tests for message buffering logic.

### DIFF
--- a/internal/controller/grpc_logger.go
+++ b/internal/controller/grpc_logger.go
@@ -107,10 +107,13 @@ func (b *BufferedGrpcWriteSyncer) flush() {
 		sentMessageCount += 1
 	}
 
-	// Only unbuffer the messages that could be sent and preserve the rest for the next flush()
-	newBufferLen := len(b.buffer) - sentMessageCount
-	copy(b.buffer, b.buffer[sentMessageCount:])
-	b.buffer = b.buffer[:newBufferLen]
+	// Only un-buffer the messages that were sent
+	// and keep the remaining messages buffered for the next flush()
+	if sentMessageCount > 0 {
+		newBufferLen := len(b.buffer) - sentMessageCount
+		copy(b.buffer, b.buffer[sentMessageCount:])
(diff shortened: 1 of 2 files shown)


internal/controller/grpc_logger.go
```diff
@@ -87,24 +88,29 @@ func (b *BufferedGrpcWriteSyncer) flush() {
...
-		}
-		b.lostLogEntriesCount = 0
 	}
 
+	sentMessageCount := 0
 	for _, jsonMessage := range b.buffer {
 		if err := b.sendLogEntry(jsonMessage); err != nil {
-			b.lostLogEntriesCount += 1
 			b.lostLogEntriesErr = err
+			break
```
evanj80-illumio, line 105: Is there a reason you did the shifting of elements in memory rather than just deleting the message from the buffer right after you send it?

evanj80-illumio (approved):

aaronIllumio (approved):
LGTM

internal/controller/grpc_logger.go
rlenglet: I couldn't think of any reason why, in `flush`,  we would discard up to 1023 messages if sending one message fails. That didn't seem right to me.

evanj80-illumio: I was more thinking why not delete that specific message in the buffer after it is sent.

rlenglet: It does delete the messages (cf. the `copy()` call and truncation of the buffer here:
https://github.com/illumio/cloud-operator/pull/196/files#diff-a1aafdfaa36efedb76f138377c415982363df10376dd06630fd08cf7fd0f96ebR111-R113

rlenglet: I clarified the logic and the comment in the code. I hope it's more clear.

evanj80-illumio (approved):

PR #198, opened 2025-04-29
Add support to collect network flows from OVN-K CNI plugin

evanj80-illumio:
Implement an IPFIX collector to collect network flows from Open vSwitch / OVN-K CNI plugin.
Generalize the Falco flow collection to also cover IPFIX flows.


api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -176,7 +178,7 @@ message KubernetesResourceMutation {
 }
 
 // A flow received from Falco using custom rules
```
AriSweedler, line 180: comment rot - plz delete or update!

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -364,7 +366,7 @@ message SendKubernetesNetworkFlowsRequest {
     Keepalive keepalive = 3;
 
     CiliumFlow cilium_flow = 1;
-    FalcoFlow falco_flow = 2;
+    StandardFlow standard_flow = 2;
```
AriSweedler, line 369: I thought that this was a breaking change, but I was mistaken: https://stackoverflow.com/a/45431953/7531823

As long as the wire representation remains the same for each key (number) it's all good. That is the case for this.

internal/controller/cilium_flow_collector.go
```diff
@@ -45,6 +45,7 @@ func discoverCiliumHubbleRelayAddress(ctx context.Context, ciliumNamespace strin
 	}
 
 	address := fmt.Sprintf("%s:%d", service.Spec.ClusterIP, service.Spec.Ports[0].Port)
+	fmt.Println("Hubble Relay Address", address)
```
AriSweedler, line 48: `fmt.Println` instead of `log`?

Pass in a logger as an arg if it's easy. Or if you wanna,,, add a logger to the `ctx` in the parent, and then extract it here. That way it is minimally invasive 👀

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,146 @@
...
+// IsOVNDeployed Checks for the presence of the `ovn-kubernetes` namespace, detection based on OVN namespace.
+// https://ovn-kubernetes.io/installation/launching-ovn-kubernetes-on-kind/#run-the-kind-deployment-with-podman
+func (sm *streamManager) isOVNDeployed(logger *zap.Logger) bool {
+	logger.Debug("Checking for OVN deployment")
+	clientset, err := NewClientSet()
+	if err != nil {
+		logger.Error("Failed to create clientset", zap.Error(err))
+		return false
+	}
+	_, err = clientset.CoreV1().Namespaces().Get(context.TODO(), "openshift-ovn-kubernetes", metav1.GetOptions{})
```
AriSweedler, line 23: Can u make `"openshift-ovn-kubernetes"` a `const`?

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,146 @@
...
+		return ctx.Err()
+	default:
+		logger.Info("Creating custom listener for OVN flows")
+		listener, err := net.ListenPacket("udp", ":4739")
+		if err != nil {
+			logger.Fatal("Failed to listen on OVN port", zap.Error(err))
+		}
+
+		logger.Info("OVN server listening", zap.String("address", sm.streamClient.IpfixCollectorPort))
+		buf := make([]byte, 65535)
```
AriSweedler, line 54: just curious - why this #? 2^16? Seems large. Maybe also make it a const?

internal/controller/streams.go
```diff
@@ -593,6 +605,30 @@ func (sm *streamManager) connectAndStreamConfigurationUpdates(logger *zap.Logger
...
 }
 
+// connectAndStreamOVNNetworkFlows creates OVN networkFlowsStream client and
+// begins the streaming of OVN network flows. Also starts a goroutine to send
+// keepalives at the configured period
+func (sm *streamManager) connectAndStreamOVNNetworkFlows(logger *zap.Logger, keepalivePeriod time.Duration) error {
+	ovnContext, ovnCancel := context.WithCancel(context.Background())
+	defer ovnCancel()
+
+	// go func() {
```
AriSweedler, line 615: keepalives commented out for now?

internal/controller/streams.go
```diff
@@ -777,20 +805,38 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
...
-			// Only start network flows stream if not disabled
-			if !sm.streamClient.disableNetworkFlowsCilium {
-				ciliumDone = make(chan struct{})
+			ovnDone := make(chan struct{})
+			falcoDone := make(chan struct{})
+			ciliumDone := make(chan struct{})
+			logger.Info("Evaluating case: Cilium", zap.String("ciliumNamespace", sm.streamClient.ciliumNamespace))
+			logger.Info("Cilium case is true: Hubble Relay found", zap.String("ciliumNamespace", sm.streamClient.ciliumNamespace))
+			logger.Info("IsOVNDeployed", zap.Bool("isOVNDeployed", sm.isOVNDeployed(logger)))
+			switch {
```
AriSweedler, line 814: Instead of figuring it out in the switch and then setting `sm.streamClient.flowCollector` in the body, can u write a function that returns data?

I have 3 reasons for saying this (really only 1 reason):
1. Aesthetics lol sorry (not a real reason so only 2 reasons)
2. Any new member adding a CNI had better remember to set `sm.streamClient.FlowCollector` in the body of the loop and there's no benefit to requiring that. This is kinda a weak reason as it's not that hard to figure out...
3. And this is the most compelling reason saved for last: if you ever wanted to write a helper script or a test function that used our CNI detection, then this type of refactor would be necessary

internal/controller/cilium_flow_collector.go
evanj80-illumio: Sorry not sure why this was in here

internal/controller/ovn_ipfix_collector.go
evanj80-illumio: Thats the maximum size of a udp packet from IPFIX

internal/controller/streams.go
evanj80-illumio: Gave it a try, let me know what you think

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+const ovnNamespace = "openshift-ovn-kubernetes"
+
+// https://datatracker.ietf.org/doc/rfc7011/
+// The IPFIX Message Header 16-bit Length field limits the length of an
+// IPFIX Message to 65535 octets, including the header.  A Collecting
+// Process MUST be able to handle IPFIX Message lengths of up to
+// 65535 octets.
+const ipfixMaxLength = 65535
```
rlenglet, line 50: OK. That might be true when using TCP or TLS as the transport layer.
But with UDP we're going to be capped by the MTU, e.g. max 9000 bytes with jumbo frames.
And OVS caps the message size to 1024 anyway:
https://github.com/openvswitch/ovs/blob/1b2eff486f30a4c78e6a936c6f81cd4961b58f8f/ofproto/ofproto-dpif-ipfix.c#L560

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+// IsOVNDeployed Checks for the presence of the `ovn-kubernetes` namespace, detection based on OVN namespace.
+// https://ovn-kubernetes.io/installation/launching-ovn-kubernetes-on-kind/#run-the-kind-deployment-with-podman
+func (sm *streamManager) isOVNDeployed(logger *zap.Logger) bool {
+	logger.Debug("Checking for OVN deployment")
+	clientset, err := NewClientSet()
+	if err != nil {
+		logger.Error("Failed to create clientset", zap.Error(err))
+		return false
+	}
+	_, err = clientset.CoreV1().Namespaces().Get(context.TODO(), ovnNamespace, metav1.GetOptions{})
```
rlenglet, line 61: Make the namespace configurable like for Cilium, etc.

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+// The IPFIX Message Header 16-bit Length field limits the length of an
+// IPFIX Message to 65535 octets, including the header.  A Collecting
+// Process MUST be able to handle IPFIX Message lengths of up to
+// 65535 octets.
+const ipfixMaxLength = 65535
+
+// IsOVNDeployed Checks for the presence of the `ovn-kubernetes` namespace, detection based on OVN namespace.
+// https://ovn-kubernetes.io/installation/launching-ovn-kubernetes-on-kind/#run-the-kind-deployment-with-podman
+func (sm *streamManager) isOVNDeployed(logger *zap.Logger) bool {
+	logger.Debug("Checking for OVN deployment")
```
rlenglet, line 55:
```suggestion
// isOVNKDeployed checks for the presence of the `openshift-ovn-kubernetes` namespace, detection based on OVN-K namespace.
// https://ovn-kubernetes.io/installation/launching-ovn-kubernetes-on-kind/#run-the-kind-deployment-with-podman
func (sm *streamManager) isOVNKDeployed(logger *zap.Logger) bool {
	logger.Debug("Checking for OVN-K deployment")
```

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+type OVNFlow struct {
+	SourceIP        string
+	DestinationIP   string
+	SourcePort      uint16
+	DestinationPort uint16
+	Protocol        string
+	IPVersion       string
+}
+
+const ovnNamespace = "openshift-ovn-kubernetes"
```
rlenglet, line 43: Make this configurable like for Cilium, etc.
```suggestion
const ovnkNamespace = "openshift-ovn-kubernetes"
```

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+// FieldInfo represents the offset and size of a field in a template.
+type FieldInfo struct {
+	Offset int
+	Size   int
+}
+
+// OVNFlow represents a flow captured from OVN.
+type OVNFlow struct {
```
rlenglet, line 34:
```suggestion
// OVNKFlow represents a flow captured from OVN-Kubernetes.
type OVNKFlow struct {
```

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+		logger.Error("Failed to create clientset", zap.Error(err))
+		return false
+	}
+	_, err = clientset.CoreV1().Namespaces().Get(context.TODO(), ovnNamespace, metav1.GetOptions{})
+	if err != nil {
+		logger.Debug("OVN namespace not found", zap.Error(err))
+		return false
+	}
+	logger.Debug("OVN namespace found")
+	return true
```
rlenglet, line 67:
```suggestion
	if err != nil {
		logger.Debug("OVN-K namespace not found", zap.Error(err))
		return false
	}
	logger.Debug("OVN-K namespace found")
	return true
```

rlenglet: Same with the rest of this file. Please make sure to use the term "OVN-K" (for OVN-Kubernetes) instead of just OVN since OVN itself is not specific to Kubernetes, and OVN-K is a CNI plugin that is built on top of OVN.

rlenglet (changes requested):
First batch of comments.

Make sure to use the term "OVN-K" (for OVN-Kubernetes) consistently instead of just OVN since OVN itself is not specific to Kubernetes, and OVN-K is a CNI plugin that is built on top of OVN.

internal/controller/ovn_ipfix_collector.go
rlenglet, line 1: Rename the file: ovn -> ovnk
Same with the _test.go file.

rlenglet (changes requested):

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+		defer cancelProcessOVNFlow()
+		logger.Debug("Starting processOVNFlow goroutine")
+		if err := sm.processOVNFlow(ctxProcessOVNFlow, logger); err != nil {
+			logger.Error("Error in processOVNFlow", zap.Error(err))
+		}
+	}()
+	select {
+	case <-ctx.Done():
+		logger.Info("Context canceled in startOVNIPFIXCollector")
+		return ctx.Err()
```
rlenglet, line 86: What is the point of this case?
Nothing is blocking until this point.

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+func (sm *streamManager) startOVNIPFIXCollector(ctx context.Context, logger *zap.Logger) error {
+	logger.Info("Starting OVN IPFIX Collector")
+	go func() {
+		ctxProcessOVNFlow, cancelProcessOVNFlow := context.WithCancel(ctx)
+		defer cancelProcessOVNFlow()
+		logger.Debug("Starting processOVNFlow goroutine")
+		if err := sm.processOVNFlow(ctxProcessOVNFlow, logger); err != nil {
+			logger.Error("Error in processOVNFlow", zap.Error(err))
+		}
+	}()
```
rlenglet, line 82: What is the point of creating this cancelable context here?
Nothing can call `cancelProcessOVNFlow` while `processOVNFlow` is executing.
This has to be done outside of the goroutine, so that the goroutine terminates when `startOVNKIPFIXCollector` returns.
```suggestion
	ctxProcessOVNFlow, cancelProcessOVNFlow := context.WithCancel(ctx)
	defer cancelProcessOVNFlow()

	go func() {
		logger.Debug("Starting processOVNFlow goroutine")
		if err := sm.processOVNFlow(ctxProcessOVNFlow, logger); err != nil {
			logger.Error("Error in processOVNFlow", zap.Error(err))
		}
	}()
```

Of course, that could be unnecessary if `ctx` is guaranteed to be canceled by the caller shortly after this returns.
But I think this makes this function more self-contained.

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+// It decodes the IPFIX packets, extracts data records, and converts them into
+// StandardFlow objects, which are then cached for sending to CloudSecure.
+func (sm *streamManager) processOVNFlow(ctx context.Context, logger *zap.Logger) error {
+	// Initialize the BasicTemplateSystem using the CreateTemplateSystem function
+	templateSystem := netflows.CreateTemplateSystem()
+	// Intialize the template set via the binary observed
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Info("Context canceled in processOVNFlow. Exiting.")
```
rlenglet, line 123: Make this log message less cryptic. Don't refer to internal symbols. Don't terminate the message with a period.
```suggestion
			logger.Info("Context canceled while processing IPFIX message")
```

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+				sm.streamClient.ovnEventChan <- buf[:n]
+			}
+		}
+	}
+}
+
+// processOVNFlow processes OVN IPFIX flows received from the ovnEventChan.
+// It decodes the IPFIX packets, extracts data records, and converts them into
+// StandardFlow objects, which are then cached for sending to CloudSecure.
+func (sm *streamManager) processOVNFlow(ctx context.Context, logger *zap.Logger) error {
```
rlenglet, line 116: This function processes any received IPFIX message, which is not necessarily flow records.
Rename accordingly and rephrase the comment.
```suggestion
func (sm *streamManager) processIPFIXMessage(ctx context.Context, logger *zap.Logger) error {
```

internal/controller/ovn_ipfix_collector.go
rlenglet: Same for the log messages in the rest of this function.

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+	SequenceNumber uint32
+	ObservationID  uint32
+}
+
+// FieldInfo represents the offset and size of a field in a template.
+type FieldInfo struct {
+	Offset int
+	Size   int
+}
+
```
rlenglet, line 32: Remove. Those are now unused.
```suggestion
```

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+}
+
+// processDataRecord processes a single data record and converts it into an OVNFlow.
+// If any parsing step fails, it returns an error and skips the record.
+func processDataRecord(dataRecord netflows.DataRecord) (OVNFlow, error) {
+	ovnFlow := OVNFlow{}
+	for _, field := range dataRecord.Values {
+		switch field.Type {
+		case 8: // Assuming 8 corresponds to sourceIPv4Address
```
rlenglet, line 247: Yes, you're assuming. No need to comment about that. :)
Those field types are standard and officially defined by the IANA.
```suggestion
		case 8: // sourceIPv4Address
```
Same for the other comments below.

internal/controller/ovn_ipfix_collector.go
rlenglet, line 1: You have to extract the timestamps from the flow records.

You first get the _"export time"_ from the IPFIX header:
https://github.com/netsampler/goflow2/blob/fb53b559f1abe986081384d0470c02937ef087e5/decoders/netflow/ipfix.go#L451

Then each flow record exported by OVS has two deltas (start and end) in microseconds that are relative to that export time.
Those fields have types 608 and 609:
https://github.com/netsampler/goflow2/blob/fb53b559f1abe986081384d0470c02937ef087e5/decoders/netflow/ipfix.go#L608-L609
Buy subtracting the delta from the export time, you get each flow's start and end timestamps.

rlenglet (changes requested):

internal/controller/ovn_ipfix_collector.go
```diff
@@ -0,0 +1,288 @@
...
+		}
+	}
+}
+
+// processOVNFlow processes OVN IPFIX flows received from the ovnEventChan.
+// It decodes the IPFIX packets, extracts data records, and converts them into
+// StandardFlow objects, which are then cached for sending to CloudSecure.
+func (sm *streamManager) processOVNFlow(ctx context.Context, logger *zap.Logger) error {
+	// Initialize the BasicTemplateSystem using the CreateTemplateSystem function
+	templateSystem := netflows.CreateTemplateSystem()
```
rlenglet, line 118: This needs to be inititalized from a statically compiled template set, as received from OVS.
To store the binary messages in the code, I'd suggest using `go-bindata`.
https://github.com/go-bindata/go-bindata
Then decode them here just as you just received them from the exporter.

rlenglet (changes requested):

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -175,15 +177,15 @@ message KubernetesResourceMutation {
   }
 }
 
-// A flow received from Falco using custom rules
-message FalcoFlow {
+// A flow received from a CNI
+message StandardFlow {
```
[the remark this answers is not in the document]
evanj80-illumio: Sure that is fine with me. Did not have anything particular for the name in my head so just chose standard

cloud-operator/values.schema.json
```diff
@@ -177,6 +177,14 @@
...
         },
+        "service": {
+            "properties": {
+                "clusterIP": {
+                    "type": "string"
+                }
+            },
+            "type": "object"
+        },
         "serviceAccount": {
```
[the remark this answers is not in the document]
evanj80-illumio: Will do automagically now with GHA

cloud-operator/templates/cloud-operator-service.yaml
```diff
@@ -9,11 +9,18 @@ metadata:
     app: {{ include "cloud-operator.name" . }}
     release: {{ .Release.Name }}
 spec:
```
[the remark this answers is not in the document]
evanj80-illumio: Good catch. Im wondering how I was able to receive packets in the first place with this network policy already applied...... Should it not have blocked the ingress traffic from everywhere but falco?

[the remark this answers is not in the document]
rlenglet: It depends on whether the NetworkPolicy was enforced!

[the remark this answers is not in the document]
evanj80-illumio: Aw OVN-K by default may not do policy enforcement

internal/controller/ovn_ipfix_collector.go
evanj80-illumio: Dealio

internal/controller/ovn_ipfix_collector.go
evanj80-illumio: Totally missed this my linter did not alert me!

internal/controller/ovn_ipfix_collector.go
evanj80-illumio: Gotcha that makes sense!

internal/controller/ovn_ipfix_collector.go
evanj80-illumio: Tried my own way! Let me know what you think

internal/controller/ovn_ipfix_collector.go
evanj80-illumio: Should I change the value or just the comment?

internal/controller/streams.go
```diff
@@ -182,7 +187,7 @@ func (sm *streamManager) buildResourceApiGroupMap(resources []string, clientset
 	// Iterate over all API groups and versions
 	for _, group := range apiGroups.Groups {
 		// Skip metrics API group
-		if group.Name == "metrics.k8s.io" {
+		if group.Name == "metrics.k8s.io" || group.Name == "config.openshift.io" {
```
evanj80-illumio, line 190: Temp fix until @ganesh-talla  changes lands

cloud-operator/templates/cloud-operator-template-set.yaml
```diff
@@ -0,0 +1,6 @@
...
...ABrcg38AAQAAGtyDgAACAAAa3IOBAAIAABrcg3sAAQAAGtyDfP//AAAa3ACeAAQAnwAEAIUACACHAAgAAgAIAFYACAFiAAgBYwAIAWQACAFlAAgBZgAIABMACACuAAgBYAAIAWEACACIAAEBAwAlAIoABAA9AAEAOAAGAFAABgEAAAIA8AABAAoABAFwAAQAUv//AFP//wAOAAQBcQAEAFL//wBT//+DfQAEAAAa3IN+AAQAABrcg38AAQAAGtyDgAACAAAa3IOBAAIAABrcg3sAAQAAGtyDfP//AAAa3ACeAAQAnwAEAIUACACHAAgAAgAIAFYACAFiAAgBYwAIAWQACAFlAAgBZgAIABMACACuAAgBYAAIAWEACACIAAEBFAAsAIoABAA9AAEAOAAGAFAABgEAAAIA8AABAAoABAFwAAQAUv//AFP//wA8AAEAwAABAAQAAQDDAAEAxAABAAUAAQAIAAQADAAEAJ4ABACfAAQAhQAIAIcACAACAAgAVgAIAWIACAFjAAgBZAAIAWUACAFmAAgAEwAIAK4ACAFgAAgBYQAIAIgAAQCEAAgAhgAIAAEACABVAAgAxgAIAMcACAAZAAgAGgAIABQACACvAAgBFQAwAIoABAA9AAEAOAAGAFAABgEAAAIA8AABAAoABAFwAAQAUv//AFP//wAOAAQBcQAEAFL//wBT//8APAABAMAAAQAEAAEAwwABAMQAAQAFAAEACAAEAAwABACeAAQAnwAEAIUACACHAAgAAgAIAFYACAFiAAgBYwAIAWQACAFlAAgBZgAIABMACACuAAgBYAAIAWEACACIAAEAhAAIAIYACAABAAgAVQAIAMYACADHAAgAGQAIABoACAAUAAgArwAIARYAMwCKAAQAPQABADgABgBQAAYBAAACAPAAAQAKAAQBcAAEAFL//wBT//8APAABAMAAAQAEAAEAwwABAMQAAQAFAAEACAAEAAwABIN9AAQAABrcg34ABAAAGtyDfwABAAAa3IOAAAIAABrcg4EAAgAAGtyDewABAAAa3IN8//8AABrcAJ4ABACfAAQAhQAIAIcACAACAAgAVgAIAWIACAFjAAgBZAAIAWUACAFmAAgAEwAIAK4ACAFgAAgBYQAIAIgAAQCEAAgAhgAIAAEACABVAAgAxgAIAMcACAAZAAgAGgAIABQACACvAAg=
```
rlenglet, line 6: I'd rename to make obvious that this is specific to Open vSwitch.
That way, if you also rename the mount point, the full path becomes `/ipfix-template-sets/openvswitch.bin`.
```suggestion
  openvswitch.bin: AAoEuGgrbqoAAAACAAAAAAACBKgBAAAaAIoABAA9AAEAOAAGAFAABgEAAAIA8AABAAoABAFwAAQAUv//AFP//wCeAAQAnwAEAIUACACHAAgAAgAIAFYACAFiAAgBYwAIAWQACAFlAAgBZgAIABMACACuAAgBYAAIAWEACACIAAEBAQAeAIoABAA9AAEAOAAGAFAABgEAAAIA8AABAAoABAFwAAQAUv//AFP//wAOAAQBcQAEAFL//wBT//8AngAEAJ8ABACFAAgAhwAIAAIACABWAAgBYgAIAWMACAFkAAgBZQAIAWYACAATAAgArgAIAWAACAFhAAgAiAABAQIAIQCKAAQAPQABADgABgBQAAYBAAACAPAAAQAKAAQBcAAEAFL//wBT//+DfQAEAAAa3IN+AAQAABrcg38AAQAAGtyDgAACAAAa3IOBAAIAABrcg3sAAQAAGtyDfP//AAAa3ACeAAQAnwAEAIUACACHAAgAAgAIAFYACAFiAAgBYwAIAWQACAFlAAgBZgAIABMACACuAAgBYAAIAWEACACIAAEBAwAlAIoABAA9AAEAOAAGAFAABgEAAAIA8AABAAoABAFwAAQAUv//AFP//wAOAAQBcQAEAFL//wBT//+DfQAEAAAa3IN+AAQAABrcg38AAQAAGtyDgAACAAAa3IOBAAIAABrcg3sAAQAAGtyDfP//AAAa3ACeAAQAnwAEAIUACACHAAgAAgAIAFYACAFiAAgBYwAIAWQACAFlAAgBZgAIABMACACuAAgBYAAIAWEACACIAAEBFAAsAIoABAA9AAEAOAAGAFAABgEAAAIA8AABAAoABAFwAAQAUv//AFP//wA8AAEAwAABAAQAAQDDAAEAxAABAAUAAQAIAAQADAAEAJ4ABACfAAQAhQAIAIcACAACAAgAVgAIAWIACAFjAAgBZAAIAWUACAFmAAgAEwAIAK4ACAFgAAgBYQAIAIgAAQCEAAgAhgAIAAEACABVAAgAxgAIAMcACAAZAAgAGgAIABQACACvAAgBFQAwAIoABAA9AAEAOAAGAFAABgEAAAIA8AABAAoABAFwAAQAUv//AFP//wAOAAQBcQAEAFL//wBT//8APAABAMAAAQAEAAEAwwABAMQAAQAFAAEACAAEAAwABACeAAQAnwAEAIUACACHAAgAAgAIAFYACAFiAAgBYwAIAWQACAFlAAgBZgAIABMACACuAAgBYAAIAWEACACIAAEAhAAIAIYACAABAAgAVQAIAMYACADHAAgAGQAIABoACAAUAAgArwAIARYAMwCKAAQAPQABADgABgBQAAYBAAACAPAAAQAKAAQBcAAEAFL//wBT//8APAABAMAAAQAEAAEAwwABAMQAAQAFAAEACAAEAAwABIN9AAQAABrcg34ABAAAGtyDfwABAAAa3IOAAAIAABrcg4EAAgAAGtyDewABAAAa3IN8//8AABrcAJ4ABACfAAQAhQAIAIcACAACAAgAVgAIAWIACAFjAAgBZAAIAWUACAFmAAgAEwAIAK4ACAFgAAgBYQAIAIgAAQCEAAgAhgAIAAEACABVAAgAxgAIAMcACAAZAAgAGgAIABQACACvAAg=
```

cmd/main.go
```diff
@@ -99,11 +100,13 @@ func main() {
 	bindEnv(logger, "stream_success_period_connect", "STREAM_SUCCESS_PERIOD_CONNECT")
 	bindEnv(logger, "stream_success_period_auth", "STREAM_SUCCESS_PERIOD_AUTH")
 	bindEnv(logger, "https_proxy", "HTTPS_PROXY")
+	bindEnv(logger, "ovnk_namespace", "OVNK_NAMESPACE")
```
rlenglet, line 103: I would prefer to keep those in alphabetical order.

cmd/main.go
```diff
@@ -114,13 +117,15 @@ func main() {
 	viper.SetDefault("stream_success_period_connect", defaultStreamSuccessPeriodConnect)
 	viper.SetDefault("stream_success_period_auth", defaultStreamSuccessPeriodAuth)
 	viper.SetDefault("https_proxy", "")
+	viper.SetDefault("ovnk_namespace", "openshift-ovn-kubernetes")
```
rlenglet, line 120: I would prefer to keep those in alphabetical order.

internal/controller/ovn_template_sets/template_set_0_1747676841915736072.bin
rlenglet, line 1: Add a README to explain:
* How you got this file.
* Why it's named like that.
* What it's used for (copied and pasted into a ConfigMap).

cloud-operator/templates/cloud-operator-template-set.yaml
rlenglet, line 1: Rename into `ipfix-template-sets.yaml`?

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+
+// OVNKFlow represents a flow captured from OVN-Kubernetes.
+type OVNKFlow struct {
+	SourceIP        string
+	DestinationIP   string
+	SourcePort      uint16
+	DestinationPort uint16
+	Protocol        string
+	IPVersion       string
+	StartTimestamp  *timestamppb.Timestamp
```
rlenglet, line 29: Also capture the end timestamp. It's not a significant effort to do that.
We're not sending the end timestamp to CloudSecure yet, but it would make sense to do that since we're aggregating flows in cloud-operator now.

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// https://datatracker.ietf.org/doc/rfc7011/
+// The IPFIX Message Header 16-bit Length field limits the length of an
+// IPFIX Message to 65535 octets, including the header.  A Collecting
+// Process MUST be able to handle IPFIX Message lengths of up to
+// 65535 octets.
+const ipfixMaxLength = 65535
```
rlenglet, line 37:
```suggestion
const ipfixMessageMaxLength = 65535
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+
+// isOVNKDeployed checks for the presence of the `openshift-ovn-kubernetes` namespace, detection based on OVN-K namespace, this is configurable.
+// https://ovn-kubernetes.io/installation/launching-ovn-kubernetes-on-kind/#run-the-kind-deployment-with-podman
+func (sm *streamManager) isOVNKDeployed(logger *zap.Logger, ovnkNamespace string) bool {
+	logger.Debug("Checking for OVN-K deployment")
+	clientset, err := NewClientSet()
+	if err != nil {
+		logger.Error("Failed to create clientset", zap.Error(err))
+		return false
```
rlenglet, line 46: I don't like ignoring errors like that.
You should return the error and do something about it.

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	_, err = clientset.CoreV1().Namespaces().Get(context.TODO(), ovnkNamespace, metav1.GetOptions{})
```
rlenglet, line 48: Pass a proper Context.
```suggestion
	_, err = clientset.CoreV1().Namespaces().Get(ctx, ovnkNamespace, metav1.GetOptions{})
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+// https://datatracker.ietf.org/doc/rfc7011/
+// The IPFIX Message Header 16-bit Length field limits the length of an
+// IPFIX Message to 65535 octets, including the header.  A Collecting
+// Process MUST be able to handle IPFIX Message lengths of up to
+// 65535 octets.
+const ipfixMaxLength = 65535
+
+// isOVNKDeployed checks for the presence of the `openshift-ovn-kubernetes` namespace, detection based on OVN-K namespace, this is configurable.
+// https://ovn-kubernetes.io/installation/launching-ovn-kubernetes-on-kind/#run-the-kind-deployment-with-podman
+func (sm *streamManager) isOVNKDeployed(logger *zap.Logger, ovnkNamespace string) bool {
```
rlenglet, line 41: This needs a proper context.
```suggestion
func (sm *streamManager) isOVNKDeployed(ctx context.Context, logger *zap.Logger, ovnkNamespace string) bool {
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+// It creates a custom listener for OVN-K flows on port 4739 and processes
+// the received flows in a separate goroutine.
+func (sm *streamManager) startOVNKIPFIXCollector(ctx context.Context, logger *zap.Logger) error {
+	logger.Info("Starting OVN-K IPFIX Collector")
+	ctxProcessOVNKFlow, cancelProcessOVNKFlow := context.WithCancel(ctx)
+	defer cancelProcessOVNKFlow()
+	go func() {
+		logger.Debug("Starting processIPFIXMessage goroutine")
+		if err := sm.processIPFIXMessage(ctxProcessOVNKFlow, logger); err != nil {
+			logger.Error("Error in processIPFIXMessage", zap.Error(err))
```
rlenglet, line 67: Log messages at Info level and above are meant for users.
Make them readable. Esp. don't use Golang symbols in there.
```suggestion
			logger.Error("OVN-K IPFIX collector failed; terminating collector", zap.Error(err))
```
Also, if the context is canceled, that will be logged in many places, in `processIPFIXMessage` and in the main reading loop below. That will be quite noisy. You should ignore those specific errors.

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	logger.Debug("OVN-K namespace found")
+	return true
+}
+
+// startOVNKIPFIXCollector starts the OVN-K IPFIX collector.
+// It creates a custom listener for OVN-K flows on port 4739 and processes
+// the received flows in a separate goroutine.
+func (sm *streamManager) startOVNKIPFIXCollector(ctx context.Context, logger *zap.Logger) error {
+	logger.Info("Starting OVN-K IPFIX Collector")
```
rlenglet, line 61:
```suggestion
	logger.Info("Starting OVN-K IPFIX collector")
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+	defer cancelProcessOVNKFlow()
+	go func() {
+		logger.Debug("Starting processIPFIXMessage goroutine")
+		if err := sm.processIPFIXMessage(ctxProcessOVNKFlow, logger); err != nil {
+			logger.Error("Error in processIPFIXMessage", zap.Error(err))
+		}
+	}()
+	select {
+	case <-ctx.Done():
+		logger.Info("Context canceled in startOVNKIPFIXCollector")
```
rlenglet, line 72:
```suggestion
		logger.Info("Context canceled in OVN-K IPFIX collector")
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+		return ctx.Err()
+	default:
+		logger.Info("Creating custom listener for OVN-K flows")
```
rlenglet, line 75:
```suggestion
		logger.Debug("Listening on IPFIX port")
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+		listener, err := net.ListenPacket("udp", ":4739")
+		if err != nil {
+			logger.Fatal("Failed to listen on OVN-K port", zap.Error(err))
```
rlenglet, line 78:
```suggestion
			logger.Fatal("Failed to listen on IPFIX port", zap.Error(err))
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+			logger.Error("Error in processIPFIXMessage", zap.Error(err))
+		}
+	}()
+	select {
+	case <-ctx.Done():
+		logger.Info("Context canceled in startOVNKIPFIXCollector")
+		return ctx.Err()
+	default:
+		logger.Info("Creating custom listener for OVN-K flows")
+		listener, err := net.ListenPacket("udp", ":4739")
```
rlenglet, line 76: Replace the hardcoded address with the configuration parameter.
```suggestion
		listener, err := net.ListenPacket("udp", sm.streamClient.IPFIXCollectorPort))
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+		if err != nil {
+			logger.Fatal("Failed to listen on OVN-K port", zap.Error(err))
+		}
+
+		logger.Info("OVN-k server listening", zap.String("address", sm.streamClient.IPFIXCollectorPort))
```
rlenglet, line 81:
```suggestion
		logger.Info("OVN-K IPFIX collector listening")
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+		return false
+	}
+	logger.Debug("OVN-K namespace found")
+	return true
+}
+
+// startOVNKIPFIXCollector starts the OVN-K IPFIX collector.
+// It creates a custom listener for OVN-K flows on port 4739 and processes
+// the received flows in a separate goroutine.
+func (sm *streamManager) startOVNKIPFIXCollector(ctx context.Context, logger *zap.Logger) error {
```
rlenglet, line 60:
```suggestion
func (sm *streamManager) startOVNKIPFIXCollector(ctx context.Context, logger *zap.Logger) error {
	logger = logger.With(
		zap.String("address", sm.streamClient.IPFIXCollectorPort),
	)
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+				sm.streamClient.ovnkEventChan <- buf[:n]
+			}
+		}
+	}
+}
+
+// processIPFIXMessage processes OVN-K IPFIX flows received from the ovnkEventChan.
+// It decodes the IPFIX packets, extracts data records, and converts them into
+// FiveTupleFlow objects, which are then cached for sending to CloudSecure.
+func (sm *streamManager) processIPFIXMessage(ctx context.Context, logger *zap.Logger) error {
```
rlenglet, line 103:
```suggestion
// processIPFIXMessages processes OVN-K IPFIX flows received from the ovnkEventChan.
// It decodes the IPFIX packets, extracts data records, and converts them into
// FiveTupleFlow objects, which are then cached for sending to CloudSecure.
func (sm *streamManager) processIPFIXMessages(ctx context.Context, logger *zap.Logger) error {
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+	// This segment of code is used to decode the template set and add it to the template system
+	templateSystem := netflows.CreateTemplateSystem()
+	packet, err := os.ReadFile("/template-storage/template_set.bin")
```
rlenglet, line 106:
```suggestion
	packet, err := os.ReadFile("/ipfix-template-sets/openvswitch.bin")
```

internal/controller/ovnk_ipfix_collector.go
rlenglet, line 67: You're doing nothing if `processIPFIXMessage` returns an error???
This should at the minimum cancel the context so that the reading loop can exit.
Otherwise, as-is, if `processIPFIXMessage` returns an error, cloud-operator can't recover without rebooting the pod.
```suggestion
			logger.Error("OVN-K IPFIX collector failed; terminating collector", zap.Error(err))
			cancelProcessOVNKFlow()
```
Processing each message directly in the loop below instead of unnecessarily doing that in a goroutine will eliminate that problem.

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// startOVNKIPFIXCollector starts the OVN-K IPFIX collector.
+// It creates a custom listener for OVN-K flows on port 4739 and processes
+// the received flows in a separate goroutine.
+func (sm *streamManager) startOVNKIPFIXCollector(ctx context.Context, logger *zap.Logger) error {
+	logger.Info("Starting OVN-K IPFIX Collector")
+	ctxProcessOVNKFlow, cancelProcessOVNKFlow := context.WithCancel(ctx)
+	defer cancelProcessOVNKFlow()
+	go func() {
```
rlenglet, line 64: Why does this need to be done in a separate goroutine?
That brings nothing, it's unnecessary complexity.
Just call `processIPFIXMessage` after receiving each message, synchronously.

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+			case <-ctx.Done():
+				logger.Info("Context canceled in OVN-K listener loop")
+				return ctx.Err()
+			default:
+				n, _, err := listener.ReadFrom(buf)
+				if err != nil {
+					logger.Error("Failed to read from OVN-K listener", zap.Error(err))
+					return err
+				}
+				sm.streamClient.ovnkEventChan <- buf[:n]
```
rlenglet, line 94: You need to do this the other way around, like for the other streams.
This function should:
1. Start a goroutine that does the blocking reads from the socket, i.e. which should:
    1.  Call `ReadFrom` to receive a message.
    1. Parse the message, _in the same goroutine_ (doing that in a separate goroutine brings nothing).
    1. Send the message into the cache.
1. Wait for the context to get canceled, and in this case:
    1. Close the listener, which should unblock any blocked call to `ReadFrom` and cause it to exit.

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+
+		logger.Info("OVN-k server listening", zap.String("address", sm.streamClient.IPFIXCollectorPort))
+		for {
+			buf := make([]byte, ipfixMaxLength)
+			select {
+			case <-ctx.Done():
+				logger.Info("Context canceled in OVN-K listener loop")
+				return ctx.Err()
+			default:
+				n, _, err := listener.ReadFrom(buf)
```
rlenglet, line 89: This can block indefinitely, until the listener is closed.
The problem is that nothing in this code is closing the listener.
The listener should be closed when the context is canceled.
So the only proper way to do this is to do these `ReadFrom` blocking calls in a separate goroutine, and have this main function here wait for the context to be canceled and then close the listener, which will unblock any `ReadFrom` blocking call in the goroutine.`

internal/controller/ovnk_ipfix_collector.go
rlenglet, line 76: _**This listener MUST be closed.**_ Right now, this listener is leaked. And more importantly, closing the listener is the only way to unblock a call to `ReadFrom` on it. Cf. [the documentation of `PacketConn.Close`](ListenPacket):
```
	// Close closes the connection.
	// Any blocked ReadFrom or WriteTo operations will be unblocked and return errors.
	Close() error
```

rlenglet (changes requested):

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+				return OVNKFlow{}, err
+			}
+			ovnkFlow.Protocol = protocol
+		case 60: // ipVersion
+			ipVersion, err := parseIPVersion(field.Value.([]byte))
+			if err != nil {
+				return OVNKFlow{}, err
+			}
+			ovnkFlow.IPVersion = ipVersion
+		case 158: // start time delta
```
rlenglet, line 283:
```suggestion
		case 158: // flowStartDeltaMicroseconds
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+			startTimeDelta := binary.BigEndian.Uint32(field.Value.([]byte))
+			ovnkFlow.StartTimestamp = timestamppb.New(time.Unix(int64(exportTime), 0).Add(-time.Duration(startTimeDelta) * time.Second))
```
rlenglet, line 285: The `flowStartDeltaMicroseconds` element is in microseconds, not seconds, as the name indicates.
```suggestion
			ovnkFlow.StartTimestamp = timestamppb.New(time.Unix(int64(exportTime), 0).Sub(startTimeDelta * time.Microseconds))
```

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,291 @@
...
+		default:
```
rlenglet, line 286: Also record the flow end timestamp as we should update the API to send that for every flow at some point in the future.
```suggestion
		case 159: // flowEndDeltaMicroseconds
			// ...
		default:
```

internal/controller/streams.go
```diff
@@ -444,8 +448,8 @@ func (sm *streamManager) StreamFalcoNetworkFlows(ctx context.Context, logger *za
 				continue
 			}
 
-			convertedFalcoFlow, err := parsePodNetworkInfo(match[1])
-			if convertedFalcoFlow == nil {
+			convertedStandardFlow, err := parsePodNetworkInfo(match[1])
```
rlenglet, line 451: "standard" -> "five-tuple" to reflect the changes to the API.

internal/controller/streams_helper_test.go
```diff
@@ -130,8 +130,8 @@ func TestSendNetworkFlowRequest(t *testing.T) {
 		assert.Equal(t, expected, mockStream.lastRequest)
 	})
 
-	t.Run("falco flow", func(t *testing.T) {
-		flow := &pb.FalcoFlow{
+	t.Run("standard flow", func(t *testing.T) {
```
rlenglet, line 133:
```suggestion
	t.Run("five-tuple flow", func(t *testing.T) {
```

internal/controller/ovn_template_sets/template_set_0_1747676841915736072.bin
evanj80-illumio: I opted to get rid of this, will add comments into the configmap

internal/controller/ovnk_ipfix_collector.go
evanj80-illumio: I think this is just extra noise, so im going to get rid of this log.

internal/controller/streams.go
```diff
@@ -37,16 +37,17 @@ const (
...
-	flowCollector             pb.FlowCollector
-	logStream                 pb.KubernetesInfoService_SendLogsClient
-	networkFlowsStream        pb.KubernetesInfoService_SendKubernetesNetworkFlowsClient
-	resourceStream            pb.KubernetesInfoService_SendKubernetesResourcesClient
-	configStream              pb.KubernetesInfoService_GetConfigurationUpdatesClient
+	ciliumNamespace    string
+	conn               *grpc.ClientConn
+	client             pb.KubernetesInfoServiceClient
+	falcoEventChan     chan string
+	ovnkEventChan      chan []byte
```
rlenglet, line 44: You don't need this anymore, correct?
```suggestion
```

internal/controller/streams.go
```diff
@@ -90,6 +91,10 @@ type EnvironmentConfig struct {
 	OnboardingClientSecret string
 	// URL of the onboarding endpoint.
 	OnboardingEndpoint string
+	// Port for the IPFIX collector
+	IPFIXCollectorPort string
+	// Namespace of OVN-Kubernetes
+	OvnkNamespace string
```
rlenglet, line 97:
```suggestion
	OVNKNamespace string
```

internal/controller/streams.go
```diff
@@ -761,11 +817,12 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
...
-				conn:                      authConn,
-				client:                    client,
-				ciliumNamespace:           envMap.CiliumNamespace,
-				disableNetworkFlowsCilium: false,
-				falcoEventChan:            falcoEventChan,
+				conn:               authConn,
+				client:             client,
+				ciliumNamespace:    envMap.CiliumNamespace,
+				falcoEventChan:     falcoEventChan,
+				ovnkEventChan:      make(chan []byte),
```
rlenglet, line 824: You don't need this anymore, correct?
```suggestion
```

rlenglet (changes requested):
Looks good! A few more less critical comments.

cloud-operator/patch_file.yaml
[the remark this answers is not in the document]
evanj80-illumio: Sorry this file was not intended to be pushed.

internal/controller/ovnk_ipfix_collector.go
```diff
@@ -0,0 +1,314 @@
...
+// Returns an error if the slice is not the correct size or the protocol is unknown.
+func parseProtocol(decodedValue []byte) (string, error) {
+	if len(decodedValue) < 1 {
+		return "", errors.New("insufficient data to parse protocol")
+	}
+
+	protocol := uint8(decodedValue[0])
+	switch protocol {
+	case 1:
+		return "icmpt", nil
```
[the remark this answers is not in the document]
evanj80-illumio: Added a shared_constants file and replaced in repo

internal/controller/streams.go
```diff
@@ -37,16 +37,16 @@ const (
...
-	flowCollector             pb.FlowCollector
-	logStream                 pb.KubernetesInfoService_SendLogsClient
-	networkFlowsStream        pb.KubernetesInfoService_SendKubernetesNetworkFlowsClient
-	resourceStream            pb.KubernetesInfoService_SendKubernetesResourcesClient
-	configStream              pb.KubernetesInfoService_GetConfigurationUpdatesClient
+	ciliumNamespace    string
+	conn               *grpc.ClientConn
+	client             pb.KubernetesInfoServiceClient
+	falcoEventChan     chan string
+	IPFIXCollectorPort string
```
rlenglet, line 44: Nit _(not blocking the PR)_: Do you really need to export this field?
```suggestion
	ipfixCollectorPort string
```

rlenglet (approved):
Looks great, thanks!

cmd/main.go
```diff
@@ -98,12 +101,14 @@ func main() {
 	bindEnv(logger, "pod_namespace", "POD_NAMESPACE")
 	bindEnv(logger, "stream_success_period_connect", "STREAM_SUCCESS_PERIOD_CONNECT")
 	bindEnv(logger, "stream_success_period_auth", "STREAM_SUCCESS_PERIOD_AUTH")
-	bindEnv(logger, "https_proxy", "HTTPS_PROXY")
 
 	// Set default values
 	viper.SetDefault("cluster_creds", "clustercreds")
 	viper.SetDefault("cilium_namespace", "kube-system")
+	viper.SetDefault("https_proxy", "")
+	viper.SetDefault("ipfix_collector_port", ":4739")
```
rlenglet, line 109:
```suggestion
	viper.SetDefault("ipfix_collector_port", "4739")
```

cloud-operator/templates/cloud-operator-deployment.yaml
```diff
@@ -52,6 +52,7 @@ spec:
           ports:
             - containerPort: 8080
             - containerPort: 5000
+            - containerPort: 4739
```
rlenglet, line 55:
```suggestion
            - containerPort: 4739
              protocol: UDP
```

fakeserver/server.go
```diff
@@ -224,7 +224,7 @@ func (s *server) SendKubernetesNetworkFlows(stream pb.KubernetesInfoService_Send
 		switch req.Request.(type) {
 		case *pb.SendKubernetesNetworkFlowsRequest_CiliumFlow:
 			logger.Info("Received CiliumFlow")
-		case *pb.SendKubernetesNetworkFlowsRequest_FalcoFlow:
+		case *pb.SendKubernetesNetworkFlowsRequest_FiveTupleFlow:
 			logger.Info("Received FalcoFlow")
```
rlenglet, line 228:
```suggestion
			logger.Info("Received FiveTupleFlow")
```

rlenglet (changes requested):

rlenglet (approved):

rlenglet (dismissed):

internal/controller/streams.go
```diff
@@ -45,12 +45,12 @@ type streamClient struct {
 	client             pb.KubernetesInfoServiceClient
 	falcoEventChan     chan string
 	ipfixCollectorPort string
+  tlsAuthProperties         tls.AuthProperties
```
rlenglet: Reformat the file to align.

rlenglet (dismissed):

aaronIllumio (approved):
LPTM

PR #199, opened 2025-04-29
JSON Schema for Helm Chart Values Validation

pavankumarinnamuri:


cloud-operator/values.schema.json
```diff
@@ -0,0 +1,212 @@
...
+    "priorityClassName": {
+      "type": "string",
+      "description": "Priority class name for the operator pods"
+    },
+    "image": {
+      "type": "object",
+      "properties": {
+        "repository": {
+          "type": "string",
+          "description": "Docker image repository"
```
evanj80-illumio, line 23:
```suggestion
          "description": "container image repository"
```

cloud-operator/values.schema.json
```diff
@@ -0,0 +1,212 @@
...
+        "annotations": {
+          "type": "object",
+          "description": "Annotations for the service account"
+        },
+        "automount": {
+          "type": "boolean",
+          "description": "Whether to automount the service account token"
+        }
+      },
+      "required": ["create", "name"]
```
evanj80-illumio, line 57: I think automount should be in here as well

cloud-operator/values.schema.json
```diff
@@ -0,0 +1,212 @@
...
+    },
+    "resources": {
```
evanj80-illumio, line 59: I think this should also have a required field. We need to always limit the resources our operator consumes

cloud-operator/values.schema.json
evanj80-illumio, line 1: I think most things in this schema validation should be required.

evanj80-illumio (changes requested):

cloud-operator/tests/cloud-operator-deployment_test.yaml
```diff
@@ -6,7 +6,7 @@ tests:
   # Test for environment variables
   - it: should set the correct environment variables
     set:
-      env.tlsSkipVerify: "true"
```
evanj80-illumio, line 9: We should be careful switching the types here. I think this may cause issues

.github/workflows/update-schema.yml
```diff
@@ -0,0 +1,34 @@
...
+        uses: losisin/helm-values-schema-json-action@v1
+        with:
+          path: cloud-operator
+          values-file: values.yaml
+          schema-file: values.schema.json
+
+      - name: Check for changes
+        id: git-check
+        run: |
+          git diff --exit-code cloud-operator/values.schema.json || echo "changes=true" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
evanj80-illumio: Bump on this

cloud-operator/tests/cloud-operator-deployment_test.yaml
evanj80-illumio: Bump on this. please make sure if you change this type it will not cause downstream effects on tlsSkipVerify in the operator

pavankumarinnamuri: i think  Changing the test to use a boolean value aligns with the schema definition and it doesn't affect the runtime behavior since the template  handles the type conversions.

AriSweedler: environment variables are always strings. Even when the data is an int or float or bool or `time.Duration`  or whatever. Env variables get read as strings and then parsed into the type they actually are.

So this looks strange to me, seeing a non-string `env.*` variable.

.github/workflows/update-schema.yml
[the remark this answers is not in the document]
AriSweedler: The bad scenario is that we update a values.yaml file without updating the schema, and don't catch it until a customer runs into an issue.

The problem I want this gha to solve is: If we update the shape of the values.yaml file, how can we make sure that the schema file keeps up?

Maybe it's not so easy as we have customer statements around what is required and what is not, what is not easy to generate from the values.yaml file. But I imagine there's a solution to even that problem.

AriSweedler:
> As part of this PR it would be awesome if you could create a helm lint GHA similar to the unit test one. That is how we will enforce our schema for ourselves.

@evanj80-illumio a separate GHA for that is not necessary:

https://helm.sh/docs/topics/charts/#schema-files

> This schema will be applied to the values to validate it. Validation occurs when any of the following commands are invoked:
>
> * helm install
> * helm upgrade
> * helm lint
> * helm template

.github/workflows/update-schema.yml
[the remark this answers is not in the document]
evanj80-illumio: My only worry is if you let a bot do it, then someone could put something that is wrong? But you are probably right, allowing a bot to update the schema should be okay since we would still have to review what the changes are to the schema. Lets get this merged in and @pavankumarinnamuri make a new PR with this change accounted for.

evanj80-illumio (approved):

.github/workflows/update-schema.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - name: Fail if schema needs updating
+        if: steps.git-check.outputs.changes == 'true'
+        run: |
+          echo "::error::The values.schema.json file needs to be updated to match changes in values.yaml."
+          echo "::error::Please run 'helm schema-gen values.yaml -o values.schema.json' or similar tool to update the schema."
+          exit 1
+
```
pavankumarinnamuri, line 46: include two jobs: one for validating schema changes on pull requests and another for automatically updating the schema on the main branch

.github/workflows/update-schema.yml
```diff
@@ -0,0 +1,81 @@
...
+      - 'cloud-operator/values.yaml'
+  workflow_dispatch:
+
+jobs:
+  # Job for PR validation - keeps contents: read permission
+  validate-schema:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    if: github.event_name == 'pull_request'
```
evanj80-illumio, line 18: Im not sure if you need this since you have it above with 
on:
   pull_request:

evanj80-illumio (dismissed):

.github/workflows/update-schema.yml
```diff
@@ -37,8 +38,44 @@ jobs:
...
+        uses: losisin/helm-values-schema-json-action@v1
+        with:
+          path: cloud-operator
+          values-file: values.yaml
+          schema-file: values.schema.json
+
+      - name: Check for changes
+        id: git-check
+        run: |
+          git diff --exit-code cloud-operator/values.schema.json || echo "changes=true" >> $GITHUB_OUTPUT
```
AriSweedler, line 72: This job updates the schema, and if it changes then it states that changes are true. It will then make a commit to main.

Whereas the `validate-schema` function will FAIL if there are changes. Idk GHA syntax very well, but it feels strange that you are running the same 4 steps & then doing something different. I can't quite visualize the dependency graph. Won't these both just run in parallel? The `if` filters you set up (push, workflow_dispatch) vs. (pull_request) are good enough?

AriSweedler: Is there any way to define the 4 steps (checkout, helm, generate, check for changes) as a reusable component?

pavankumarinnamuri: made changes and removed the duplicated

.github/workflows/update-schema.yml
AriSweedler, line 1: This file is not just `update-schema`, but also `validate-schema`. And `schema` is too broad a term. Every `proto` file is a schema. The values schema json file is also a schema. Thus I would recommend that you rename this file to: `values-schema-json` so there is no doubt as to which schema it works with.

AriSweedler (dismissed):
I'll approve but I think you should still change the filename. I also am confused as I don't see the `./.github/actions/generate-schema` file 🤔 But the build succeeded

AriSweedler (approved):

PR #205, opened 2025-05-07
changes to buildResourceApiGroupMap

ganesh-talla:
Excluded metrics.k8s.io API group in buildResourceApiGroupMap as the service account has no permissions to read this.


ganesh-talla:
@evanj80-illumio Do we want this configurable?

evanj80-illumio (approved):

PR #213, opened 2025-05-13
 resource manager for each watcher

pavankumarinnamuri:


internal/controller/streams.go
```diff
@@ -262,10 +256,18 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
-
-		resourceVersion, err := resourceLister.DynamicListResources(ctx, resourceLister.logger, resource, apiGroup)
+		// Create a new resource manager for each resource type
+		resourceManager := NewResourceManager(ResourceManagerConfig{
+			ResourceName:  resource,
+			Clientset:     clientset,
+			BaseLogger:    logger,
+			DynamicClient: dynamicClient,
+			StreamManager: sm,
+			Limiter:       rate.NewLimiter(1, 5),
```
evanj80-illumio, line 266: I think this limiter may need to be global to all of the resourceManagers. if I remember correctly this limiter is to limit the rate in which we hit the k8s api server on our watch calls. By making this per resource it no longer does any useful rate limiting.

internal/controller/streams.go
```diff
@@ -262,10 +256,18 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
+		})
+		resourceVersion, err := resourceManager.DynamicListResources(ctx, resourceManager.GetLogger(), apiGroup)
 		if err != nil {
-			resourceLister.logger.Error("Failed to list resource", zap.String("resource", resource), zap.Error(err))
+			resourceManager.GetLogger().Error("Failed to list resource", zap.Error(err))
```
evanj80-illumio, line 270: I dont see the point of this GetLogger() function if it just returns resourceManager.logger. Usually a getter function like this is safe to call if the logger would be nil. However, if that is what you intend to do you may need to rewrite how we exactly call GetLogger().Error........

internal/controller/streams.go
```diff
@@ -279,16 +281,25 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
 	}
 	logger.Info("Successfully sent resource snapshot")
 
 	// PHASE 3: Start watchers concurrently
 	for _, info := range allWatchInfos {
-		go func(info watcherInfo) {
-			resourceLister.WatchK8sResources(ctx, cancel, info.resource, info.apiGroup, info.resourceVersion)
-		}(info)
+		// Create a new resource manager for each watcher
+		watchResourceManager := NewResourceManager(ResourceManagerConfig{
```
evanj80-illumio, line 292: Why recreate these same ResourceManagers for the watchers after? I think the only difference is the casing of the resource so maybe just overwrite the existing resourceManagers resource?

internal/controller/resource_manager.go
```diff
@@ -30,11 +42,31 @@ type ResourceManager struct {
...
+		logger:        logger,
+		dynamicClient: config.DynamicClient,
+		streamManager: config.StreamManager,
+		limiter:       config.Limiter,
+	}
+}
+
+// GetLogger returns the logger with resource context
+func (r *ResourceManager) GetLogger() *zap.Logger {
+	return r.logger
```
evanj80-illumio, line 62: This function has no value.
`
func (x *Flow) GetTime() *timestamppb.Timestamp {
	if x != nil {
		return x.Time
	}
	return nil
}
`

Example of a Get function that has value above (from cilium source code)

internal/controller/streams.go
```diff
@@ -279,16 +286,21 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
 	// PHASE 2: Send snapshot complete
 	err = sm.sendResourceSnapshotComplete(logger)
 	if err != nil {
-		resourceLister.logger.Error("Failed to send snapshot complete", zap.Error(err))
+		logger.Error("Failed to send snapshot complete", zap.Error(err))
 		return err
 	}
 	logger.Info("Successfully sent resource snapshot")
 
 	// PHASE 3: Start watchers concurrently
```
evanj80-illumio, line 294: Im actually not convinced there is any difference between the resource being passed into DynamicListResources vs WatchK8sResources. Can you double check? That would make the extra steps from 296 - 300 not needed and would simplify things

evanj80-illumio (changes requested):
Once change! Looking good!

evanj80-illumio (approved):

internal/controller/resource_manager.go
```diff
@@ -16,8 +16,20 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+// ResourceManagerConfig holds the configuration for creating a new ResourceManager
+type ResourceManagerConfig struct {
+	ResourceName  string
```
AriSweedler, line 21: Can you make this a proper type?

```
type ResourceName string

const (
  X = ResourceName("X")
  Y = ResourceName("Y")
  ...
```

Adds some measure of type safety, and gives a developer a concept of what's correct. (`"Namespace"` is NOT a Resource Name, but `"namespace"` IS)

internal/controller/resource_manager.go
```diff
@@ -30,11 +42,26 @@ type ResourceManager struct {
 	limiter *rate.Limiter
 }
 
+// NewResourceManager creates a new ResourceManager for a specific resource type.
+// The logger will automatically include the resource name in all log messages.
+func NewResourceManager(config ResourceManagerConfig) *ResourceManager {
```
AriSweedler, line 47: Can just call this `New`. Romain just recommended I read through https://google.github.io/styleguide/go/best-practices.html#avoid-repetition, and I think it's great, I recommend you read thru it too.

internal/controller/streams.go
```diff
@@ -240,32 +240,39 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
-	}
+
 	err = sm.sendClusterMetadata(ctx, logger)
 	if err != nil {
 		logger.Error("Failed to send cluster metadata", zap.Error(err))
 		return err
 	}
 	allWatchInfos := make([]watcherInfo, 0, len(resourceAPIGroupMap))
 
-	// PHASE 1: List all resources in deterministic order
```
AriSweedler, line 257: Don't remove the `PHASE 1` comment without removing the `PHASE 2` comment. Need consistency - keep both or remove/reword both.

internal/controller/streams.go
```diff
@@ -240,32 +240,39 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
+	resourceManagers := make(map[string]*ResourceManager)
 	for _, resource := range slices.Sorted(maps.Keys(resourceAPIGroupMap)) {
 		apiGroup := resourceAPIGroupMap[resource]
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 		default:
 		}
+		// Create a new resource manager for each resource type
+		resourceManager := NewResourceManager(ResourceManagerConfig{
```
AriSweedler, line 263: If you change the name of this function to `New`, then you can pass this struct as a recv'er, or as an argument. I think as a recv'er makes more sense. 
```
		(ResourceManagerConfig{
			ResourceName:  resource,
			Clientset:     clientset,
			BaseLogger:    logger,
			DynamicClient: dynamicClient,
			StreamManager: sm,
			Limiter:       sharedLimiter,
		}).New()
```

AriSweedler (commented):
Looks good! Just have a few suggestions

PR #214, opened 2025-05-13
Fixing flaky unit tests

evanj80-illumio:
Adding retry logic for test cleanup. Ran GHA 9 times and no failures

### DIFF
--- a/internal/controller/authenticator_test.go
+++ b/internal/controller/authenticator_test.go
@@ -11,6 +11,7 @@ import (
 	"net/url"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


internal/controller/authenticator_test.go
```diff
@@ -193,6 +194,19 @@ func (suite *ControllerTestSuite) TestWriteK8sSecret() {
 
 			// Since go test does not follow any order, always make sure namespace is deleted before each test
 			_ = suite.clientset.CoreV1().Namespaces().Delete(context.TODO(), "illumio-cloud", metav1.DeleteOptions{})
+
+			// Add polling logic to ensure namespace deletion
```
AriSweedler, line 198: This is an example of a mix of action & control flow. Translating this code to english it is:
1. Do an action
2. poll until the action is finished

The action is:
```
suite.clientset.CoreV1().Namespaces().Delete(context.TODO(), "illumio-cloud", metav1.DeleteOptions{})
```
The poll is like:
```
func() {
suite.clientset.CoreV1().Namespaces().Get(context.TODO(), "illumio-cloud", metav1.GetOptions{})
returns IsNotFound
}
```

I thiiiink it makes sense to have a quick helper function that accepts an action & a poll & an optional config (only config variable is `how many times do we try before giving up` and `how do we wait in between sleeps`). This could actually extend the `exponential backoff` lib that I added a few weeks ago. Instead of
```
func exponentialBackoff(opts backoffOpts, action Action) error {
```

it will be 
```
func poll(opts backoffOpts, poll Action) error {
```

And then the action just gets inlined above it.

AriSweedler: Alternatively, we could have a set of testutils with functions that make deletions like this synchronous. I'll open up a PR for that rq...

internal/controller/k8s_resources_test.go
```diff
@@ -963,6 +964,16 @@ func (suite *ControllerTestSuite) TestConvertToKubernetesServiceData() {
 	for name, tt := range tests {
 		suite.Run(name, func() {
 			ctx := context.TODO()
+
+			// Ensure the service is deleted before running the test
+			if tt.service == nil {
```
AriSweedler, line 969: This is similar logic as before - these `.Delete` functions return when the call has been launched, not when the promise has been fulfilled

AriSweedler (approved):
Approved! Looks good to merge. I'll make that PR to delete synchronously

PR #221, opened 2025-05-19
Did error handling to identify Access forbidden errors

ganesh-talla:
Added check to check for forbidden errors in buildResourceApiGroupMap


internal/controller/streams.go
```diff
@@ -232,7 +235,6 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
 	//This builds resourceAPIGroupMap from Kubernetes API
 	resourceAPIGroupMap, err = sm.buildResourceApiGroupMap(resources, clientset, logger)
 	if err != nil {
-		logger.Error("Failed to build resource api group map", zap.Error(err))
```
evanj80-illumio, line 235: I think its still a good idea to log this

internal/controller/streams.go
```diff
@@ -272,7 +274,11 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
 
 		resourceVersion, err := resourceManager.DynamicListResources(ctx, resourceManager.logger, apiGroup)
 		if err != nil {
-			resourceManager.logger.Error("Failed to list resource", zap.Error(err))
+			if strings.Contains(err.Error(), "access forbidden") {
+				logger.Info("Access forbidden for resource", zap.String("resource", resource), zap.String("apiGroup", apiGroup), zap.Error(err))
```
evanj80-illumio, line 278: I think I would make this a debug log

evanj80-illumio (changes requested):

internal/controller/streams.go
ganesh-talla: Resolved this.

internal/controller/streams.go
ganesh-talla: made this a debug log

internal/controller/streams.go
```diff
@@ -181,16 +183,17 @@ func (sm *streamManager) buildResourceApiGroupMap(resources []string, clientset
...
-		if group.Name == "metrics.k8s.io" {
-			continue
-		}
-
 		for _, version := range group.Versions {
 			resourceList, err := discoveryClient.ServerResourcesForGroupVersion(version.GroupVersion)
 			if err != nil {
-				logger.Error("Error fetching resources for groupVersion", zap.Error(err))
-				return resourceAPIGroupMap, err
+				if apiErr, ok := err.(*apierrors.StatusError); ok && apiErr.ErrStatus.Code == 403 {
```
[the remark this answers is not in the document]
ganesh-talla: @rlenglet The reason I can't use errors.Is is the error cannot be deterministically said for future api groups as I am writing this to exclude several api-groups and resources the illumio-cloud service account has no access to. 

But I can use this package in "k8s.io/apimachinery/pkg/api/errors" IsForbidden(err) [https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsForbidden](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsForbidden)  to determine if the error is related to forbidden api groups.

and in errors.Is we need to compare it with error type I can do this if we know only which error we need to handle gracefully. @evanj80-illumio Does this need to handle all other errors like this or only the errors we saw till now.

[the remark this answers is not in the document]
evanj80-illumio: I think it should cover all errors that could show up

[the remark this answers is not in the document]
rlenglet: Yeah, using `IsForbidden(err)` is a good idea.

rlenglet (approved):
Looks great, thanks!

internal/controller/streams.go
```diff
@@ -272,7 +271,10 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
 
 		resourceVersion, err := resourceManager.DynamicListResources(ctx, resourceManager.logger, apiGroup)
 		if err != nil {
-			resourceManager.logger.Error("Failed to list resource", zap.Error(err))
+			if apierrors.IsForbidden(err) {
+				logger.Warn("Access forbidden for resource", zap.String("resource", resource), zap.String("apiGroup", apiGroup), zap.Error(err))
```
rlenglet, line 275: Nit. We consistently use camel_case for structured log fields.
```suggestion
				logger.Warn("Access forbidden for resource", zap.String("resource", resource), zap.String("api_group", apiGroup), zap.Error(err))
```

rlenglet (changes requested):

PR #222, opened 2025-05-20
Added seperate auth conn and client to different streams

ganesh-talla:
This PR refactors the ConnectStreams function to use separate connections and clients for each stream (resources, logs, configuration, and network). 
**Before**
1. A single connection and client were shared across all streams (resources, logs, configuration updates, and network flows).
2. This approach made it harder to isolate and debug issues specific to stream overwhelming.
3. Stream bursts could potentially affect other streams due to shared oauth2 GRPC connection.

**After**

Separate connections and clients are now created for each stream:
  > Resource Stream: resourceAuthConn and resourceClient
  > Logs Stream: logsAuthConn and logsClient
  > Configuration Updates Stream: configAuthConn and configClient
  > Network Flows Stream: flowsAuthConn and flowsClient

### DIFF
--- a/internal/controller/streams.go
+++ b/internal/controller/streams.go
@@ -37,9 +37,19 @@ const (
 )
 
 type streamClient struct {
-	ciliumNamespace           string
-	conn                      *grpc.ClientConn
-	client                    pb.KubernetesInfoServiceClient
+	ciliumNamespace string
+	conn            *grpc.ClientConn
+	resourceConn    *grpc.ClientConn
+	flowsConn       *grpc.ClientConn
+	logsConn        *grpc.ClientConn
+	configConn      *grpc.ClientConn
+
+	resourceClient pb.KubernetesInfoServiceClient
+	flowsClient    pb.KubernetesInfoServiceClient
+	logsClient     pb.KubernetesInfoServiceClient
+	configClient   pb.KubernetesInfoServiceClient
+	client         pb.KubernetesInfoServiceClient
+
 	disableNetworkFlowsCilium bool
 	falcoEventChan            chan string
 	flowCollector             pb.FlowCollector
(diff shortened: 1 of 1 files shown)


internal/controller/streams.go
```diff
@@ -881,8 +953,20 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
...
 }
 
-// NewAuthenticatedConnection gets a valid token and creats a connection to CloudSecure.
-func NewAuthenticatedConnection(ctx context.Context, logger *zap.Logger, envMap EnvironmentConfig) (*grpc.ClientConn, pb.KubernetesInfoServiceClient, error) {
+// GetAuthConnAndClient gets a connection and client for the given credentials.
+func GetAuthConnAndClient(ctx context.Context, logger *zap.Logger, envMap EnvironmentConfig, clientID string, clientSecret string, resetTimer *time.Timer) (*grpc.ClientConn, pb.KubernetesInfoServiceClient, error) {
+	authConn, err := SetUpOAuthConnection(ctx, logger, envMap.TokenEndpoint, envMap.TlsSkipVerify, clientID, clientSecret)
+	if err != nil {
+		logger.Error("Failed to set up an OAuth connection", zap.Error(err))
+		resetTimer.Reset(jitterTime(10*time.Second, 0.20))
```
evanj80-illumio, line 961: Whats the point of this jitter everywhere?

ganesh-talla: @evanj80-illumio I am not sure actually it is not required but the existing code has that jitter if the authconn or client failed so thought of adding this as it is the same logic.

		case <-resetTimer.C:
			authConContext, authConContextCancel := context.WithCancel(ctx)
			authConn, client, err := NewAuthenticatedConnection(authConContext, logger, envMap)
			if err != nil {
				logger.Error("Failed to establish initial connection; will retry", zap.Error(err))
				// When we try this loop again, we wait 10 seconds with 20% jitter.
				resetTimer.Reset(jitterTime(10*time.Second, 0.20))
				authConContextCancel()
				failureReason = "Failed to establish initial connection"
				break
			}

ganesh-talla: I feel that this is not required at all as we don't contend for connections

internal/controller/streams.go
```diff
@@ -760,9 +772,68 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
...
+				resetTimer.Reset(jitterTime(10*time.Second, 0.20))
+				authConContextCancel()
+				failureReason = "Failed to set up an OAuth connection"
+				break
+			}
+
+			// Get config connection and client
+			configAuthConn, configClient, err := GetAuthConnAndClient(ctx, logger, envMap, clientID, clientSecret, resetTimer)
+			if err != nil {
+				logger.Error("Failed to set up an OAuth connection", zap.Error(err))
```
evanj80-illumio, line 809: Would it be possible to take all of these error catching that you are doing and put it into a function? Seems like almost the exact same code per client error

PR #227, opened 2025-06-04
Refactor ResourceManager to Follow Go Best Practices

pavankumarinnamuri:


internal/controller/resource_manager.go
```diff
@@ -17,9 +17,56 @@ import (
...
+	ClusterRoleBinding ResourceName = "clusterrolebindings"
+)
+
+// String returns the string representation of the ResourceName
+func (r ResourceName) String() string {
+	return string(r)
+}
+
+// IsValid returns true if the ResourceName is a valid Kubernetes resource type
+func (r ResourceName) IsValid() bool {
```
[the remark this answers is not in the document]
pavankumarinnamuri: IsValid() method is not used anywhere ,but its need if we validate dynamically created ResourceName values

internal/controller/resource_manager.go
```diff
@@ -17,9 +17,56 @@ import (
...
+	switch r {
+	case Pod, Node, Service, Namespace, ConfigMap, Secret, ServiceAccount,
```
evanj80-illumio, line 57: This case statement seems weird, like all of these are good and anything else is bad, why cant we have doing the opposite logic?

internal/controller/resource_manager.go
```diff
@@ -17,9 +17,56 @@ import (
...
+
+	// RBAC resources
+	Role               ResourceName = "roles"
+	RoleBinding        ResourceName = "rolebindings"
+	ClusterRole        ResourceName = "clusterroles"
+	ClusterRoleBinding ResourceName = "clusterrolebindings"
+)
+
+// String returns the string representation of the ResourceName
+func (r ResourceName) String() string {
```
evanj80-illumio, line 50: This seems odd, they are always strings? Why do I need a function like this.

internal/controller/resource_manager.go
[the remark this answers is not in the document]
evanj80-illumio: I also dont understand what this is even doing

evanj80-illumio (changes requested):
Based on this PR I am not quite understanding what is the point of changing what we already have, What best practices are we following and why does it make maintaining this any better?

internal/controller/resource_manager.go
rlenglet, line 1: This looks similar to the code generated by the go-enum package.
Why not use that directly?
https://github.com/abice/go-enum

internal/controller/resource_manager.go
```diff
@@ -43,18 +90,18 @@ type ResourceManager struct {
 	limiter *rate.Limiter
 }
 
-// NewResourceManager creates a new ResourceManager for a specific resource type.
+// New creates a ResourceManager from the config.
 // The logger will automatically include the resource name in all log messages.
-func NewResourceManager(config ResourceManagerConfig) *ResourceManager {
+func (c ResourceManagerConfig) New() *ResourceManager {
```
rlenglet, line 95: OK with making this a method on `NewResourceManager`.
But why rename this into `New`?
`configblah.New()` doesn't make sense to me.

rlenglet (changes requested):

PR #231, opened 2025-06-10
Handle missing tls handshake error

aaronIllumio:
It is possible that cert files are generated for hubble relay, but tls verification is disabled.

This change tries tls connection first then disables it if it fails

### DIFF
--- a/internal/controller/streams.go
+++ b/internal/controller/streams.go
@@ -164,10 +164,10 @@ func ServerIsHealthy() bool {
 }
 
 // disableSubsystemCausingError mutates the `streamManager`
-// receiver by inspecting the error. Thanks to HandleTLSHandshakeError(1)
+// receiver by inspecting the error. Thanks to HandleTLSHandshakeError
 // we have a nicely typed error. If we recognize the specific error,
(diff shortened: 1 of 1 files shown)


internal/controller/cilium_flow_collector.go
```diff
@@ -161,6 +161,10 @@ func (fm *CiliumFlowCollector) exportCiliumFlows(ctx context.Context, sm *stream
 			fm.logger.Error("ALPN handshake failed, retrying with ALPN disabled")
 			return tls.ErrTLSALPNHandshakeFailed
 		}
+		if strings.Contains(err.Error(), "first record does not look like a TLS handshake") {
```
AriSweedler, line 164: Do you foresee these 2 cases expanding to 3 or 4? I think that converting the `ovserverClient`'s error into a `tls.Err*` is a hoistable thunk. It can live in the `tls` package & even be unit tested `tls.ErrorFrom<Type>`

rlenglet (dismissed):

rlenglet (dismissed):

AriSweedler (approved):

PR #234, opened 2025-06-18
Adding verboseDebugging value in helm chart

evanj80-illumio:
1. Adding verboseDebugging for troubleshooting with users.
2. Upping version to 1.24.4 due to CVE.


internal/controller/resource_manager.go
```diff
@@ -104,6 +105,7 @@ func (r *ResourceManager) DynamicListResources(ctx context.Context, logger *zap.
 			return "", err
 		}
 	}
+	r.logger.Debug("Succesfully sent K8s workloads", zap.Int("count", len(objs)))
```
AriSweedler, line 108: You're gunna wanna remove all these logs in the next version. Perhaps add a tag or a string to them so that's easier. You can of course also just use `git revert`, but... Well there are some changes in here that you will want to keep. Up to you

internal/controller/streams.go
```diff
@@ -386,7 +389,11 @@ func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger
 				logger.Info("Received configuration update",
 					zap.Stringer("log_level", update.UpdateConfiguration.LogLevel),
 				)
-				sm.bufferedGrpcSyncer.updateLogLevel(update.UpdateConfiguration.LogLevel)
+				if sm.verboseDebugging {
+					sm.bufferedGrpcSyncer.updateLogLevel(pb.LogLevel_LOG_LEVEL_DEBUG)
```
AriSweedler, line 393: Log when this happens "verbose debugging override is set"

AriSweedler (approved):

internal/controller/resource_manager.go
evanj80-illumio: I think its actually fine to keep these going forward. These debug logs should actually only be sent if debug is turned on via api or this values flag. Its more of a troubleshooting step.

aaronIllumio (approved):
Lgtm

PR #239, opened 2025-06-23
Add nodeSelector, tolerances, and affinity to deployment

AriSweedler:
This will give the user the ability to configure these elements of their helm deployment without relying on patching


evanj80-illumio:
Ignore grype scan for now. Am fixing it in a different PR

PR #248, opened 2025-07-16
Addressing server side rate limit + adding mutexes to stream

evanj80-illumio:
1. Adding server side rate limiting, with local testing with fakeserver
2. Adding proper mutexes to protect stream contention between keep_alives and real messages.
3. Modifying the logger in order to make sure keep_alives and logs have no contention.
4. Preventing flow stream from starting UNLESS resource stream has committed resources.


internal/controller/streams.go
```diff
@@ -426,24 +432,57 @@ func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger
...
+
+	once := &sync.Once{} // Ensure networkFlowStreamOpen is triggered only once
+
 	for {
 		select {
 		case <-ctx.Done():
 			return ctx.Err()
 		case flow := <-sm.FlowCache.outFlows:
+			if flowCount >= allowedFlows {
+				continue // Wait for a new ServerWindow message
```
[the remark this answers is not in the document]
evanj80-illumio: I think we actually do not want to care or log if we drop flows since there is not anything we can really do about it.

internal/controller/streams.go
```diff
@@ -240,7 +243,7 @@ func (sm *streamManager) buildResourceApiGroupMap(resources []string, clientset
 }
 
 // StreamResources handles the resource stream.
-func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger, cancel context.CancelFunc) error {
+func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger, cancel context.CancelFunc, snapshotCommitSignal chan struct{}) error {
```
rlenglet, line 246: How about passing a `func()` instead of directly the `chan`, to make this a bit more abstract?
Like the `CancelFunc` of a cancelable `Context`...
```suggestion
func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger, cancel context.CancelFunc, signalSnapshotCommit func()) error {

File: internal/controller/streams.go [hunk 3]
```diff
@@ -328,6 +331,9 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
 	}
 	logger.Info("Successfully sent resource snapshot")
 
+	// Signal that the snapshot commit has been sent
+	close(snapshotCommitSignal)
```
rlenglet, line 335:
```suggestion
	signalSnapshotCommit()
```

internal/controller/streams.go
```diff
@@ -426,24 +432,57 @@ func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger
 }
 
 // startFlowCacheOutReader starts a goroutine that reads flows from the flow cache and sends them to the cloud secure.
-func (sm *streamManager) startFlowCacheOutReader(ctx context.Context, logger *zap.Logger) error {
+func (sm *streamManager) startFlowCacheOutReader(ctx context.Context, logger *zap.Logger, networkFlowStreamOpen chan struct{}) error {
```
rlenglet, line 435:
```suggestion
func (sm *streamManager) startFlowCacheOutReader(ctx context.Context, logger *zap.Logger, networkFlowStreamOpen chan<- struct{}) error {
```

internal/controller/streams.go
```diff
@@ -426,24 +432,57 @@ func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger
...
 	flowCount := 0
-	ticker := time.NewTicker(60 * time.Second) // Create a ticker that triggers every 60 seconds
-	defer ticker.Stop()                        // Ensure the ticker is stopped when the function exits
+	allowedFlows := 200                           // Initialize allowed flows to some default value IE 200, This value is set by the server in the ServerWindow message
```
rlenglet: 1. I'd pick a more descriptive name, consistent with the new message.
2. Shouldn't this be initialized to zero?! Only the server should be setting the window size. Like in TCP.
```suggestion
	serverWindow := 0 // Number of flow messages that can be sent without blocking until the server increases this window with a ServerWindow message.
```

internal/controller/streams.go
```diff
@@ -426,24 +432,57 @@ func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger
...
 			flowCount++
+			once.Do(func() {
+				networkFlowStreamOpen <- struct{}{}
+				logger.Info("Network flow stream opened")
+			})
 		case <-ticker.C: // Triggered every 60 seconds
-			logger.Debug("Reading from flow cache and sending flows... current flow count", zap.Int("flow_count", flowCount))
-			logger.Debug("Resetting flow count")
-			flowCount = 0
+			logger.Debug("Reading from flow cache and sending flows... current flow count + granted server window", zap.Int("flow_count", flowCount), zap.Int("allowed_flows", allowedFlows))
```
rlenglet, line 485: Add a TODO comment to remove this log and the ticker once we verify that this new logic is working fine.

internal/controller/streams.go
```diff
@@ -426,24 +432,57 @@ func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger
...
 	flowCount := 0
-	ticker := time.NewTicker(60 * time.Second) // Create a ticker that triggers every 60 seconds
-	defer ticker.Stop()                        // Ensure the ticker is stopped when the function exits
+	allowedFlows := 200                           // Initialize allowed flows to some default value IE 200, This value is set by the server in the ServerWindow message
+	logger.Info("Starting flow cache out reader") // Ensure the ticker is stopped when the function exits
+	ticker := time.NewTicker(60 * time.Second)    // Create a ticker that triggers every 60 seconds
+	defer ticker.Stop()                           // Ensure the ticker is stopped when the function exits
+	// Goroutine to read ServerWindow messages from the network flow stream
+	go func() {
+		<-networkFlowStreamOpen // Wait for network flow stream to be open
```
rlenglet, line 443: Put this in a `select` and read from the `ctx.Done()` in the same select, so we don't block forever in case the other function returns an error and never closes the channel.
```suggestion
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-networkFlowStreamOpen: // Wait for network flow stream to be ready
		}
```

internal/controller/streams.go
rlenglet, line 335: Make sure that the channel is always closed, even if this function returns prematurely above, e.g. in case or error.
Otherwise, this code currently leaks channels.

internal/controller/streams.go
```diff
@@ -426,24 +432,57 @@ func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger
...
+			}
+			if err != nil {
+				logger.Error("Error receiving response from network flows stream", zap.Error(err))
+				return
+			}
+
+			if serverWindow, ok := resp.Response.(*pb.SendKubernetesNetworkFlowsResponse_ServerWindow); ok {
+				allowedFlows = int(serverWindow.ServerWindow.AllowedMessages)
+				logger.Debug("Received ServerWindow message", zap.Int("allowed_flows", allowedFlows))
+				flowCount = 0 // Reset flowCount when a new ServerWindow message is received
```
rlenglet, line 458: Both `allowedFlows` and `flowCount` are read in another goroutine, so all reads and writes to both variables must be protected by a mutex.

internal/controller/streams.go
```diff
@@ -426,24 +432,57 @@ func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger
...
+			}
+		}
+	}()
+
+	once := &sync.Once{} // Ensure networkFlowStreamOpen is triggered only once
+
 	for {
```
rlenglet, line 465: You must not start sending flows until the window is >0, i.e., until you've received a message from the server to increase the window.
Each iteration of this loop **_must_** have a select that somehow blocks on either `<-ctx.Done()` or `allowedFlows > 0`. Not sure how you'll do the latter.

rlenglet (changes requested):

internal/controller/streams.go
evanj80-illumio: Sure that is fine with me!

evanj80-illumio: I tried this and it did not make sense since chans make more sense for signaling.

PR #252, opened 2025-07-23
Application level keepalives in single goroutine and flow cache reader recovery

evanj80-illumio:
…eader recovery

### DIFF
--- a/internal/controller/streams.go
+++ b/internal/controller/streams.go
@@ -240,7 +240,7 @@ func (sm *streamManager) buildResourceApiGroupMap(resources []string, clientset 
 }
 
 // StreamResources handles the resource stream.
-func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger, cancel context.CancelFunc) error {
+func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger, cancel context.CancelFunc, keepalivePeriod time.Duration) error {
 	defer cancel()
 	defer func() {
 		dd.processingResources = false
@@ -341,12 +341,24 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
 	dd.processingResources = false
 	dd.mutex.Unlock()
 
-	<-ctx.Done()
-	return ctx.Err()
+	ticker := time.NewTicker(jitterTime(keepalivePeriod, 0.10))
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			err := sm.sendKeepalive(logger, STREAM_RESOURCES)
+			if err != nil {
+				return err
+			}
+		}
+	}
 }
 
 // StreamLogs handles the log stream.
-func (sm *streamManager) StreamLogs(ctx context.Context, logger *zap.Logger) error {
+func (sm *streamManager) StreamLogs(ctx context.Context, logger *zap.Logger, keepalivePeriod time.Duration) error {
 	errCh := make(chan error, 1)
 	defer close(errCh)
 
@@ -366,19 +378,29 @@ func (sm *streamManager) StreamLogs(ctx context.Context, logger *zap.Logger) err
 		}
 	}()
 
-	select {
-	case <-ctx.Done():
-		return ctx.Err()
-	case err := <-errCh:
-		if err != nil {
-			return err
+	ticker := time.NewTicker(jitterTime(keepalivePeriod, 0.10))
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case err := <-errCh:
+			if err != nil {
+				return err
+			}
+			return nil
+		case <-ticker.C:
+			err := sm.sendKeepalive(logger, STREAM_LOGS)
+			if err != nil {
+				return err
+			}
 		}
 	}
-	return nil
 }
 
 // StreamConfigurationUpdates streams configuration updates and applies them dynamically.
-func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger *zap.Logger) error {
+func (sm *streamManager) StreamConfigurationUpdates(ctx context.Context, logger *zap.Logger, keepalivePeriod time.Duration) error {
 	errCh := make(chan error, 1)
 	defer close(errCh)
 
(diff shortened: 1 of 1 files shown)


internal/controller/config_stream_test.go
evanj80-illumio, line 1: This change is just accounting for new keep alive param

internal/controller/streams.go
```diff
@@ -328,25 +328,49 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
 	}
 	logger.Info("Successfully sent resource snapshot")
 
+	mutationChan := make(chan *pb.KubernetesResourceMutation)
```
rlenglet, line 331: This channel is leaked?
```suggestion
	mutationChan := make(chan *pb.KubernetesResourceMutation)

	defer close(mutationChan)
```

Hmmm this is tricky since you must wait until all the writers are done, and can't write into the channel anymore, before closing the channel.
What would be a good place to close it?
You probably need to use a `WaitGroup` to wait for all the `WatchK8sResources` goroutines to return?

internal/controller/streams.go
```diff
@@ -328,25 +328,49 @@ func (sm *streamManager) StreamResources(ctx context.Context, logger *zap.Logger
...
 		}(info, resourceManager)
 	}
 
 	dd.mutex.Lock()
 	dd.processingResources = false
 	dd.mutex.Unlock()
 
-	<-ctx.Done()
-	return ctx.Err()
+	ticker := time.NewTicker(jitterTime(keepalivePeriod, 0.10))
```
rlenglet, line 346: I still think that using a Timer would be more appropriate as it fires only once, whereas a Ticker writes multiple ticker timestamps into the channel in case we missed reading one, e.g., in case a Send takes too long.
And all those ticker timestamps get sent as multiple messages all at once, instead of just one.

Not a blocker, though.

internal/controller/grpc_logger.go
```diff
@@ -116,17 +119,36 @@ func (b *BufferedGrpcWriteSyncer) flush() {
...
+	keepAliveTicker := time.NewTicker(jitterTime(keepAlivePeriod, 0.10))
+	defer keepAliveTicker.Stop()
 	for {
 		select {
 		case <-ticker.C:
 			b.mutex.Lock()
 			b.flush()
 			b.mutex.Unlock()
+		case <-keepAliveTicker.C:
+			b.mutex.Lock()
```
rlenglet, line 144: Locking here is a good call.
In the gRPC logger, any goroutine may write a log message, which may result in a Send call to the stream.
So all usages of the stream must be protected by the mutex.

internal/controller/streams.go
```diff
@@ -847,12 +810,13 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
 		failureReason := ""
 		attempt++
 		logger.Debug("Trying to authenticate and open streams", zap.Int("attempt", attempt))
-
+		logger.Info("ABOUT TO ENTER SELECT IN FOR LOOP")
```
rlenglet, line 813:
```suggestion
```
And delete your other debugging log statements below.

internal/controller/streams.go
```diff
@@ -968,13 +942,18 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
...
+			case <-flowCacheRunDone:
+				failureReason = "Flow cache running process failed"
 			}
 			authConContextCancel()
 		}
 		logger.Warn("One or more streams have been closed; closing and reopening the connection to CloudSecure",
 			zap.String("failureReason", failureReason),
 			zap.Int("attempt", attempt),
 		)
+		resetTimer.Reset(jitterTime(5*time.Second, 0.20))
```
rlenglet, line 956: Wow, this scares me.
Why is this necessary?
It seems like it is necessary, but how come we've missed that until now???

In case the connection fails, we already reset this time to 10s-jitter here:
https://github.com/illumio/cloud-operator/pull/252/files#diff-65d4066e2022c168a97a2083474d348fa6d5d1fffb6e8d7378282935ad187940R829
Why reset it afterwards to 5s?

I would rather move this after the `if err != nil` block here, maybe in an `else` clause to make it more clear:
https://github.com/illumio/cloud-operator/pull/252/files#diff-65d4066e2022c168a97a2083474d348fa6d5d1fffb6e8d7378282935ad187940R833
So that both `Reset` calls are close to each other.

internal/controller/resource_manager.go
```diff
@@ -63,9 +65,9 @@ func NewResourceManager(config ResourceManagerConfig) *ResourceManager {
 
 // WatchK8sResources initiates a watch stream for the specified Kubernetes resource starting from the given resourceVersion.
 // This function blocks until the watch ends or the context is canceled.
-func (r *ResourceManager) WatchK8sResources(ctx context.Context, cancel context.CancelFunc, apiGroup string, resourceVersion string) {
+func (r *ResourceManager) WatchK8sResources(ctx context.Context, cancel context.CancelFunc, apiGroup string, resourceVersion string, mutationChan chan *pb.KubernetesResourceMutation, watcherWaitGroup *sync.WaitGroup) {
 	defer cancel()
-
+	defer watcherWaitGroup.Done()
```
rlenglet, line 70: Revert adding the `WaitGroup` as a parameter. Do that logic in the calling goroutine function.
```suggestion
func (r *ResourceManager) WatchK8sResources(ctx context.Context, cancel context.CancelFunc, apiGroup string, resourceVersion string, mutationChan chan *pb.KubernetesResourceMutation) {
	defer cancel()
```

rlenglet (changes requested):
Looking good. A few more comments, esp. about the new timer `Reset` call, which seems to be a good catch, but I have questions about it.

internal/controller/streams.go
```diff
@@ -515,43 +557,18 @@ func (sm *streamManager) StreamFalcoNetworkFlows(ctx context.Context, logger *za
 
 // StreamOVNKNetworkFlows handles the OVN-K network flow stream.
 func (sm *streamManager) StreamOVNKNetworkFlows(ctx context.Context, logger *zap.Logger) error {
+	errCh := make(chan error, 1)
```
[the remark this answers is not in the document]
evanj80-illumio: Yeah this was a left over from the old go routines

rlenglet (approved):

PR #254, opened 2025-07-29
Fix possible deadlocking in grpc logger

evanj80-illumio:


internal/controller/grpc_logger.go
```diff
@@ -143,10 +143,7 @@ func (b *BufferedGrpcWriteSyncer) run() {
 		case <-keepAliveTicker.C:
 			b.mutex.Lock()
 			if b.client != nil {
-				err := b.sendLogsKeepalive()
-				if err != nil {
-					b.logger.Error("Failed to send logs keepalive", zap.Error(err))
-				}
+				_ = b.sendLogsKeepalive()
 			}
 			b.mutex.Unlock()
```
rlenglet, line 148: We can keep the log, if we unlock before logging.
```suggestion
			b.mutex.Lock()
			var err error
			if b.client != nil {
				err = b.sendLogsKeepalive()
			}
			b.mutex.Unlock()
			if err != nil {
				b.logger.Error("Failed to send logs keepalive", zap.Error(err))
			}
```

rlenglet (changes requested):

PR #258, opened 2025-08-11
 added the gkenamespce and tls

pavankumarinnamuri:


internal/controller/cilium_flow_collector.go
pavankumarinnamuri, line 1: ciliumHubbleRelayNamespace string = "kube-system"
).      this can be removed , its not used anywhere

internal/controller/hubble/hubble.go
```diff
@@ -104,23 +104,39 @@ func loadMTLSConfigFromData(logger *zap.Logger, caCertData, clientCertData, clie
...
+			lastErr = fmt.Errorf("failed to fetch mTLS certificates from namespace %s: %w", namespace, fetchErr)
+			lastErr1 = errors.Join(lastErr1, lastErr)
+		}
+		logger.Debug("Could not load TLS certificates from namespace, trying next",
+			zap.String("namespace", namespace),
+			zap.Error(fetchErr))
 	}
 
-	return tlsConfig, nil
+	return nil, lastErr1
```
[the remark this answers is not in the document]
pavankumarinnamuri: will chanage

internal/controller/hubble/hubble.go
```diff
@@ -115,9 +121,9 @@ func GetTLSConfig(ctx context.Context, clientset kubernetes.Interface,
 	}
 
 	// Load config
-	tlsConfig, loadErr := loadMTLSConfigFromData(logger, caData, clientCertData, clientKeyData)
-	if loadErr != nil {
-		return nil, fmt.Errorf("failed to load mTLS credentials: %w", loadErr)
+	tlsConfig, aggregatedErrors := loadMTLSConfigFromData(logger, caData, clientCertData, clientKeyData, namespace)
```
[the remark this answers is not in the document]
aaronIllumio: just keep this the same or change it to err

internal/controller/streams.go
```diff
@@ -492,7 +496,13 @@ func (sm *streamManager) startFlowCacheOutReader(ctx context.Context, logger *za
 
 // findHubbleRelay returns a *CiliumFlowCollector if hubble relay is found in the given namespace
 func (sm *streamManager) findHubbleRelay(ctx context.Context, logger *zap.Logger, ciliumNamespace string) *CiliumFlowCollector {
-	ciliumFlowCollector, err := newCiliumFlowCollector(ctx, logger, ciliumNamespace, sm.streamClient.tlsAuthProperties)
+	namespaces := []string{ciliumNamespace}
```
aaronIllumio, line 499:
```suggestion
	namespaces := sm.streamClient.ciliumNamespaces
```

aaronIllumio (dismissed):
Everything looks good. you should just set all of the namespaces to one list of namespaces and use that no need to pass around namespace variables.

cloud-operator/templates/role.yaml
```diff
@@ -21,4 +21,15 @@ rules:
...
+# Role to read secrets in the GKE managed observability namespace
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: illumio-cloud-operator-secret-reader
+  namespace: gke-managed-dpv2-observability
+rules:
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get", "list"]
```
rlenglet, line 35: Why would "list" be required for namespace `gke-managed-dpv2-observability` and not for `kube-system`, even though the logic to read the secret is identical for either namespace?
```suggestion
  verbs: ["get"]
```

cloud-operator/templates/role.yaml
```diff
@@ -21,4 +21,15 @@ rules:
...
+---
+# Role to read secrets in the GKE managed observability namespace
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: illumio-cloud-operator-secret-reader
+  namespace: gke-managed-dpv2-observability
+rules:
+- apiGroups: [""]
+  resources: ["secrets"]
```
rlenglet, line 34: The secret name is hardcoded ([here](https://github.com/illumio/cloud-operator/pull/258/files#diff-6d4dd1d6018369bf594f74c982493806edfd65dbc3a000f9157924d645baddbcR28)), so you should limit the role to access only this one resource, like in the Role for namespace `kube-system` above:
```suggestion
  resources: ["secrets"]
  resourceNames: ["hubble-relay-client-certs"] # Name of the Secret
```
Both Roles should be completely identical except for the namespace.

cloud-operator/templates/role.yaml
```diff
@@ -21,4 +21,15 @@ rules:
...
   resourceNames: ["hubble-relay-client-certs"] # Name of the Secret
-  verbs: ["get"]
\ No newline at end of file
+  verbs: ["get"]
+---
+# Role to read secrets in the GKE managed observability namespace
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: illumio-cloud-operator-secret-reader
```
rlenglet, line 30: Use the same name as the Role we already create in namespace `kube-system`, for consistency.
```suggestion
  name: hubble-client-certs-reader
```

cloud-operator/templates/rolebinding.yaml
```diff
@@ -26,4 +26,18 @@ subjects:
...
   kind: Role
   name: hubble-client-certs-reader
-  apiGroup: rbac.authorization.k8s.io
\ No newline at end of file
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: illumio-cloud-operator-secret-reader-binding
```
rlenglet, line 34: Use the same name as the RoleBinding we create in `kube-system` above, for consistency.
```suggestion
  name: cloud-operator-read-hubble-client-certs
```

cloud-operator/templates/rolebinding.yaml
```diff
@@ -26,4 +26,18 @@ subjects:
...
+  namespace: gke-managed-dpv2-observability
+subjects:
+- kind: ServiceAccount
+  name: {{ .Values.serviceAccount.name }}
+  namespace: {{ .Values.defaultNamespace | default .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: illumio-cloud-operator-secret-reader
```
rlenglet, line 43:
```suggestion
  name: hubble-client-certs-reader
```

cloud-operator/templates/rolebinding.yaml
```diff
@@ -26,4 +26,18 @@ subjects:
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: illumio-cloud-operator-secret-reader-binding
+  namespace: gke-managed-dpv2-observability
+subjects:
+- kind: ServiceAccount
+  name: {{ .Values.serviceAccount.name }}
+  namespace: {{ .Values.defaultNamespace | default .Release.Namespace }}
```
rlenglet, line 39: Why would we ever refer to the namespace?
```suggestion
  namespace: {{ .Release.Namespace }}
```
This RoleBinding should be identical to the RoleBinding created in `kube-system` above, except for the namespace.

cloud-operator/templates/role.yaml
```diff
@@ -21,4 +21,15 @@ rules:
 - apiGroups: [""]
   resources: ["secrets"]
   resourceNames: ["hubble-relay-client-certs"] # Name of the Secret
-  verbs: ["get"]
\ No newline at end of file
+  verbs: ["get"]
+---
+# Role to read secrets in the GKE managed observability namespace
+apiVersion: rbac.authorization.k8s.io/v1
```
rlenglet, line 27: Have you validated that this Helm chart now works in non-GKE environments?
I suspect that trying to create a resource in namespace `gke-managed-dpv2-observability` will fail if that namespace doesn't exist, which is the case in any non-GKE cluster.
The creation of this resource must be conditional on the existence of the `gke-managed-dpv2-observability` namespace.
Wrap this with:
```
{{- if (lookup "v1" "Namespace" "gke-managed-dpv2-observability" "") }}
...
{{- end }}
```

cloud-operator/templates/rolebinding.yaml
```diff
@@ -26,4 +26,18 @@ subjects:
 roleRef:
   kind: Role
   name: hubble-client-certs-reader
-  apiGroup: rbac.authorization.k8s.io
\ No newline at end of file
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
rlenglet, line 31: Only create this resource if the `gke-managed-dpv2-observability` namespace exists, i.e., if the chart is deployed into a GKE cluster:
```
{{- if (lookup "v1" "Namespace" "gke-managed-dpv2-observability" "") }}
...
{{- end }}
```

internal/controller/cilium_flow_collector.go
```diff
@@ -28,32 +29,63 @@ const (
...
 
-	service, err := hubble.DiscoverCiliumHubbleRelay(ctx, ciliumNamespace, clientset)
-	if err != nil {
-		return nil, fmt.Errorf("failed to discover Cilium Hubble Relay service: %w", err)
-	}
+	for _, ciliumNamespace := range ciliumNamespaces {
+		// Step 1: Try to discover Hubble Relay
+		service, err := hubble.DiscoverCiliumHubbleRelay(ctx, ciliumNamespace, clientset)
+		if err != nil {
+			logger.Warn("Failed to discover Cilium Hubble Relay service",
```
rlenglet, line 47: On GKE, this would just be noise at Warn level. Lower to Debug.
```suggestion
			logger.Debug("Failed to discover Cilium Hubble Relay service",
```

internal/controller/hubble/hubble.go
```diff
@@ -21,10 +21,16 @@ import (
 )
 
 const (
-	ciliumHubbleRelayExpectedServerName = "ui.hubble-relay.cilium.io"
-	ciliumHubbleRelayServiceName        = "hubble-relay"
+	ciliumHubbleRelayExpectedServerName     = "ui.hubble-relay.cilium.io"
+	gkeManagedHubbleRelayExpectedServerName = "hubble-relay.gke-managed-dpv2-observability.svc.cluster.local"
```
rlenglet, line 25: It doesn't make sense to hardcode this, since you made the package name configurable, and the Service is somewhat discovered dynamically.
This service FQDN should be constructed as `<service name>.<package name>.svc.cluster.local`.
Cf. https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-aaaa-records

That is normal. The `ui.hubble-relay.cilium.io` domain name is an anomaly that we have to deal with.

rlenglet (changes requested):

cmd/main.go
```diff
@@ -105,7 +106,7 @@ func main() {
 	bindEnv(logger, "verbose_debugging", "VERBOSE_DEBUGGING")
 	// Set default values
 	viper.SetDefault("cluster_creds", "clustercreds")
-	viper.SetDefault("cilium_namespace", "kube-system")
+	viper.SetDefault("cilium_namespaces", "kube-system,gke-managed-dpv2-observability")
```
rlenglet, line 109: Viper natively supports string slices.
`BindEnv` is supposed to parse env variable values as comma-separated in this case, but that needs to be tested.
Can you please test this?
```suggestion
	viper.SetDefault("cilium_namespaces", []string{"kube-system", "gke-managed-dpv2-observability"})
```

Make sure to test by setting `CILIUM_NAMESPACES` to e.g. `"foo,bar"` and see that it gets parsed into `[]string{"foo", "bar"}`.

cmd/main.go
```diff
@@ -124,7 +125,7 @@ func main() {
 
 	envConfig := controller.EnvironmentConfig{
 		ClusterCreds:           viper.GetString("cluster_creds"),
-		CiliumNamespace:        viper.GetString("cilium_namespace"),
+		CiliumNamespaces:       strings.Split(viper.GetString("cilium_namespaces"), ","),
```
rlenglet, line 128:
```suggestion
		CiliumNamespaces:       viper.GetStringSlice("cilium_namespaces"),
```

internal/controller/cilium_flow_collector.go
```diff
@@ -4,6 +4,7 @@ package controller
 
 import (
 	"context"
+	cryptoTls "crypto/tls"
```
rlenglet, line 7: No capital letters in package names/aliases.
```suggestion
	cryptotls "crypto/tls"
```

rlenglet (changes requested):
Please try Viper's native support for string slices.

PR #260, opened 2025-08-13
Enable more golangci-lint linters and formatters and fix errors

rlenglet:
Bump Buf to v1.56.0 in devcontainer.
Update Buf config format to v2.
Update Buf grpc and go plugins.
Re-generate the gRPC code.

Bump Helm to v3.18.5 in devcontainer and GitHub Actions.
Bump Kind to v0.29.0 in devcontainer.
Install GCI in devcontainer for import formatting.

Bump golangci-lint to v2.3.1 in devcontainer and GitHub Actions.
Enable most golangci-lint plugins and fix linter errors.
Reformat all imports with `gci write */* -s standard -s default -s 'prefix(github.com/illumio/cloud-operator)'`
Refactor function `NewAuthenticatedConnection` and file `authenticator.go` to reduce the code complexity.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -45,9 +45,7 @@ formatters:
   enable:
     - gci
     - gofmt
-    # - gofumpt
     - goimports
-    # - golines
   settings:
     gci:
       custom-order: true


.github/workflows/go-security-checks.yml
```diff
@@ -33,9 +33,9 @@ jobs:
         run: staticcheck ./...
 
       - name: Install golangci-lint
-        uses: golangci/golangci-lint-action@v6
+        uses: golangci/golangci-lint-action@v8
         with:
-          version: v1.64.2
+          version: v2.3.1
```
evanj80-illumio, line 38: We were this far behind on the versioning?!

.golangci.yml
```diff
@@ -0,0 +1,57 @@
...
+    paths:
+      # Files in package internal/pkg/tls were copied from the gRPC package.
+      # Ignore issues in those files.
+      - "internal/pkg/tls/.*\\.go"
+
+formatters:
+  enable:
+    - gci
+    - gofmt
+    # - gofumpt
```
evanj80-illumio, line 48: Delete these rather than commenting them out?

.golangci.yml
```diff
@@ -0,0 +1,57 @@
...
+    - goimports
+    # - golines
```
evanj80-illumio, line 50: Same here, delete instead of commenting out?

internal/controller/authenticator.go
evanj80-illumio, line 1: Just trying to simplify logging situation?

internal/controller/streams.go
```diff
@@ -958,58 +1053,15 @@ func ConnectStreams(ctx context.Context, logger *zap.Logger, envMap EnvironmentC
...
-		err = authn.WriteK8sSecret(ctx, responseData, envMap.ClusterCreds, envMap.PodNamespace)
-		if err != nil {
-			logger.Error("Failed to write secret to Kubernetes", zap.Error(err))
-		}
-
-		// k8s may take some time writing the secret. Here we will try 'maxRetries'
-		// times, waiting 'waitDuration' seconds between each try. Even a single 1
-		// second wait is probably fine, but just to be semantic this wait is done
-		// as a poll.
-		maxRetries := 5
```
evanj80-illumio, line 992: Im not sure why we are deleting all of this?

internal/controller/authenticator.go
rlenglet: Reducing the complexity, really.

internal/controller/streams.go
rlenglet: This function was too long and complex, which caused an linter error. I cut out most of this logic into a method of Authenticator, in authenticator.go, because it was coupled to several methods of that type. Then I realized that Authenticator was just a wrapper around a Logger, which was not useful and added unnecessary complexity, so I got rid of that type and made all its methods simple functions. Then I realized that the Logger was not really useful either, and replaced it with returning error wrappers.
Then I simplified further by removing the logic that retried reading from the Secret after writing into the Secret. I didn’t see the point of doing that.
Overall, this cut quite a bit of code in both streams.go and authenticator.go.

evanj80-illumio: Ah okay makes sense!

.golangci.yml
rlenglet: Done.

Evanj80 (approved):

PR #266, opened 2025-08-19
Generate the CycloneDX SBOM for every release

rlenglet:
Install Syft and CycloneDX's sbom-utility in devcontainer.
Update the release GitHub Action to generate and upload the SBOM.

Temporarily define the new workflow to only trigger on releases with tags named `test*`.
Will make it the main workflow on `v*` tags later, after this has been merged and tested.


evanj80-illumio (approved):
Send it in and give this a try!

PR #371, opened 2026-03-24
Ingest Cilium CRDs (CNP, CCNP, CIDRGroup) into resource stream

pavankumarinnamuri:
## Summary

Adds support for ingesting Cilium CRDs — CiliumNetworkPolicy, CiliumClusterwideNetworkPolicy, 
and CiliumCIDRGroup — into the resource stream sent to CloudSecure.

### What changed

- **Proto schema**: Added Cilium-specific message types to `KubernetesObjectData` 
  (policy rules, ingress/egress, CIDR sets, FQDN selectors, ICMP, services, AWS groups, 
  authentication, default deny)
- **Typed deserialization** (`cilium.go`): Converts unstructured Cilium CRDs into typed 
  Cilium structs via `CodecFactory`/`UniversalDeserializer`, then maps them to proto. 
  Handles both `v2` and `v2alpha1` CIDRGroup GVKs.
- **Converter injection** (`watcher.go`): Introduced `ResourceConverter` function type 
  so each Watcher uses the right converter (core vs Cilium) without the Watcher knowing 
  about either. Removed `ListResources`/`ExtractObjectMetas` in favor of the converter 
  operating directly on unstructured objects.
- **Resilience**: Converter errors during list operations log-and-skip rather than 
  failing the entire stream. `FetchResources` handles NotFound (deleted CRDs) alongside 
  Forbidden.
- **RBAC**: ClusterRole updated to grant get/list/watch on `cilium.io` resources.


api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,81 @@ message KubernetesNetworkPolicyData {
...
+  // ToEndpoints selects destination endpoints by labels (for egress rules).
+  repeated LabelSelector to_endpoints = 2;
+  // FromCIDR allows traffic from specific CIDR ranges.
+  repeated string from_cidr = 3;
+  // ToCIDR allows traffic to specific CIDR ranges.
+  repeated string to_cidr = 4;
+  // FromCIDRSet allows traffic from CIDR ranges with exceptions.
+  repeated CiliumCIDRSet from_cidr_set = 5;
+  // ToCIDRSet allows traffic to CIDR ranges with exceptions.
+  repeated CiliumCIDRSet to_cidr_set = 6;
```
sungkyunchung: Can we use a single field instead of having both `from_` and `to_`?  It looks like ingress rules only use from, and egress rules only use to.

internal/controller/stream/resources/watcher.go
```diff
@@ -88,7 +88,12 @@ func (r *Watcher) WatchK8sResources(ctx context.Context, cancel context.CancelFu
 
 // DynamicListResources lists a specified resource dynamically and sends down the current gRPC stream.
 func (r *Watcher) DynamicListResources(ctx context.Context, logger *zap.Logger, apiGroup string) (string, error) {
-	objGVR := schema.GroupVersionResource{Group: apiGroup, Version: "v1", Resource: r.resourceName}
+	objGVR := schema.GroupVersionResource{Group: apiGroup, Version: getVersionForGroup(apiGroup), Resource: r.resourceName}
+
+	// For Cilium policies, we need the full unstructured object to extract the spec
+	if controller.IsCiliumPolicy(removeListSuffix(r.resourceName)) {
```
rigdomin-illumio: Where are we registering Cilium resource watcher?
Based on watch object struct, it looks like a single watch instance is dedicated to resource name, no?

Also, what would `r.resourceName` be here?
The method checks for `CiliumNetworkPolicy` or `CiliumClusterwideNetworkPolicy`; but i suspect `r.resourceName` would contain something more generic like `ciliumnetworkpolicies` based of clusterrole.yml

rigdomin-illumio (dismissed):
Called out some nit picks and concerns.
Big item to discuss will be around design comment:
https://github.com/illumio/cloud-operator/pull/371/changes#r3019112187

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
pavankumarinnamuri: toPorts is actually used in both ingress and egress rule. that's why we need the separation.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,337 @@
...
+)
+
+// IsCiliumPolicy returns true if the resource kind is a Cilium network policy.
+func IsCiliumPolicy(kind string) bool {
+	return kind == "CiliumNetworkPolicy" || kind == "CiliumClusterwideNetworkPolicy"
+}
+
+// ConvertUnstructuredToCiliumPolicy converts an unstructured Cilium policy to a KubernetesObjectData proto.
+// This handles both CiliumNetworkPolicy and CiliumClusterwideNetworkPolicy.
+func ConvertUnstructuredToCiliumPolicy(obj *unstructured.Unstructured) *pb.KubernetesObjectData {
```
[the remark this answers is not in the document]
rlenglet: > Having strict parsing on operator puts a lot of weight on operator maintaining schema correctness.

How much effort do you expect this parsing to be? Those CRDs are immutable. And these resources must be parsed anyway. Why does it matter where they are parsed?
IMHO, parsing on the client side and strongly typing APIs makes interoperability much easier. There's no surprise about what the client is sending and what the server expects.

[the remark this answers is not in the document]
rlenglet: Also, there's data we don't care about, esp. the `status`.
Parsing on the client side and having a focused API allows for eliminating that unnecessary data, which can sometimes be quite big.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,81 @@ message KubernetesNetworkPolicyData {
...
+  // Uses Cilium's endpoint selector which supports label-based and identity-based selection.
+  LabelSelector endpoint_selector = 1;
+  // Ingress rules define allowed/denied incoming traffic.
+  repeated CiliumNetworkPolicyRule ingress_rules = 2;
+  // Egress rules define allowed/denied outgoing traffic.
+  repeated CiliumNetworkPolicyRule egress_rules = 3;
+  // IngressDeny rules explicitly deny incoming traffic.
+  repeated CiliumNetworkPolicyRule ingress_deny_rules = 4;
+  // EgressDeny rules explicitly deny outgoing traffic.
+  repeated CiliumNetworkPolicyRule egress_deny_rules = 5;
```
rlenglet: Clarify somewhere that the rules are aggregated from both the `spec` and `specs` fields.
https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/k8s/apis/cilium.io/v2/cnp_types.go#L39-L47
Same for `KubernetesCiliumClusterwideNetworkPolicyData` below.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,81 @@ message KubernetesNetworkPolicyData {
...
+}
+
+// KubernetesCiliumClusterwideNetworkPolicyData contains information specific to a CiliumClusterwideNetworkPolicy object.
+// CiliumClusterwideNetworkPolicy is a cluster-scoped policy that applies across all namespaces.
+// https://docs.cilium.io/en/stable/security/policy/language/#ciliumclusterwidenetworkpolicy
+message KubernetesCiliumClusterwideNetworkPolicyData {
+  // Endpoint selector selects the pods to which this policy applies cluster-wide.
+  LabelSelector endpoint_selector = 1;
+  // Node selector selects the nodes to which this policy applies (for host policies).
+  optional LabelSelector node_selector = 2;
```
rlenglet: This is not modeled correctly. `endpointSelector` and `nodeSelector` are specified per-rule, not per-policy:
https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/policy/api/rule.go#L71-L83
Move these fields into `CiliumNetworkPolicyRule`.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,81 @@ message KubernetesNetworkPolicyData {
...
+  repeated CiliumNetworkPolicyRule egress_rules = 4;
+  // IngressDeny rules explicitly deny incoming traffic.
+  repeated CiliumNetworkPolicyRule ingress_deny_rules = 5;
+  // EgressDeny rules explicitly deny outgoing traffic.
+  repeated CiliumNetworkPolicyRule egress_deny_rules = 6;
+}
+
+// CiliumNetworkPolicyRule defines a single Cilium ingress or egress rule.
+// Cilium rules support richer selectors including identity-based and CIDR-based selection.
+message CiliumNetworkPolicyRule {
```
rlenglet: Add fields that you omitted:
- `enableDefaultDeny`: https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/policy/api/rule.go#L138
- `labels`: https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/policy/api/rule.go#L119
- `description`: https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/policy/api/rule.go#L145

I don't think we need to include `log` (https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/policy/api/rule.go#L150).

rlenglet: Split this into multiple messages, one for ingress rule and another for egress rule.
https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/policy/api/rule.go#L85-L111
Mixing ingress and egress into a single message is very confusing. Not sure if it's worth splitting between allow rules and deny rules, like Cilium does. Let's see how it goes first with splitting ingress vs. egress.

rlenglet: This is missing important fields.
For example:
- `toFQDNs`: https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/policy/api/egress.go#L167
- `icmps`: https://github.com/cilium/cilium/blob/0dec54c5a7a314ad7feec23b75d3ef1768f20838/pkg/policy/api/egress.go#L176-L177
- `authentication`: https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/policy/api/egress.go#L182
- `fromGroups`: https://github.com/cilium/cilium/blob/0dec54c5a7a314ad7feec23b75d3ef1768f20838/pkg/policy/api/ingress.go#L78-L89
- `fromNodes`: https://github.com/cilium/cilium/blob/0dec54c5a7a314ad7feec23b75d3ef1768f20838/pkg/policy/api/ingress.go#L91-L96
- `toServices`: https://github.com/cilium/cilium/blob/0dec54c5a7a314ad7feec23b75d3ef1768f20838/pkg/policy/api/egress.go#L77-L82
- `toGroups`: https://github.com/cilium/cilium/blob/0dec54c5a7a314ad7feec23b75d3ef1768f20838/pkg/policy/api/egress.go#L84-L95
- `toNodes`: https://github.com/cilium/cilium/blob/0dec54c5a7a314ad7feec23b75d3ef1768f20838/pkg/policy/api/egress.go#L97-L101

PLEASE review the CRDs and make sure that all relevant fields from policy and rule types are included in this API.
https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/k8s/apis/cilium.io/v2/cnp_types.go#L32-L54
https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/k8s/apis/cilium.io/v2/ccnp_types.go#L28-L53
https://github.com/cilium/cilium/blob/0dec54c5a7a314ad7feec23b75d3ef1768f20838/pkg/policy/api/rule.go#L70-L151

rlenglet (changes requested):
I looked only at the API changes.
Please structure this API to more closely follow Cilium's CRDs. Esp. it would be easier to follow if you split ingress and egress rules into different messages.
And there were quite a few discrepancies with the Cilium CRDs and missing fields. Please do another pass to make sure you closely follow the CRDs and include all relevant fields.

internal/controller/cilium_network_policy.go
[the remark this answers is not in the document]
rlenglet: More generally also, I prefer to tightly couple our backend server with our cloud-operator, than tightly coupling our backend server with Cilium and each CNI plugin.
I'd rather have the backend server have dependencies on this one API defined in cloud-operator, rather than have it directly coupled with 4 CNI plugins for policy APIs, and 4 different flow formats.

In any case, @rigdomin-illumio, you will have to parse those policies inside cloud-operator to implement the operator logic. So we might as well send the result of that parsing to the server. cloud-operator cannot keep those resources opaque.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rigdomin-illumio: Done

rigdomin-illumio: Created separate CiliumIngressRule and CiliumEgressRule messages.

rigdomin-illumio: Done

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rigdomin-illumio: **Went for a Hierarchical Approach**

Rationale:
- Accurate 1:1 mapping to Cilium's CRD structure
- Solves both Comment 2 and Comment 3 together
- No data loss - can reconstruct original policy structure
- Spec-level fields (description, labels, enableDefaultDeny) have a proper home
- Easier to validate and debug by comparing with `kubectl get cnp -o yaml`

### Changes Made
1. Created `CiliumPolicySpec` message with selector + rules + metadata
2. Created `CiliumDefaultDeny` message for enable_default_deny field
3. Updated `KubernetesCiliumNetworkPolicyData` to use `repeated CiliumPolicySpec specs`
4. Updated `KubernetesCiliumClusterwideNetworkPolicyData` similarly
5. Updated conversion code:
   - `collectCiliumSpecs()` reads from both `spec` and `specs` fields
   - `convertCiliumPolicySpec()` converts each spec with its selector and rules
   - `convertCiliumDefaultDeny()` handles enableDefaultDeny config

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rigdomin-illumio: The hierarchical spec structure naturally handles both `spec` and `specs` by converting them to a unified `repeated CiliumPolicySpec specs` array.

internal/controller/stream/resources/watcher.go
```diff
@@ -117,6 +122,41 @@ func (r *Watcher) DynamicListResources(ctx context.Context, logger *zap.Logger,
...
+	unstructuredResources, err := r.FetchResources(ctx, objGVR, metav1.NamespaceAll)
+	if err != nil {
+		return "", err
+	}
+
+	for i := range unstructuredResources.Items {
+		item := &unstructuredResources.Items[i]
+		metadataObj := controller.ConvertUnstructuredToCiliumPolicy(item)
+
+		r.logger.Info("Sending Cilium policy",
```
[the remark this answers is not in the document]
rigdomin-illumio: Future TODO.
Context is empty at this time

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,229 @@ message KubernetesNetworkPolicyData {
...
+  }
+}
+
+// CiliumAWSGroup defines AWS-specific security group selector fields.
+message CiliumAWSGroup {
+  optional string region = 1;
+  // Security group IDs (Cilium supports multiple per group selector)
+  repeated string security_group_ids = 2;
+  optional string vpc_id = 3;
+}
```
sungkyunchung, line 248: I think we are missing Labels and SecurityGroupNames here. Other fields look good to me https://github.com/cilium/cilium/blob/3977f6a1120bf5ec0206c1b3c484ba094da82c7d/pkg/policy/api/groups.go#L34

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,229 @@ message KubernetesNetworkPolicyData {
...
+// The duplication is minimal (one repeated field each) and preserves semantic clarity.
+
+// KubernetesCiliumNetworkPolicyData contains information specific to a CiliumNetworkPolicy object.
+// CiliumNetworkPolicy is a namespaced policy that applies to pods selected by the endpoint selector.
+// https://docs.cilium.io/en/stable/security/policy/language/#ciliumnetworkpolicy
+message KubernetesCiliumNetworkPolicyData {
+  // Policy specs from both `spec` (single) and `specs` (array) fields, unified as an array.
+  // Each spec has its own endpoint selector and rules, mirroring Cilium's CRD structure.
+  repeated CiliumPolicySpec specs = 1;
+}
```
ryan-er, line 116: Ohh so this is more recent than Sungkyun's one that I reviewed. I suggest moving this out to a separate PR and get it reviewed and finalized. If it gets merged it can unblock server side i.e. k8sclustersync too cc @rlenglet

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,229 @@ message KubernetesNetworkPolicyData {
...
+
+// KubernetesCiliumClusterwideNetworkPolicyData contains information specific to a CiliumClusterwideNetworkPolicy object.
+// CiliumClusterwideNetworkPolicy is a cluster-scoped policy that applies across all namespaces.
+// https://docs.cilium.io/en/stable/security/policy/language/#ciliumclusterwidenetworkpolicy
+message KubernetesCiliumClusterwideNetworkPolicyData {
+  // Policy specs from both `spec` (single) and `specs` (array) fields, unified as an array.
+  // Each spec has its own endpoint/node selector and rules, mirroring Cilium's CRD structure.
+  repeated CiliumPolicySpec specs = 1;
```
ryan-er, line 124: I think there's status field here? https://docs.cilium.io/en/stable/network/kubernetes/policy/#ciliumclusterwidenetworkpolicy

internal/controller/stream/resources/watcher.go
```diff
@@ -117,6 +136,50 @@ func (r *Watcher) DynamicListResources(ctx context.Context, logger *zap.Logger,
...
+			zap.String("name", item.GetName()),
+			zap.String("namespace", item.GetNamespace()),
+			zap.String("kind", item.GetKind()))
+
+		err = r.streamManager.SendObjectData(logger, metadataObj)
+		if err != nil {
+			r.logger.Error("Cannot send Cilium policy metadata", zap.Error(err))
+
+			return "", err
+		}
```
ryan-er: Do we need to duplicate sending code that is already in `DynamicListResources`? Unless there's a good reason. Sending should still be exclusively done inside `DynamicListResources`. It's cleaner that way IMO.

rlenglet: +1

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: Yeah, but I think only a subset is relevant/interesting in there.
https://github.com/cilium/cilium/blob/fa61925279472dff994027c66046aaf4ab709df0/pkg/k8s/apis/cilium.io/v2/cnp_types.go#L78-L93
We can add that subset later if needed.

internal/controller/stream/resources/watcher.go
```diff
@@ -107,6 +107,12 @@ func (r *Watcher) WatchK8sResources(ctx context.Context, cancel context.CancelFu
 
 // DynamicListResources lists a specified resource dynamically and sends down the current gRPC stream.
 func (r *Watcher) DynamicListResources(ctx context.Context, logger *zap.Logger) (string, error) {
+	// For Cilium policies, we need the full unstructured object to extract the spec.
+	// Note: r.resourceName is lowercase plural (e.g., "ciliumnetworkpolicies").
+	if controller.IsCiliumPolicy(r.resourceName) {
+		return r.listCiliumResources(ctx, logger)
```
sungkyunchung: Can we modify this function to return *pb.KubernetesObjectData so both Cilium resources and core resources can be sent in the same place like we do in processMutation()?

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,756 @@
...
+	objMetadata := &pb.KubernetesObjectData{
+		Annotations:       obj.GetAnnotations(),
+		CreationTimestamp: timestamppb.New(obj.GetCreationTimestamp().Time),
+		Kind:              kind,
+		Labels:            obj.GetLabels(),
+		Name:              obj.GetName(),
+		Namespace:         proto.String(namespace),
+		OwnerReferences:   convertOwnerReferences(obj.GetOwnerReferences()),
+		ResourceVersion:   obj.GetResourceVersion(),
+		Uid:               string(obj.GetUID()),
```
sungkyunchung: Could you add ApiGroup and ApiVersion fields here like we do for core resources

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,756 @@
...
+		rule.NodeSelector = convertCiliumLabelSelector(nodeSelector)
+	}
+
+	// Extract description
+	if description, found, _ := unstructured.NestedString(spec, "description"); found {
+		rule.Description = proto.String(description)
+	}
+
+	// Extract labels
+	if labels, found, _ := unstructured.NestedStringMap(spec, "labels"); found {
```
sungkyunchung: https://github.com/cilium/cilium/blob/40fafc202f3c16dfa287af9eb4dc3f3e72a120f3/pkg/policy/api/rule.go#L119
https://github.com/cilium/cilium/blob/4d49b59a9e62bdf8c5c3b7c51ac85bec0cb02789/pkg/labels/array.go#L14
https://github.com/cilium/cilium/blob/4d49b59a9e62bdf8c5c3b7c51ac85bec0cb02789/pkg/labels/labels.go#L189
Cilium's labels is labels.LabelArray ([]Label — array of {key, value} structs), not map[string]string.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,756 @@
...
+				if v >= 0 && v <= 255 {
+					protoField.Type = &pb.CiliumPolicyICMPField_TypeInt{TypeInt: uint32(v)}
+				}
+			case string:
+				if intVal, err := strconv.ParseUint(v, 10, 32); err == nil && intVal <= 255 {
+					protoField.Type = &pb.CiliumPolicyICMPField_TypeInt{TypeInt: uint32(intVal)}
+				}
+			}
+		}
+
```
sungkyunchung: https://github.com/cilium/cilium/blob/40fafc202f3c16dfa287af9eb4dc3f3e72a120f3/pkg/policy/api/icmp.go#L80-L109
ICMP type can be a string, CamelCase name (for example, "EchoReply").

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,756 @@
...
+	kind := obj.GetKind()
+	namespace := obj.GetNamespace()
+
+	objMetadata := &pb.KubernetesObjectData{
+		Annotations:       obj.GetAnnotations(),
+		CreationTimestamp: timestamppb.New(obj.GetCreationTimestamp().Time),
+		Kind:              kind,
+		Labels:            obj.GetLabels(),
+		Name:              obj.GetName(),
+		Namespace:         proto.String(namespace),
```
sungkyunchung: https://github.com/illumio/cloud-operator/blob/4cbbadec02913b6d05f525b711051bc1d9c6ef7c/internal/controller/k8s_resources.go#L96

For k8s core resources, we drop this field if it's absent. Could you fix one of them for consistency.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,756 @@
...
+		selectorMap, ok := selector.(map[string]any)
+		if !ok {
+			continue
+		}
+
+		result = append(result, convertCiliumLabelSelector(selectorMap))
+	}
+
+	return result
+}
```
sungkyunchung: This function seems identical to convertCiliumEndpointSelectors()

sungkyunchung (changes requested):

sungkyunchung:
FYI, CiliumCIDRGroup ingestion is also on the backlog.

internal/controller/stream/resources/watcher.go
```diff
@@ -134,6 +139,46 @@ func (r *Watcher) DynamicListResources(ctx context.Context, logger *zap.Logger)
...
+		}
+	}
+
+	r.logger.Debug("Successfully sent Cilium policies", zap.Int("count", len(unstructuredResources.Items)))
+
+	select {
+	case <-ctx.Done():
+		return "", ctx.Err()
+	default:
+	}
```
ryan-er, line 156: This part maybe don't need to be duplicated? Once the cilium object  is converted to `*pb.KubernetesObjectData` it should be the same logic afterwards.

internal/controller/stream/resources/watcher.go
```diff
@@ -107,68 +107,49 @@ func (r *Watcher) WatchK8sResources(ctx context.Context, cancel context.CancelFu
 
 // DynamicListResources lists a specified resource dynamically and sends down the current gRPC stream.
 func (r *Watcher) DynamicListResources(ctx context.Context, logger *zap.Logger) (string, error) {
-	// For Cilium policies, we need the full unstructured object to extract the spec.
-	// Note: r.resourceName is lowercase plural (e.g., "ciliumnetworkpolicies").
-	if controller.IsCiliumPolicy(r.resourceName) {
-		return r.listCiliumResources(ctx, logger)
-	}
-
-	objs, resourceListVersion, resourceK8sKind, apiGroup, apiVersion, err := r.ListResources(ctx, metav1.NamespaceAll)
```
sungkyunchung, line 110: Nice refactoring, Can we remove unused functions like ListResources? looks like it's no longer used.

internal/controller/cilium_network_policy.go
sungkyunchung: can we add a default case here just to be safer?

sungkyunchung (commented):
LGTM other than a few minor nits

internal/controller/stream/resources/watcher.go
```diff
@@ -107,31 +107,57 @@ func (r *Watcher) WatchK8sResources(ctx context.Context, cancel context.CancelFu
...
 func (r *Watcher) DynamicListResources(ctx context.Context, logger *zap.Logger) (string, error) {
-	objs, resourceListVersion, resourceK8sKind, apiGroup, apiVersion, err := r.ListResources(ctx, metav1.NamespaceAll)
+	unstructuredResources, err := r.FetchResources(ctx, metav1.NamespaceAll)
 	if err != nil {
 		return "", err
 	}
 
-	for _, obj := range objs {
-		metadataObj := controller.ConvertMetaObjectToMetadata(ctx, obj, r.clientset, resourceK8sKind, apiGroup, apiVersion)
+	isCilium := controller.IsCiliumPolicy(r.resourceName)
```
ryan-er: It would be nice if `DynamicListResources` doesn't have special case for any resource type like it does here all of a sudden for cilium policy. It used to not have that.

At the time of `Watcher` creation we should already know whether this is cilium policy resource or other regular existing k8s resources, right? Can we have the watcher be initialized with a converter function (basically a `map` function) that knows how to convert `unstructured.Unstructured` to `metav1.ObjectMeta`? That way when we watch a new resource type we just need to write a new converterfunction and don't need to touch `DynamicListResources` at all. The mapping of resource type -> converter function can be initialized statically.

I feel like @rlenglet will have opinion/suggestions on this, if he has time to look at it.

internal/controller/stream/resources/watcher.go
```diff
@@ -368,13 +401,36 @@ func getResourceVersionFromBookmark(event watch.Event) (string, error) {
...
 	}
 
+	// Note: resourceGvk.Kind is PascalCase (e.g., "CiliumNetworkPolicy").
 	resourceGvk := event.Object.GetObjectKind().GroupVersionKind()
-	metadataObj := controller.ConvertMetaObjectToMetadata(ctx, *convertedData, r.clientset, resourceGvk.Kind, resourceGvk.Group, resourceGvk.Version)
+
+	var metadataObj *pb.KubernetesObjectData
+
+	// Handle Cilium policies specially to extract full spec.
+	if controller.IsCiliumPolicy(resourceGvk.Kind) {
```
ryan-er: This could also be made to not special case cilium policy, I think. Like my comment here: https://github.com/illumio/cloud-operator/pull/371/changes#r3211700893.

ryan-er (commented):
I'm done with my review for sure now. Looks good but could be nicer/cleaner.

ryan-er (dismissed):

sungkyunchung (dismissed):

cloud-operator/templates/clusterrole.yaml
```diff
@@ -19,6 +19,9 @@ rules:
 - apiGroups: ["networking.k8s.io"]
   resources: ["networkpolicies", "ingresses", "ingressclasses"]
   verbs: ["get", "list", "watch"]
+- apiGroups: ["cilium.io"]
+  resources: ["ciliumnetworkpolicies", "ciliumclusterwidenetworkpolicies"]
```
byjustina: Should cilium CIDR groups get included here as well?

rlenglet: Yes!

internal/controller/stream/resources/watcher.go
```diff
@@ -107,31 +109,52 @@ func (r *Watcher) WatchK8sResources(ctx context.Context, cancel context.CancelFu
...
+	// Dynamic client may not set GVK on individual list items; derive it from the list GVK.
+	listGvk := unstructuredResources.GroupVersionKind()
+	itemGvk := schema.GroupVersionKind{
+		Group:   listGvk.Group,
+		Version: listGvk.Version,
+		Kind:    removeListSuffix(listGvk.Kind),
+	}
+
+	for i := range unstructuredResources.Items {
+		item := &unstructuredResources.Items[i]
```
rlenglet: Simplify the loop since you don't use the index.
```suggestion
	for _, item := range unstructuredResources.Items {
```

internal/controller/stream/resources/client.go
```diff
@@ -68,6 +70,11 @@ func (c *resourcesClient) Run(ctx context.Context) error {
 		return err
 	}
 
+	coreConverter := controller.NewCoreResourceConverter(clientset, c.logger)
+	ciliumConverter := func(_ context.Context, obj *unstructured.Unstructured) (*pb.KubernetesObjectData, error) {
+		return controller.ConvertUnstructuredToCiliumPolicy(obj)
+	}
```
rlenglet, line 76: Why can't both methods be factorized into one?
It seems like unnecessary complexity. Cilium resources don't need to be handled specially.
Factorize them and remove this configuration mechanism.

```suggestion
```

internal/controller/stream/resources/resources.go
```diff
@@ -9,6 +9,8 @@ import (
 )
 
 var resourceList = []string{
+	"ciliumclusterwidenetworkpolicies",
+	"ciliumnetworkpolicies",
```
rlenglet, line 14: Add `ciliumcidrgroup`.

internal/controller/stream/resources/watcher.go
```diff
@@ -30,6 +28,10 @@ type ResourceStreamSender interface {
 	CreateMutationObject(metadata *pb.KubernetesObjectData, eventType watch.EventType) *pb.KubernetesResourceMutation
 }
 
+// ResourceConverter converts an unstructured Kubernetes object into a
+// KubernetesObjectData proto. Each resource type provides its own implementation.
```
rlenglet: "Each resource type provides its own implementation." <- that comment is false. There are only two implementations: one for Cilium resources, and one for all other resources. And I don't understand why there have to be two implementations.

internal/controller/stream/resources/watcher.go
```diff
@@ -107,31 +109,52 @@ func (r *Watcher) WatchK8sResources(ctx context.Context, cancel context.CancelFu
...
+			item.SetGroupVersionKind(itemGvk)
+		}
 
-		err = r.resourcesClient.SendObjectData(logger, metadataObj)
+		metadataObj, err := r.converter(ctx, item)
 		if err != nil {
+			r.logger.Error("Cannot convert resource",
+				zap.String("name", item.GetName()),
+				zap.String("namespace", item.GetNamespace()),
+				zap.Error(err))
```
rlenglet: namespace+name are not sufficent to identify a resource. The kind is also necessary.
```suggestion
			r.logger.Error("Cannot convert resource",
				zap.String("group", itemGvk.Group),
				zap.String("version", itemGvk.Version),
				zap.String("kind", itemGvk.Kind),
				zap.String("namespace", item.GetNamespace()),
				zap.String("name", item.GetName()),
				zap.Error(err))
```

internal/controller/stream/resources/watcher.go
```diff
@@ -107,31 +109,52 @@ func (r *Watcher) WatchK8sResources(ctx context.Context, cancel context.CancelFu
...
+
+			return "", fmt.Errorf("failed to convert resource %s/%s: %w", item.GetNamespace(), item.GetName(), err)
```
rlenglet:
```suggestion
			return "", fmt.Errorf("failed to convert %s resource %s/%s: %w", itemGvk.Kind, item.GetNamespace(), item.GetName(), err)
```

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
+// Copyright 2024 Illumio, Inc. All Rights Reserved.
```
rlenglet:
```suggestion
// Copyright 2026 Illumio, Inc. All Rights Reserved.
```

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+// Accepts both Kind (PascalCase) and resource name (lowercase plural).
+func IsCiliumPolicy(kindOrResource string) bool {
+	switch kindOrResource {
+	case "CiliumNetworkPolicy", "CiliumClusterwideNetworkPolicy",
+		"ciliumnetworkpolicies", "ciliumclusterwidenetworkpolicies":
+		return true
+	default:
+		return false
+	}
+}
```
rlenglet: Move that switch into the converter function.

```suggestion
```

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+
+// ConvertUnstructuredToCiliumPolicy converts an unstructured Cilium policy to a KubernetesObjectData proto.
+// This handles both CiliumNetworkPolicy and CiliumClusterwideNetworkPolicy.
+func ConvertUnstructuredToCiliumPolicy(obj *unstructured.Unstructured) (*pb.KubernetesObjectData, error) {
```
rlenglet: This must also convert `CiliumCIDRGroup` objects, not just policies.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+}
+
+// extractCiliumSpecs extracts CiliumPolicyRule specs from both 'spec' and 'specs' fields.
+func extractCiliumSpecs(obj *unstructured.Unstructured) []*pb.CiliumPolicyRule {
+	var result []*pb.CiliumPolicyRule
+
+	// Handle single 'spec' field
+	if spec, found, _ := unstructured.NestedMap(obj.Object, "spec"); found {
+		if rule := convertCiliumPolicyRule(spec); rule != nil {
+			result = append(result, rule)
```
rlenglet:
```suggestion
			result = []*pb.CiliumPolicyRule{rule}
```

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		}
+	}
+
+	// Handle 'specs' array field
+	if specs, found, _ := unstructured.NestedSlice(obj.Object, "specs"); found {
```
rlenglet: Pre-allocate the `result` slice:

```suggestion
	if specs, found, _ := unstructured.NestedSlice(obj.Object, "specs"); found {
		result = make([]*pb.CiliumPolicyRule, 0, len(specs))
```

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+	return objMetadata, nil
+}
+
+// extractCiliumSpecs extracts CiliumPolicyRule specs from both 'spec' and 'specs' fields.
+func extractCiliumSpecs(obj *unstructured.Unstructured) []*pb.CiliumPolicyRule {
+	var result []*pb.CiliumPolicyRule
+
+	// Handle single 'spec' field
+	if spec, found, _ := unstructured.NestedMap(obj.Object, "spec"); found {
+		if rule := convertCiliumPolicyRule(spec); rule != nil {
```
rlenglet: This can never return `nil` in practice. Simplify the code and remove this `if`.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+			result = append(result, rule)
+		}
+	}
+
+	// Handle 'specs' array field
+	if specs, found, _ := unstructured.NestedSlice(obj.Object, "specs"); found {
+		for _, s := range specs {
+			if specMap, ok := s.(map[string]any); ok {
+				if rule := convertCiliumPolicyRule(specMap); rule != nil {
```
rlenglet: This can never return `nil` in practice. Simplify the code and remove this `if`.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+	}
+
+	return result
+}
+
+// convertCiliumPolicyRule converts a single Cilium policy spec to proto.
+func convertCiliumPolicyRule(spec map[string]any) *pb.CiliumPolicyRule {
+	if spec == nil {
+		return nil
+	}
```
rlenglet: When would the spec ever be `nil`?
I don't think that's possible.

That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+
+	return objMetadata, nil
+}
+
+// extractCiliumSpecs extracts CiliumPolicyRule specs from both 'spec' and 'specs' fields.
+func extractCiliumSpecs(obj *unstructured.Unstructured) []*pb.CiliumPolicyRule {
+	var result []*pb.CiliumPolicyRule
+
+	// Handle single 'spec' field
+	if spec, found, _ := unstructured.NestedMap(obj.Object, "spec"); found {
```
rlenglet: Is it really OK to ignore **_all errors_** from **_all calls_** to the `unstructured` package???
I don't think it's OK. Properly return them.

Every one of these `convert*` functions must return an `error`.
And those `error`s must be correctly returned back by their callers.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		}
+	default:
+		return nil, fmt.Errorf("unsupported Cilium policy kind: %s", gvk.Kind)
+	}
+
+	return objMetadata, nil
+}
+
+// extractCiliumSpecs extracts CiliumPolicyRule specs from both 'spec' and 'specs' fields.
+func extractCiliumSpecs(obj *unstructured.Unstructured) []*pb.CiliumPolicyRule {
```
rlenglet: extract -> convert, for consistency.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.LabelSelectorRequirement, 0, len(expressions))
+
+	for _, expr := range expressions {
+		exprMap, ok := expr.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make(map[string]string, len(labels))
+
+	for _, label := range labels {
+		labelMap, ok := label.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyIngressRule, 0, len(rules))
+
+	for _, rule := range rules {
+		ruleMap, ok := rule.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyEgressRule, 0, len(rules))
+
+	for _, rule := range rules {
+		ruleMap, ok := rule.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyCIDRSet, 0, len(cidrSets))
+
+	for _, cidrSet := range cidrSets {
+		cidrSetMap, ok := cidrSet.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.LabelSelector, 0, len(selectors))
+
+	for _, selector := range selectors {
+		selectorMap, ok := selector.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyGroup, 0, len(groups))
+
+	for _, group := range groups {
+		groupMap, ok := group.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyICMPRule, 0, len(icmps))
+
+	for _, icmp := range icmps {
+		icmpMap, ok := icmp.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyICMPField, 0, len(fields))
+
+	for _, field := range fields {
+		fieldMap, ok := field.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		if family, found, _ := unstructured.NestedString(fieldMap, "family"); found {
+			protoField.Family = &family
+		}
+
+		// Type can be either int or string in Cilium CRD
+		// ICMP type values are 0-255 (fits in uint8)
+		if typeVal, found, _ := unstructured.NestedFieldNoCopy(fieldMap, "type"); found {
+			switch v := typeVal.(type) {
+			case int64:
+				if v >= 0 && v <= 255 {
```
rlenglet: And if this is not true, that's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+					protoField.Type = &pb.CiliumPolicyICMPField_TypeInt{TypeInt: uint32(v)}
+				}
+			case float64:
+				if v >= 0 && v <= 255 {
```
rlenglet: And if this is not true, that's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		protoField := &pb.CiliumPolicyICMPField{}
+
+		if family, found, _ := unstructured.NestedString(fieldMap, "family"); found {
+			protoField.Family = &family
+		}
+
+		// Type can be either int or string in Cilium CRD
+		// ICMP type values are 0-255 (fits in uint8)
+		if typeVal, found, _ := unstructured.NestedFieldNoCopy(fieldMap, "type"); found {
+			switch v := typeVal.(type) {
```
rlenglet: Add a `default`. If none of the `case`s matches, that's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyFQDNSelector, 0, len(fqdns))
+
+	for _, fqdn := range fqdns {
+		fqdnMap, ok := fqdn.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyService, 0, len(services))
+
+	for _, service := range services {
+		serviceMap, ok := service.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyPortRule, 0, len(portRules))
+
+	for _, portRule := range portRules {
+		portRuleMap, ok := portRule.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyPort, 0, len(ports))
+
+	for _, port := range ports {
+		portMap, ok := port.(map[string]any)
+		if !ok {
+			continue
+		}
```
rlenglet: That's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+
+		if portVal, found, _ := unstructured.NestedString(portMap, "port"); found {
+			protoPort.Port = portVal
+		}
+
+		// endPort for port ranges (valid port numbers: 0-65535)
+		if endPort, found, _ := unstructured.NestedFieldNoCopy(portMap, "endPort"); found {
+			switch v := endPort.(type) {
+			case int64:
+				if v >= 0 && v <= 65535 {
```
rlenglet: And if this is not true, that's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+					ep := int32(v)
+					protoPort.EndPort = &ep
+				}
+			case float64:
+				if v >= 0 && v <= 65535 {
```
rlenglet: And if this is not true, that's an error. Return it as such.

internal/controller/cilium_network_policy.go
```diff
@@ -0,0 +1,772 @@
...
+
+		protoPort := &pb.CiliumPolicyPort{}
+
+		if portVal, found, _ := unstructured.NestedString(portMap, "port"); found {
+			protoPort.Port = portVal
+		}
+
+		// endPort for port ranges (valid port numbers: 0-65535)
+		if endPort, found, _ := unstructured.NestedFieldNoCopy(portMap, "endPort"); found {
+			switch v := endPort.(type) {
```
rlenglet: Add a `default`. If none of the `case`s matches, that's an error. Return it as such.

rlenglet (changes requested):
1. Factorize the "converter" functions into one. I don't see any reason why we need the unnecessary complexity of splitting this into two. It's OK to keep the Cilium-resource-specific logic in a separate file.
1. Ingest `CiliumCIDRGroup` in addition to policies.
1. The policy conversion has _nearly zero_ error handling, and that must be fixed.

rlenglet:
OK with ingesting `CiliumCIDRGroup` in another PR, but you should at least not name files and functions specifically for policies, when we know that they will also handle non-policy resources very soon.

rlenglet:
Fix the title and description of this PR. They are outdated.

internal/controller/cilium_network_policy.go
rlenglet, line 1: We already have a dependency on the `cilium/cilium` repository.
So we can directly use Cilium's `ClientSet`, which directly provides methods to list and watch each Cilium resources and get typed objects.
https://github.com/cilium/cilium/blob/main/pkg/k8s/client/clientset/versioned/clientset.go

No need to reinvent that wheel.
Any concerns about using that directly?

rlenglet (changes requested):

internal/controller/cilium_network_policy.go
byjustina: Would this mean other resources should also use typed clients, or a mix of both typed and dynamic clients? It looks like AWS doesn't have a clientset for EKS CRDs so implementing a typed client for EKS specifically would mean running `client-gen` to update and keeping manual maintenance. But choosing a dynamic client as is currently done would handle both uniformly without that maintenance burden

rlenglet: My point is that we shouldn't reinvent the wheel. Every CNI plugin must already implement the unmarshaling of their CRD objects into Golang structs. We should reuse whatever each CNI plugin provides, instead of trying to reimplement that unmarshaling logic by hand.

If there is a single mechanism / implementation we can use for all CNI plugins, that's even better, of course.

I'd suggest digging into how Cilium and EKS policy agent implement object unmarshaling. For example, how does the Cilium ClientSet implement it? If the internal implementation of the ClientSet is easier to reuse, use that.

rlenglet: Looks like the Cilium ClientSets are implemented automatically using standard k8s APIs:
https://github.com/cilium/cilium/blob/35019efe9980de517ce40b51b1b93f16822dfbdf/pkg/k8s/client/clientset/versioned/typed/cilium.io/v2/ciliumclusterwidenetworkpolicy.go#L45-L59
https://github.com/cilium/cilium/blob/main/pkg/k8s/client/clientset/versioned/scheme/register.go

It seems like we can use the `CodecFactory` configured by Cilium?
https://github.com/cilium/cilium/blob/63d1e513a1175351ade5e474461cd35afd019e79/pkg/k8s/client/clientset/versioned/scheme/register.go#L19

Potentially something like:
```
decoder := scheme.Codecs.UniversalDeserializer()
policy := &ciliumiov2.CiliumClusterwideNetworkPolicy{}
_, _, err := decoder.Decode(rawYAML, nil, policy)
// ...
```

rlenglet: The AWS VPC CNI plugin relies on the k8s API's controller-runtime package, which handles the unmarshaling out of the box:
https://github.com/aws/amazon-network-policy-controller-k8s/blob/ecc7e6a79394528bdb397e1fbd927dae64383bec/internal/controllers/policy_controller.go#L109-L140
https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.24.1/pkg/builder#TypedBuilder.Watches

I went down the rabbit hole to try to find how it does unmarshaling, but couldn't find it. I believe it should use the same Codec mechanism that Cilium is using directly.

rigdomin-illumio: @rlenglet 
Using claude, I investigated three variants and ran a POC spike to verify feasibility. New proposal: **use `runtime.DefaultUnstructuredConverter.FromUnstructured()`** to deserialize `unstructured.Unstructured` into Cilium's typed Go structs, then convert typed structs to proto.

### Why this over ClientSet or CodecFactory

The operator's `Watcher` uses `dynamic.Interface` uniformly for all 22+ resource types (core K8s, Cilium, future EKS). This gives us a single list-watch-convert pipeline with consistent resource-version tracking and retry logic.

- **ClientSet (B1)** would require a separate typed-client path for Cilium resources, fracturing the Watcher architecture. Other resource types (including EKS CRDs that [lack published Go types](https://github.com/aws/amazon-network-policy-controller-k8s)) would still use the dynamic client, creating two parallel watch strategies permanently.

- **CodecFactory (B2)** works but requires scheme registration and a JSON round-trip (`map → json.Marshal → Decode`), since the dynamic client already gives us `map[string]any`, not raw bytes.

- **`DefaultUnstructuredConverter` (B3)** converts `map[string]any` directly to typed structs via reflection — no scheme, no JSON round-trip, no architecture changes. It's a built-in part of `k8s.io/apimachinery` ([docs](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#DefaultUnstructuredConverter), [source](https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/converter.go)). The `ResourceConverter` function signature stays identical; only the converter body changes.

### What the code looks like

```go
func ConvertUnstructuredToCiliumResource(obj *unstructured.Unstructured) (*pb.KubernetesObjectData, error) {
    var cnp ciliumv2.CiliumNetworkPolicy
    if err := runtime.DefaultUnstructuredConverter.FromUnstructured(obj.Object, &cnp); err != nil {
        return nil, fmt.Errorf("deserializing CiliumNetworkPolicy: %w", err)
    }
    // Typed field access — compile-time checked, no string literals
    // cnp.Spec.Ingress[0].FromEndpoints, cnp.Spec.Egress[0].ToFQDNs, etc.
}
```

This eliminates the 65 `NestedX` calls, 57 string-literal field names, 14 silent type-assertion drops, and 65 discarded error returns.

### POC results

We ran a spike importing `cilium.io/v2` with our `k8s.io v0.36.1`:

| Check | Result |
|-------|--------|
| Compiles with k8s.io v0.36.1 | Yes, clean |
| Unknown fields (forward-compat) | Silently ignored |
| Missing fields (backward-compat) | Zero-valued, no errors |
| go.sum growth | +101 lines (189 → 290) |
| New indirect deps | +37 (cilium/ebpf, hive, proxy, etc.) |

One friction point: Cilium's `EndpointSelector` uses [slim `LabelSelector`](https://pkg.go.dev/github.com/cilium/cilium/pkg/k8s/slim/k8s/apis/meta/v1#LabelSelector), not standard `metav1.LabelSelector`. These are not assignable to each other, so we need a small adapter function (~15 lines, reused at 6 call sites).

rlenglet: @rigdomin-illumio that sounds great!

rigdomin-illumio: @rlenglet Update on this approach.
I initially tried runtime.DefaultUnstructuredConverter.FromUnstructured() for deserializing Cilium CRDs into typed Go structs. It panics on Cilium's types.

The problem: 
Cilium's EndpointSelector has unexported fields (cachedLabelSelectorString, sanitized) tagged json:"-". FromUnstructured uses reflection to set struct fields, and Go's reflect package panics when it hits unexported fields from an external package. Since EndpointSelector appears in virtually every Cilium policy type (ingress rules, egress rules, CIDR rules, service selectors, node selectors), there's no way around it.

The better approach (CodecFactory + Dynamic Client): 
Instead will use CodecFactory.UniversalDeserializer().Decode() instead, which internally uses json.Unmarshal. JSON unmarshal respects json:"-" tags and never touches unexported fields. This is also the same mechanism Cilium's own typed ClientSet uses under the hood (scheme.Codecs), and what you've pointed to in the review.

Everything else is the same — typed Cilium structs, compile-time safety, same converter functions. Just a different deserialization entry point.

I'll be updating PR with this

rlenglet:
@rigdomin-illumio please rebase to pick up #419 which will require some changes in your code.

internal/controller/cilium.go
```diff
@@ -0,0 +1,652 @@
...
+			}
+
+			if g.AWS.Region != "" {
+				awsGroup.Region = &g.AWS.Region
+			}
+
+			pbGroup.CloudProvider = &pb.CiliumPolicyGroup_Aws{Aws: awsGroup}
+		}
+
+		result = append(result, pbGroup)
```
byjustina: Should the `pbGroup` still get appended if the cloud provider is empty? Would it make more sense to skip empty groups?

internal/controller/stream/resources/watcher.go
```diff
@@ -368,13 +367,19 @@ func getResourceVersionFromBookmark(event watch.Event) (string, error) {
...
-	resourceGvk := event.Object.GetObjectKind().GroupVersionKind()
-	metadataObj := controller.ConvertMetaObjectToMetadata(ctx, *convertedData, r.clientset, resourceGvk.Kind, resourceGvk.Group, resourceGvk.Version)
+	unstructuredObj, ok := event.Object.(*unstructured.Unstructured)
+	if !ok {
+		return "", fmt.Errorf("expected *unstructured.Unstructured, got %T", event.Object)
+	}
+
+	metadataObj, err := r.converter(ctx, unstructuredObj)
+	if err != nil {
+		return "", fmt.Errorf("failed to convert resource: %w", err)
```
byjustina: It might be helpful to add more information as to what resource failed conversion, like how it was done on line 132
```
  return "", fmt.Errorf("failed to convert resource %s/%s (%s): %w",
      unstructuredObj.GetNamespace(), unstructuredObj.GetName(), unstructuredObj.GetKind(), err)
```

internal/controller/cilium.go
```diff
@@ -0,0 +1,652 @@
...
+		if f.Type != nil {
+			switch f.Type.Type {
+			case k8sIntstr.Int:
+				v := f.Type.IntValue()
+				if v >= 0 && v <= 255 {
+					pbField.Type = &pb.CiliumPolicyICMPField_TypeInt{TypeInt: uint32(v)}
+				}
+			case k8sIntstr.String:
+				s := f.Type.StrVal
+				if intVal, err := strconv.ParseUint(s, 10, 32); err == nil && intVal <= 255 {
```
byjustina, line 450: If a negative numeric string like "-1" is passed in here, I believe it would fail this check and fall in the else case and store "-1", which is inconsistent with the behavior from the above integer case where negative numbers are dropped

pavankumarinnamuri:
@rlenglet, @rigdomin-illumio: Not blocking on this, but the go.mod / go.sum expansion , pulls a lot of additional dependencies. Since this change mainly needs typed CRD structs and decoding, just wanted to check that this is the smallest practical dependency set for the typed Cilium approach. If yes, I’m fine with it; otherwise it may be worth checking whether we can trim the transitive footprint a bit.

rlenglet:
Due to the additional deps, the binary size increased by about 25% compared to the head of `main`: 88,352,573 bytes (this PR) vs. 70,927,493 bytes (`main`). That seems significant.

internal/controller/cilium.go
```diff
@@ -0,0 +1,631 @@
...
+	if len(fields) == 0 {
+		return nil
+	}
+
+	result := make([]*pb.CiliumPolicyICMPField, 0, len(fields))
+	for _, f := range fields {
+		pbField := &pb.CiliumPolicyICMPField{}
+
+		if f.Family != "" {
+			pbField.Family = &f.Family
```
rlenglet, line 438: Any reason for inconsistently copying values into local variables and referring to only those?
E.g. above:
https://github.com/illumio/cloud-operator/pull/371/changes#diff-74c2328a1b03857492d7af3812e9423b0ea86915d2e0fc39d61894496551af6eR360-R363
But here you directly refer to deep struct fields.

internal/controller/cilium.go
```diff
@@ -0,0 +1,631 @@
...
+	result := make([]*pb.CiliumPolicyGroup, 0, len(groups))
+	for _, g := range groups {
+		if g.AWS == nil {
+			continue
+		}
+
+		awsGroup := &pb.CiliumPolicyAWSGroup{
+			Labels:             g.AWS.Labels,
+			SecurityGroupIds:   g.AWS.SecurityGroupsIds,
+			SecurityGroupNames: g.AWS.SecurityGroupsNames,
```
rlenglet, line 546: @rigdomin-illumio @byjustina any concerns about naming consistency here? (Group vs Groups)

rlenglet (approved):
LGTM.
Just a couple minor comments. Nothing blocking.
If the inconsistent naming will require renaming some fields in the API, we can do that in a next PR.
Feel free to resolve the discussions.

PR #383, opened 2026-04-01
Add network policy message handling to configuration stream

sungkyunchung:
Extend the `GetConfigurationUpdates` gRPC stream to support receiving network policies from the server.                  
                                                                                                                                      
- Add handling for network policy messages in the configuration stream (snapshot, mutations)
- Update stats tracking for network policy operations


api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -486,15 +486,34 @@ message GetConfigurationUpdatesResponse {
...
+
+// Signals end of snapshot phase.
+message NetworkPolicySnapshotComplete {}
+
+// Single policy mutation.
+message NetworkPolicyMutation {
+  oneof mutation {
+    NetworkPolicyData create_policy = 1;
+    NetworkPolicyData update_policy = 2;
+    NetworkPolicyData delete_policy = 3;
```
ryan-er, line 773: Delete probably only needs the operator to be able to identify the existing policy right, so the message could be tightened up a little bit if we want. As in it doesn't need to be able to specify any policy definitions.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -486,15 +486,34 @@ message GetConfigurationUpdatesResponse {
...
+  }
+}
 
-    // Additional configuration messages will be added here.
+// Single network policy sent during snapshot or mutation phase.
+message NetworkPolicyData {
+  string id = 1;
+  string name = 2;
+  string namespace = 3;
+  string kind = 4;
```
ryan-er: For my knowledge, what is kind?

sungkyunchung: kind is the Kubernetes resource type such as `CiliumNetworkPolicy`, `CiliumClusterwideNetworkPolicy`, or `NetworkPolicy`. It's the same as the kind field in K8s manifests. We use it to look up the correct GVR (GroupVersionResource) for the dynamic client when creating/deleting policies.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: For delete, we only use id, name, namespace, and kind. The policy resource is ignored. I'll add a comment in the proto to document that resource is optional for delete.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -486,15 +486,34 @@ message GetConfigurationUpdatesResponse {
...
+  bytes resource = 5;  // Full K8s resource as JSON. Optional for delete operations.
```
rlenglet: Use proper protocol buffers. Ideally, this should refer to the messages defined in https://github.com/illumio/cloud-operator/pull/371, to make it easier to implement the diff logic in the operator logic.
Something like:
```suggestion
  oneof kind_specific {
    KubernetesCiliumClusterwideNetworkPolicyData cilium_clusterwide_network_policy = 100;
  }
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -486,15 +486,34 @@ message GetConfigurationUpdatesResponse {
...
+    NetworkPolicyData network_policy_data = 2;
+    NetworkPolicySnapshotComplete network_policy_snapshot_complete = 3;
+    NetworkPolicyMutation network_policy_mutation = 4;
+  }
+}
 
-    // Additional configuration messages will be added here.
+// Single network policy sent during snapshot or mutation phase.
+message NetworkPolicyData {
+  string id = 1;
```
rlenglet, line 744: Add a short comment for each field.
Esp. it must be clear that this `id` field is an ID generated by Illumio.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: Add a comment to document the expected sequence of those `GetConfigurationUpdatesResponse` messages over a configuration stream, separately for the log level configuration and for network policies.
Follow for example the same pattern as for `SendKubernetesResourcesRequest`:
https://github.com/illumio/cloud-operator/blob/a78e731d19235f506e508898d133b217f49684f8/api/illumio/cloud/k8sclustersync/v1/k8s_info.proto#L275-L280

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: This `kind` field is unnecessary if we have a `oneof` below that clearly identifies the kind.
```suggestion
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -486,15 +486,34 @@ message GetConfigurationUpdatesResponse {
...
+
+// Signals end of snapshot phase.
+message NetworkPolicySnapshotComplete {}
+
+// Single policy mutation.
+message NetworkPolicyMutation {
+  oneof mutation {
+    NetworkPolicyData create_policy = 1;
+    NetworkPolicyData update_policy = 2;
+    NetworkPolicyData delete_policy = 3;
```
rlenglet, line 773: Add a short comment for each field.
Follow for example the same pattern as in `KubernetesResourceMutation`: https://github.com/illumio/cloud-operator/blob/a78e731d19235f506e508898d133b217f49684f8/api/illumio/cloud/k8sclustersync/v1/k8s_info.proto#L315-L320

internal/controller/stream/config/config.go
```diff
@@ -15,8 +15,38 @@ import (
 	"github.com/illumio/cloud-operator/internal/pkg/timeutil"
 )
 
+// handlePolicyMutation processes network policy mutation messages.
+func handlePolicyMutation(ctx context.Context, sm *stream.Manager, logger *zap.Logger, mutation *pb.NetworkPolicyMutation) {
```
rlenglet: Keep the naming consistent.

```suggestion
// handleNetworkPolicyMutation processes network policy mutation messages.
func handleNetworkPolicyMutation(ctx context.Context, sm *stream.Manager, logger *zap.Logger, mutation *pb.NetworkPolicyMutation) {
```

internal/controller/stream/config/config.go
```diff
@@ -15,8 +15,38 @@ import (
...
+		}
+	case *pb.NetworkPolicyMutation_UpdatePolicy:
+		// Update = delete + create
+		if err := DeletePolicy(ctx, sm.K8sClient.GetDynamicClient(), logger, m.UpdatePolicy); err != nil {
+			logger.Error("Failed to delete policy for update", zap.Error(err))
+		}
+
+		if err := CreatePolicy(ctx, sm.K8sClient.GetDynamicClient(), logger, m.UpdatePolicy); err != nil {
+			logger.Error("Failed to create policy for update", zap.Error(err))
+		}
```
rlenglet: That is incorrect. A policy update should not cause a policy deletion. That's a security risk.

internal/controller/stream/config/config.go
```diff
@@ -15,8 +15,38 @@ import (
...
+	case *pb.NetworkPolicyMutation_DeletePolicy:
+		if err := DeletePolicy(ctx, sm.K8sClient.GetDynamicClient(), logger, m.DeletePolicy); err != nil {
+			logger.Error("Failed to delete policy", zap.Error(err))
+		}
+	}
```
rlenglet: Remove this. Mutations cannot be directly applied to the cluster, because the server does not exactly know the state of the cluster.
The only effect of receiving a mutation should be limited to updating some in-memory state, which will be used by the operator's reconciliation logic.
To get started, I'd suggest just maintaining a `map[string]<???>` of each policy's `id` to its protobuf (maybe `map[string]*pb.NetworkPolicyMutation`???).
And each mutation either adds, replaces, or deletes the policy from the map using its `id`.

internal/controller/stream/config/config.go
```diff
@@ -51,6 +81,18 @@ func Stream(ctx context.Context, sm *stream.Manager, logger *zap.Logger, keepali
 				} else {
 					sm.BufferedGrpcSyncer.UpdateLogLevel(update.UpdateConfiguration.GetLogLevel())
 				}
+
+			case *pb.GetConfigurationUpdatesResponse_NetworkPolicyData:
+				logger.Info("Received network policy data",
```
rlenglet:
```suggestion
				logger.Debug("Received network policy data",
```

internal/controller/stream/config/config.go
```diff
@@ -51,6 +81,18 @@ func Stream(ctx context.Context, sm *stream.Manager, logger *zap.Logger, keepali
...
+					zap.String("id", update.NetworkPolicyData.GetId()),
+				)
+
+			case *pb.GetConfigurationUpdatesResponse_NetworkPolicySnapshotComplete:
+				logger.Info("Received network policy snapshot complete")
```
rlenglet:
```suggestion
				logger.Debug("Received network policy snapshot complete")
```

internal/controller/stream/config/config.go
```diff
@@ -51,6 +81,18 @@ func Stream(ctx context.Context, sm *stream.Manager, logger *zap.Logger, keepali
 				} else {
 					sm.BufferedGrpcSyncer.UpdateLogLevel(update.UpdateConfiguration.GetLogLevel())
 				}
+
+			case *pb.GetConfigurationUpdatesResponse_NetworkPolicyData:
```
rlenglet: The effect of this should be identical to a `create` mutation.
The map of policy ID to policy should  be (re)initialized to be empty at the top of this function, and each policy in a `pb.GetConfigurationUpdatesResponse_NetworkPolicyData` gets added into the map.

internal/controller/stream/config/config.go
```diff
@@ -15,8 +15,38 @@ import (
...
+
 // Stream handles the configuration update stream.
 func Stream(ctx context.Context, sm *stream.Manager, logger *zap.Logger, keepalivePeriod time.Duration) error {
+	// Initialize GVR cache for policy enforcement
+	if sm.K8sClient != nil {
+		if err := InitPolicyGVRCache(sm.K8sClient.GetClientset(), logger); err != nil {
+			logger.Warn("Failed to initialize policy GVR cache", zap.Error(err))
+		}
+	}
+
```
rlenglet:
```suggestion
```

internal/controller/stream/config/config.go
```diff
@@ -15,8 +15,38 @@ import (
...
+		}
+	case *pb.NetworkPolicyMutation_DeletePolicy:
+		if err := DeletePolicy(ctx, sm.K8sClient.GetDynamicClient(), logger, m.DeletePolicy); err != nil {
+			logger.Error("Failed to delete policy", zap.Error(err))
+		}
+	}
+}
+
 // Stream handles the configuration update stream.
 func Stream(ctx context.Context, sm *stream.Manager, logger *zap.Logger, keepalivePeriod time.Duration) error {
```
rlenglet: Count and print out stats about received mutations, as we do for resource mutations:
https://github.com/illumio/cloud-operator/blob/a78e731d19235f506e508898d133b217f49684f8/internal/controller/stream/resources/resources.go#L196

internal/controller/stream/config/policy.go
rlenglet, line 1: Delete this file. It shouldn't be needed in this PR.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: I recommend that you rebase this PR on top of https://github.com/illumio/cloud-operator/pull/371's branch instead of `main` for now, until that PR gets merged into `main`, so you can directly refer to the protobuf message types.

rlenglet (changes requested):

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,81 @@ message KubernetesNetworkPolicyData {
...
+  // IngressDeny rules explicitly deny incoming traffic.
+  repeated CiliumNetworkPolicyRule ingress_deny_rules = 4;
+  // EgressDeny rules explicitly deny outgoing traffic.
+  repeated CiliumNetworkPolicyRule egress_deny_rules = 5;
+}
+
+// KubernetesCiliumClusterwideNetworkPolicyData contains information specific to a CiliumClusterwideNetworkPolicy object.
+// CiliumClusterwideNetworkPolicy is a cluster-scoped policy that applies across all namespaces.
+// https://docs.cilium.io/en/stable/security/policy/language/#ciliumclusterwidenetworkpolicy
+message KubernetesCiliumClusterwideNetworkPolicyData {
```
ryan-er, line 121: Both `CiliumClusterwideNetworkPolicy` and `CiliumNetworkPolicy` can contain list of rules now. See `Specs` field in https://docs.cilium.io/en/stable/network/kubernetes/policy/#ciliumclusterwidenetworkpolicy.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,81 @@ message KubernetesNetworkPolicyData {
...
+  // Node selector selects the nodes to which this policy applies (for host policies).
+  optional LabelSelector node_selector = 2;
+  // Ingress rules define allowed/denied incoming traffic.
+  repeated CiliumNetworkPolicyRule ingress_rules = 3;
+  // Egress rules define allowed/denied outgoing traffic.
+  repeated CiliumNetworkPolicyRule egress_rules = 4;
+  // IngressDeny rules explicitly deny incoming traffic.
+  repeated CiliumNetworkPolicyRule ingress_deny_rules = 5;
+  // EgressDeny rules explicitly deny outgoing traffic.
+  repeated CiliumNetworkPolicyRule egress_deny_rules = 6;
```
ryan-er: Also what do we do with `Status` of the policy? I believe, it tells us whether it's successfully applied or not, so seems essential.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,81 @@ message KubernetesNetworkPolicyData {
...
+  // Endpoint selector selects the pods to which this policy applies cluster-wide.
+  LabelSelector endpoint_selector = 1;
+  // Node selector selects the nodes to which this policy applies (for host policies).
+  optional LabelSelector node_selector = 2;
+  // Ingress rules define allowed/denied incoming traffic.
+  repeated CiliumNetworkPolicyRule ingress_rules = 3;
+  // Egress rules define allowed/denied outgoing traffic.
+  repeated CiliumNetworkPolicyRule egress_rules = 4;
+  // IngressDeny rules explicitly deny incoming traffic.
+  repeated CiliumNetworkPolicyRule ingress_deny_rules = 5;
```
ryan-er: There's also `EnableDefaultDeny` field that we're missing.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,81 @@ message KubernetesNetworkPolicyData {
...
+  // EgressDeny rules explicitly deny outgoing traffic.
+  repeated CiliumNetworkPolicyRule egress_deny_rules = 6;
+}
+
+// CiliumNetworkPolicyRule defines a single Cilium ingress or egress rule.
+// Cilium rules support richer selectors including identity-based and CIDR-based selection.
+message CiliumNetworkPolicyRule {
+  // FromEndpoints selects source endpoints by labels (for ingress rules).
```
ryan-er: If we plan to use the same message for ingress and egress.  Then I don't think we need to duplicate the fields to set of `from` and `to` as it is right now, right? Basically just name them `endpoints`, `cidr`, and depending on the ingress/egress context we know to treat them as `to` or `from`

rlenglet: Why would we reuse the same message for ingress and egress?

ryan-er: NVM this one, I lost track of who owns what piece of the work. Regardless, I think the `k8s_info.proto` aka the network policy proto messages and API can be finalized first as a separate PR.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,229 @@ message KubernetesNetworkPolicyData {
...
+// identical structure but are kept as separate messages intentionally:
+// 1. Mirrors Cilium's two distinct CRDs (CiliumNetworkPolicy vs CiliumClusterwideNetworkPolicy)
+// 2. Future-proof if Cilium adds fields to one type but not the other
+// 3. The kind_specific oneof field name already distinguishes which type it is
+// The duplication is minimal (one repeated field each) and preserves semantic clarity.
+
+// KubernetesCiliumNetworkPolicyData contains information specific to a CiliumNetworkPolicy object.
+// CiliumNetworkPolicy is a namespaced policy that applies to pods selected by the endpoint selector.
+// https://docs.cilium.io/en/stable/security/policy/language/#ciliumnetworkpolicy
+message KubernetesCiliumNetworkPolicyData {
```
ryan-er, line 112: Nit: this file is getting big. I'd think it's better to split these into different proto files?

rlenglet: Good idea. One file per flow? Plus one for the policy-related messages that should be shared between the resources stream and the configuration stream?

ryan-er: Sure, I'm not that familiar with what's there currently yet, but at least separate it per CNI:
1. cilium flows and netpol
2. aws flows and netpol
3. vanilla k8s netpol
4. some shared/common stuff

rigdomin-illumio: Yea. Something like
`

File | Content
-- | --
k8s_resources.proto | KubernetesObjectData, NetworkPolicyData
flows.proto | CiliumFlow, FiveTupleFlow, flow enums
configuration.proto | LogEntry, GetConfigurationUpdates
service.proto | Service definitions

<br class="Apple-interchange-newline">
`

Then we can add per cni

internal/controller/stream/config/client.go
rigdomin-illumio, line 1: Missing UT

internal/controller/stream/config/factory.go
rigdomin-illumio, line 1: Missing UT

rigdomin-illumio (dismissed):
I don't any additional feedback than what has already been called out

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rigdomin-illumio: I was wondering same thing; but state tracking works well for Greenfield (or CS created policies); not sure how that would reflect for existing Customer Policies. 

If we do want to track state, am thinking something like:
`
  POLICY_STATUS_UNSPECIFIED = 0;
  POLICY_STATUS_PENDING = 1;     // Received, awaiting application
  POLICY_STATUS_PUSHED = 2;      // Applied to CNI API
  POLICY_STATUS_REALIZED = 3;    // CNI confirmed creation
  POLICY_STATUS_ENFORCED = 4;    // Actively enforcing
  POLICY_STATUS_ERROR = 5;       // Application failed
`

sungkyunchung: What's the use case? Is it for logging and testing purposes or do we need to report status back to the server?

rigdomin-illumio: I'd say it's more around report status back.
It's a nice-to-have bc a very common area of interested we've seen from customers on Core Containers side is knowing policy states.

sungkyunchung: More granular statuses could be useful, but they'd belong in a separate message? This proto is used by the watcher to stream resources, so any policy that appears in that stream is already REALIZED/ENFORCED. We can add sth like `ConfiguredNetworkPolicyStatus` message with policy id and status if needed.

internal/controller/stream/config/client.go
sungkyunchung: Added unit tests for mutation counting. Test coverage is limited for now since the current implementation only logs and counts mutations.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: will address this in a separate PR

ryan-er (dismissed):
Looks good to me as this is mostly just stub, we can merge this or make this PR enable more things other than just debugging/logging.

internal/controller/stream/config/client.go
```diff
@@ -75,6 +76,34 @@ func (c *configClient) handleConfigUpdate(resp *pb.GetConfigurationUpdatesRespon
...
 			c.bufferedGrpcSyncer.UpdateLogLevel(update.UpdateConfiguration.GetLogLevel())
 		}
+
+	case *pb.GetConfigurationUpdatesResponse_ResourceData:
+		c.logger.Debug("Received configured object data",
+			zap.String("id", update.ResourceData.GetId()),
+		)
+
+	case *pb.GetConfigurationUpdatesResponse_ResourceSnapshotComplete:
+		c.logger.Debug("Received configured object snapshot complete")
```
rlenglet: This is a more significant event. Log at info level.

```suggestion
		c.logger.Info("Received configured object snapshot complete")
```

internal/controller/stream/config/client.go
```diff
@@ -75,6 +76,34 @@ func (c *configClient) handleConfigUpdate(resp *pb.GetConfigurationUpdatesRespon
...
+			)
+		case *pb.ConfiguredKubernetesObjectMutation_UpdateObject:
+			c.logger.Debug("Received update object mutation",
+				zap.String("id", m.UpdateObject.GetId()),
+			)
+		case *pb.ConfiguredKubernetesObjectMutation_DeleteObject:
+			c.logger.Debug("Received delete object mutation",
+				zap.String("id", m.DeleteObject.GetId()),
+			)
+		}
```
rlenglet, line 109: Add a default clause that logs at error level.

internal/controller/stream/config/client.go
```diff
@@ -75,6 +76,34 @@ func (c *configClient) handleConfigUpdate(resp *pb.GetConfigurationUpdatesRespon
 		} else {
 			c.bufferedGrpcSyncer.UpdateLogLevel(update.UpdateConfiguration.GetLogLevel())
 		}
+
+	case *pb.GetConfigurationUpdatesResponse_ResourceData:
```
rlenglet, line 82: You need to be more strict in validating the state machine:

1. Maintain a boolean state to keep track of whether you already received a snapshot complete.
    1. If that boolean is false, you may only accept `UpdateConfiguration`, `ResourceData`, and `ResourceSnapshotComplete` messages.
    1. Set that boolean to true when receiving a `ResourceSnapshotComplete` message.
    1. If that boolean is true, you may only accept `UpdateConfiguration` and `ResourceMutation` messages.

internal/controller/stream/config/client.go
rlenglet: This logic also needs to do callbacks needed by the upcoming operator reconciliation loop:

1. When receiving a `ResourceSnapshotComplete`, to start the reconciliation loop. The reconciliation loop MUST not start until it got a complete snapshot from both the server (as indicated by this `ResourceSnapshotComplete` message here) and the k8s server (as indicated by the List operation completing).
1. When receiving a `ResourceMutation` message, notify the reconciliation loop, with at least the ID of the mutated resource, so it knows what resources to focus the reconciliation on.

rlenglet (changes requested):

internal/controller/stream/config/client.go
rlenglet: OK to do that in a future PR.

internal/controller/stream/config/client.go
sungkyunchung: Agreed, can we add this together with the in-memory map changes so this boolean state tracking and the maps can be added at the same time?

rlenglet: OK, if you'd like.

internal/controller/stream/config/client.go
sungkyunchung: Should this be at Warn level since the stream can continue?

internal/controller/stream/config/client.go
sungkyunchung: will do this in a future PR when we implement the reconciliation loop..

internal/controller/stream/config/client.go
rlenglet: We can't continue, since the server is not respecting the protocol. Or something in the middle, e.g. an HTTPS proxy, is messing up with the data sent by the server. So we can't rely on the data that we receive.
So actually, in addition to logging at error level, we should return an error to close the stream.

rlenglet (approved):
Great, thank you!

PR #387, opened 2026-04-03
Feature/stream client factory tests

pavankumarinnamuri:
Comprehensive unit tests for stream factory components and Cilium flow integration tests.


fakeserver/state.go
```diff
@@ -0,0 +1,176 @@
...
+// StreamState tracks the state of individual streams.
+type StreamState struct {
+	Opened           bool
+	LastActivity     time.Time
+	MessagesReceived int
+	KeepalivesRecv   int
+}
+
+// EnhancedServerState provides detailed tracking of all server activity.
+type EnhancedServerState struct {
```
rlenglet: Remove "Enhanced". It doesn't provide any useful information.

```suggestion
// ServerState provides detailed tracking of all server activity.
type ServerState struct {
```

fakeserver/state.go
```diff
@@ -0,0 +1,176 @@
...
+
+	// Flow tracking
+	CiliumFlowsReceived    int
+	FiveTupleFlowsReceived int
+}
+
+// NewEnhancedServerState creates a new state tracker.
+func NewEnhancedServerState() *EnhancedServerState {
+	return &EnhancedServerState{}
+}
```
rlenglet:
```suggestion
// NewServerState creates a new state tracker.
func NewServerState() *ServerState {
	return &ServerState{}
}
```

internal/controller/stream/config/client_test.go
```diff
@@ -0,0 +1,190 @@
...
+			},
+		},
+	}
+	s.mockStream.On("Recv").Return(configResp, nil).Once()
+	s.mockStream.On("Recv").Return(nil, io.EOF).Once()
+
+	err := s.client.Run(context.Background())
+
+	s.Require().NoError(err)
+	s.mockStream.AssertExpectations(s.T())
```
rlenglet, line 106: Does this test that the log level has actually been set to Debug level?

internal/controller/stream/flows/cilium/factory_test.go
rlenglet, line 1: What these test cases cover feels too basic. :/

internal/controller/stream/flows/falco/client_test.go
rlenglet, line 1: There's so much more that can and should be done here.
This test should send an HTTP request to the server and check that the flow gets written into the FlowSink, etc.

OK to do that in a separate PR.

internal/controller/stream/flows/ovnk/client_test.go
rlenglet, line 1: Same as for the `falco` tests, this test should send IPFIX messages to the server and check that the flows get decoded properly and sent to the FlowSink.

OK to do that in a separate PR.

internal/controller/stream/flows/detect_test.go
rlenglet, line 1: Wouldn't it be possible to test all the cases of the detection logic using mocks?

OK to do that in a separate PR.

internal/controller/stream/resources/resources_test.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+
+		for _, expected := range expectedResources {
+			assert.True(t, resourceSet[expected], "expected resource %q not found in resourceList", expected)
+		}
+	})
+
+	t.Run("has expected length", func(t *testing.T) {
+		assert.Len(t, resourceList, len(expectedResources), "resourceList length mismatch")
+	})
```
rlenglet: Really, this is testing that `resourceList` exactly equals a static list here.
```suggestion
	assert.ElementsMatch(t, resourceList, expectedResources)
```
This looks silly. I'd remove this test case altogether.

internal/controller/stream/deadlock_test.go
rlenglet, line 1: Rename into `deadlock_test.go`.

rlenglet (changes requested):
Thank you!
Most of my comments are about relatively minor things, or suggestions to improve coverage (esp. on the Falco server and the OVN-k IPFIX server) in subsequent PRs.

rlenglet (approved):

PR #392, opened 2026-04-14
Add proto message types for Cilium and network policy

sungkyunchung:
- Add Cilium network policy proto messages (KubernetesCiliumNetworkPolicyData, KubernetesCiliumClusterwideNetworkPolicyData, and related rule/selector types)
- Add ConfiguredKubernetesObjectData proto message
- Make namespace field optional in KubernetesObjectData to properly model cluster-scoped resources that have no namespace
- Update ConvertMetaObjectToMetadata to set Namespace to nil for cluster-scoped resources instead of empty string and add unit test for this


api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -481,20 +709,68 @@ message GetConfigurationUpdatesRequest {
...
-    // Additional configuration messages will be added here.
+// Single network policy sent during snapshot or mutation phase.
+message NetworkPolicyData {
+  // Unique identifier generated by Illumio for tracking this policy.
+  string id = 1;
+  // Kubernetes resource name for the policy.
+  string name = 2;
+  // Kubernetes namespace. Empty for cluster-scoped policies.
+  string namespace = 3;
+  // Kubernetes labels to apply to the policy resource.
```
ryan-er: Aren't name and namespace already there be in the data object?

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -481,20 +709,68 @@ message GetConfigurationUpdatesRequest {
...
+  map<string, string> labels = 4;
+  // Kubernetes annotations to apply to the policy resource.
+  map<string, string> annotations = 5;
```
ryan-er: What Illumio labels or annotations are we thinking here? If not clear perhaps we don't need these fields at this moment? I know we spoke about `ownerReference` but that's not a label.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -481,20 +709,68 @@ message GetConfigurationUpdatesRequest {
...
 
-    // Additional configuration messages will be added here.
+// Single network policy sent during snapshot or mutation phase.
+message NetworkPolicyData {
+  // Unique identifier generated by Illumio for tracking this policy.
+  string id = 1;
+  // Kubernetes resource name for the policy.
+  string name = 2;
+  // Kubernetes namespace. Empty for cluster-scoped policies.
+  string namespace = 3;
```
ryan-er: I think this `namespace` thing can be tightened up a little bit more. We can separate between cluster scoped policies and namespace scoped policies i.e. we can have `ClusterNetworkPolicyData` and `NamespaceNetworkPolicyData`, only the latter would have this `namespace` field. It's a cleaner API that way.

so NetworkPolicyData can be
``` proto
message NetworkPolicyData {
  string id = 1;
  string name = 2;
  map<string, string> labels = 4;
  map<string, string> annotations = 5;
  oneof scope {
    NamespaceNetworkPolicyData namespace_policy = 6;
    ClusterNetworkPolicyData cluster_policy = 7;
  }
}

message NamespaceNetworkPolicyData {
   string namespace = 1;
   oneof kind_specific {
    KubernetesNetworkPolicyData network_policy = 2;
    KubernetesCiliumNetworkPolicyData cilium_network_policy = 3;
   }
}

message ClusterNetworkPolicyData {
   oneof kind_specific {
    KubernetesCiliumClusterwideNetworkPolicyData cilium_clusterwide_network_policy = 1;
  }
}
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -481,20 +709,68 @@ message GetConfigurationUpdatesRequest {
...
+
+// A network policy mutation to apply in the cluster.
+message NetworkPolicyMutation {
+  oneof mutation {
+    // Metadata of a new network policy to create in the cluster.
+    NetworkPolicyData create_policy = 1;
+    // Metadata of an existing network policy to update in the cluster.
+    NetworkPolicyData update_policy = 2;
+    // Metadata of an existing network policy to delete from the cluster.
+    NetworkPolicyData delete_policy = 3;
```
ryan-er: I still think `delete_policy` should just be a separate message type, that only has name of policy and its scope (whether it's cluster wide or namespace).

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -481,20 +709,68 @@ message GetConfigurationUpdatesRequest {
...
+    // Indicates that all network policies have been sent in the initial snapshot.
+    NetworkPolicySnapshotComplete network_policy_snapshot_complete = 3;
+    // A mutation to a network policy in the cluster.
+    NetworkPolicyMutation network_policy_mutation = 4;
+  }
+}
 
-    // Additional configuration messages will be added here.
+// Single network policy sent during snapshot or mutation phase.
+message NetworkPolicyData {
```
ryan-er: This name is quite confusing vs `KubernetesNetworkPolicyData`, though I don't have a good suggestion yet to solve this.

sungkyunchung: What'd be a good name? Some options are CloudSecureNetworkPolicyData, ManagedNetworkPolicyData, NetworkPolicyEnforcement, etc

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: delete operation still needs policy id, policy name, namespace (for namespace-scoped policies) and cni types. To follow the similar structure, delete_policy will have the same scope field to determine the cni types and its scope.

rigdomin-illumio: I see your point and I agree with this suggestion, since the argument here is:
`4-5 fields of duplication or carrying 10+ irrelevant fields`

However, This existing approach is offers simplicity and avoids any potential issue with aligning Proto field numbers. Let's say we separate, and months, years later down we want to add a new field to delete: do we align the value between the two, which could mean we skip value in delete, or we simply increment and be aware they don't align. If the proto numbers don't align, we may not be able to share serialization logic between them.

Also, this is modeled after existing design for consistency.
https://github.com/illumio/cloud-operator/blob/1f50dc5c3b3eca115eb8f710f4dcd1c400157285/api/illumio/cloud/k8sclustersync/v1/k8s_info.proto#L239-L243 

I'm open either way, but if we are going to do this for Cilium, I'd recommend to retroactively update existing, so we can remain consistent moving forward.

cc: @rlenglet thoughts?

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rigdomin-illumio: If we have to rename, I'd vote for `EnforcedNetworkPolicyData`, as it's most neutral and does assume type, k8s v celium, etc. Regardless of direction, it's existence implies that Policy would be "Enforced".
The name was generic to future-proof CNI Agnostic abstraction for simplicity.

Could you give recommendation on what name "should" be, or why it "should not" be so generic?

rigdomin-illumio (dismissed):
LGTM, but I'm biased 😄

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: I need to correct my earlier comment. Since we store all policies in-memory, the only field we need for a delete operation is the policy ID. The in-memory map stores the entire NetworkPolicyData message, which already contains all the essential fields (name, namespace, type) needed to remove the policy from the cluster. Given this, I think it makes sense to have a separate, minimal message type for delete with just the ID field.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
ryan-er: It's more of a nit. That it's so similar to `KubernetesNetworkPolicyData`. If I see both `NetworkPolicyData` and `KubernetesNetworkPolicyData`, the name doesn't really tell me what is different between those two. Maybe it's ok though.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -481,20 +709,97 @@ message GetConfigurationUpdatesRequest {
...
+
+// Namespaced policy data.
+message NamespacedPolicyData {
+  // Kubernetes namespace for the policy.
+  string namespace = 1;
+  // Policy specification based on the CNI type.
+  oneof policy {
+    KubernetesNetworkPolicyData k8s_network_policy = 2;
+    KubernetesCiliumNetworkPolicyData cilium_network_policy = 3;
+  }
```
ryan-er, line 770: Yeah, I like this more, it gives clear separation on which policies are cluster wide and which are not.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -481,20 +709,97 @@ message GetConfigurationUpdatesRequest {
...
+}
+
+// Network policy reference for delete operations.
+message DeleteNetworkPolicy {
+  // Unique identifier generated by Illumio for tracking this policy.
+  string id = 1;
+  // Kubernetes resource name for the policy.
+  string name = 2;
+  // Policy scope - namespaced or cluster-wide.
+  // The inner policy is empty; only the type matters for delete.
```
ryan-er: I don't really like this as much, as it's relying on comment on know what field is going to be populated i.e. just the namespace for `NamespacedPolicyData`. If we want parity with the other message, might as well make new messages i.e. `DeleteNamespacePolicy`, `DeleteClusterPolicy` (which would be empty struct message). They are not costly and provide clearer APIs anyway.

ryan-er (dismissed):
LGTM.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rigdomin-illumio: I think we need to think about the behavior of the 2 types:
- KubernetesNetworkPolicyData: is more about ingestion of NPD seen in the cluster and is reported to CS
- NetworkPolicyData: is pushed to Operator(cluster) from CS for Enforcement

Because of this, perhaps we can expand on name to reflect either behavior or direction.
Suggestions:
- EnforcedNetworkPolicyData
- ManagedNetworkPolicyData
- ProvisionedNetworkPolicyData
- CloudSecureNetworkPolicyData
- OutboundNetworkPolicyData
- DownstreamNetworkPolicyData
etc.

thoughts?

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rigdomin-illumio: I agree for explicit separation of behaviors, and safe message constructs.
With your suggestions it might end up looking like this:

`
// Delete a namespaced policy (e.g., CiliumNetworkPolicy)
message DeleteNamespacedNetworkPolicy {
  string policy_id = 1;
  string name = 2;
  string namespace = 3;  // Required — always meaningful
}

// Delete a cluster-scoped policy (e.g., CiliumClusterwideNetworkPolicy)
message DeleteClusterNetworkPolicy {
  string policy_id = 1;
  string name = 2;
  // No namespace field — impossible to misuse
}

// Usage in the config stream
message NetworkPolicyOperation {
  oneof operation {
    NetworkPolicyData create_policy = 1;
    NetworkPolicyData update_policy = 2;
    DeleteNamespacedNetworkPolicy delete_namespaced_policy = 3;
    DeleteClusterNetworkPolicy delete_cluster_policy = 4;
  }
}
`

The question then is, do we want a full separation of messages to include Create/Update as well?

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: EnforcedNetworkPolicyData gives the impression that the policy has already been enforced, whereas this message is more of an enforcement request coming from the server. I'd vote for `ManagedNetworkPolicyData` or `DesiredNetworkPolicyData`.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: To correct my earlier message, the only field we need for a delete operation is the policy ID. The in-memory map will store all the required fields to remove the policy from the cluster, so DeleteNetworkPolicy only needs the ID field. I don't think we need to separate Create/Update unless there are fields that are irrelevant like Delete.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
ryan-er: I like `ProvisionedNetworkPolicyData` or `ConfiguredNetworkPolicyData`.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: This should have been declared `optional`, to cleanly model clusterwide resources.
```suggestion
  optional string namespace = 6;
```
It's ok to do that now. It won't break backward compatibility on the wire.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+}
+
+// CiliumPolicySpec represents a single spec within a Cilium policy.
+// Maps to Cilium's Rule type which contains selectors, metadata, and rules.
+// https://docs.cilium.io/en/stable/security/policy/language/#rule
+message CiliumPolicySpec {
+  // Endpoint selector selects the pods to which this spec's rules apply.
+  LabelSelector endpoint_selector = 1;
+  // Node selector selects the nodes to which this spec applies (for host policies).
+  optional LabelSelector node_selector = 2;
```
rlenglet, line 136:
```suggestion
  // Node selector selects the nodes to which this spec applies (for host policies).
  // May only be set in CiliumClusterwideNetworkPolicy resources.
  optional LabelSelector node_selector = 2;
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+  repeated CiliumPolicySpec specs = 1;
+}
+
+// KubernetesCiliumClusterwideNetworkPolicyData contains information specific to a CiliumClusterwideNetworkPolicy object.
+// CiliumClusterwideNetworkPolicy is a cluster-scoped policy that applies across all namespaces.
+// https://docs.cilium.io/en/stable/security/policy/language/#ciliumclusterwidenetworkpolicy
+message KubernetesCiliumClusterwideNetworkPolicyData {
+  // Policy specs from both `spec` (single) and `specs` (array) fields, unified as an array.
+  // Each spec has its own endpoint/node selector and rules, mirroring Cilium's CRD structure.
+  repeated CiliumPolicySpec specs = 1;
```
rlenglet: The `spec` field is a rule, and `specs` is a list of rules, per the spec:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/k8s/apis/cilium.io/v2/ccnp_types.go#L36-L44

Don't invent your own types. Just follow the Cilium source code.

```suggestion
  // Rules from both `spec` (single) and `specs` (array) fields, unified as an array.
  repeated CiliumPolicyRule specs = 1;
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+// Note: KubernetesCiliumNetworkPolicyData and KubernetesCiliumClusterwideNetworkPolicyData have
+// identical structure but are kept as separate messages intentionally:
+// 1. Mirrors Cilium's two distinct CRDs (CiliumNetworkPolicy vs CiliumClusterwideNetworkPolicy)
+// 2. Future-proof if Cilium adds fields to one type but not the other
+// 3. The kind_specific oneof field name already distinguishes which type it is
+// The duplication is minimal (one repeated field each) and preserves semantic clarity.
+
+// KubernetesCiliumNetworkPolicyData contains information specific to a CiliumNetworkPolicy object.
+// CiliumNetworkPolicy is a namespaced policy that applies to pods selected by the endpoint selector.
+// https://docs.cilium.io/en/stable/security/policy/language/#ciliumnetworkpolicy
```
rlenglet: Refer to the code, which is more clear:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/k8s/apis/cilium.io/v2/cnp_types.go#L31-L55

```suggestion
// https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/k8s/apis/cilium.io/v2/cnp_types.go#L31-L55
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+message KubernetesCiliumNetworkPolicyData {
+  // Policy specs from both `spec` (single) and `specs` (array) fields, unified as an array.
+  // Each spec has its own endpoint selector and rules, mirroring Cilium's CRD structure.
+  repeated CiliumPolicySpec specs = 1;
+}
+
+// KubernetesCiliumClusterwideNetworkPolicyData contains information specific to a CiliumClusterwideNetworkPolicy object.
+// CiliumClusterwideNetworkPolicy is a cluster-scoped policy that applies across all namespaces.
+// https://docs.cilium.io/en/stable/security/policy/language/#ciliumclusterwidenetworkpolicy
```
rlenglet: Refer to the code, which is more clear:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/k8s/apis/cilium.io/v2/ccnp_types.go#L26-L54

```suggestion
// https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/k8s/apis/cilium.io/v2/ccnp_types.go#L26-L54
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+// 3. The kind_specific oneof field name already distinguishes which type it is
+// The duplication is minimal (one repeated field each) and preserves semantic clarity.
+
+// KubernetesCiliumNetworkPolicyData contains information specific to a CiliumNetworkPolicy object.
+// CiliumNetworkPolicy is a namespaced policy that applies to pods selected by the endpoint selector.
+// https://docs.cilium.io/en/stable/security/policy/language/#ciliumnetworkpolicy
+message KubernetesCiliumNetworkPolicyData {
+  // Policy specs from both `spec` (single) and `specs` (array) fields, unified as an array.
+  // Each spec has its own endpoint selector and rules, mirroring Cilium's CRD structure.
+  repeated CiliumPolicySpec specs = 1;
```
rlenglet: The `spec` field is a rule, and `specs` is a list of rules, per the spec:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/k8s/apis/cilium.io/v2/cnp_types.go#L40-L48

Don't invent your own types. Just follow the Cilium source code.

```suggestion
  // Rules from both `spec` (single) and `specs` (array) fields, unified as an array.
  repeated CiliumPolicyRule specs = 1;
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+message KubernetesCiliumClusterwideNetworkPolicyData {
+  // Policy specs from both `spec` (single) and `specs` (array) fields, unified as an array.
+  // Each spec has its own endpoint/node selector and rules, mirroring Cilium's CRD structure.
+  repeated CiliumPolicySpec specs = 1;
+}
+
+// CiliumPolicySpec represents a single spec within a Cilium policy.
+// Maps to Cilium's Rule type which contains selectors, metadata, and rules.
+// https://docs.cilium.io/en/stable/security/policy/language/#rule
+message CiliumPolicySpec {
```
```suggestion
// CiliumPolicyRule represents a single rule within a Cilium policy.
// Maps to Cilium's Rule type which contains selectors, metadata, and rules.
// https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/api/rule.go#L57-L151
message CiliumPolicyRule {
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+
+// CiliumIngressRule defines a Cilium ingress rule.
+// Used for both allow and deny rules; allow vs deny is determined by which array
+// the rule belongs to (ingress_rules vs ingress_deny_rules) in CiliumPolicySpec.
+// Separate from CiliumEgressRule to mirror Cilium's CRD structure and ensure
+// ingress rules only have from_* fields.
+// https://docs.cilium.io/en/stable/security/policy/language/#ingress
+message CiliumIngressRule {
+  // FromEndpoints selects source endpoints by labels.
+  repeated LabelSelector from_endpoints = 1;
```
rlenglet: This is one of those repeated fields where we need to differentiate between `null` and `[]`.
Per the spec:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/api/ingress.go#L26-L29

> 	// Note that while an empty non-nil FromEndpoints does not select anything,
>	// nil FromEndpoints is implicitly treated as a wildcard selector if ToPorts
>	// are also specified.
>	// To select everything, use one EndpointSelector without any match requirements.

So you must wrap the repeated field into a message and make this field here `optional`.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+  optional CiliumAuthentication authentication = 9;
+}
+
+// CiliumEgressRule defines a Cilium egress rule.
+// Used for both allow and deny rules; allow vs deny is determined by which array
+// the rule belongs to (egress_rules vs egress_deny_rules) in CiliumPolicySpec.
+// https://docs.cilium.io/en/stable/security/policy/language/#egress
+message CiliumEgressRule {
+  // ToEndpoints selects destination endpoints by labels.
+  repeated LabelSelector to_endpoints = 1;
```
rlenglet: Same as for `CiliumIngressRule`, I think you must wrap this an `optional` message to be able to differentiate between `null` and `[]`.
Per the spec:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/api/egress.go#L25-L28

> 	// Note that while an empty non-nil ToEndpoints does not select anything,
>	// nil ToEndpoints is implicitly treated as a wildcard selector if ToPorts
>	// are also specified.
>	// To select everything, use one EndpointSelector without any match requirements.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+  // ICMPs specifies ICMP type/code matching rules.
+  repeated CiliumICMPRule icmps = 8;
+  // Authentication specifies required authentication for allowed traffic.
+  // Note: Not applicable for deny rules (field is ignored).
+  optional CiliumAuthentication authentication = 9;
+}
+
+// CiliumEgressRule defines a Cilium egress rule.
+// Used for both allow and deny rules; allow vs deny is determined by which array
+// the rule belongs to (egress_rules vs egress_deny_rules) in CiliumPolicySpec.
```
```suggestion
// the rule belongs to (egress_rules vs egress_deny_rules) in CiliumPolicyRule.
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+message CiliumDefaultDeny {
+  // If true, enables default deny for ingress traffic.
+  optional bool ingress = 1;
+  // If true, enables default deny for egress traffic.
+  optional bool egress = 2;
+}
+
+// CiliumIngressRule defines a Cilium ingress rule.
+// Used for both allow and deny rules; allow vs deny is determined by which array
+// the rule belongs to (ingress_rules vs ingress_deny_rules) in CiliumPolicySpec.
```
```suggestion
// the rule belongs to (ingress_rules vs ingress_deny_rules) in CiliumPolicyRule.
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+// Separate from CiliumEgressRule to mirror Cilium's CRD structure and ensure
+// ingress rules only have from_* fields.
+// https://docs.cilium.io/en/stable/security/policy/language/#ingress
+message CiliumIngressRule {
```
rlenglet: I would prefer to prefix every one of the new messages with "CiliumPolicy*" for consistency, and to clarify that they are all about policies.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+  repeated CiliumIngressRule ingress_rules = 6;
+  // Egress rules define allowed outgoing traffic.
+  repeated CiliumEgressRule egress_rules = 7;
+  // IngressDeny rules explicitly deny incoming traffic.
+  repeated CiliumIngressRule ingress_deny_rules = 8;
+  // EgressDeny rules explicitly deny outgoing traffic.
+  repeated CiliumEgressRule egress_deny_rules = 9;
+  // API version of the source Cilium CRD (e.g., "cilium.io/v2").
+  // Useful for debugging and supporting multiple Cilium API versions.
+  optional string api_version = 10;
```
rlenglet: This should already be captured in `KubernetesObjectData`.
```suggestion
```

We don't capture that currently, but it's easy to include that for every resource.
`ListResources` can easily return group-version-kind, since [`unstructured.UnstructuredList.GroupVersionKind`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#UnstructuredList.GroupVersionKind) returns all the info we need. We don't return that now, but we easily can, and add it into each `KubernetesObjectData`.
https://github.com/illumio/cloud-operator/blob/b7b8b46c9f95868b1f56031b0ff8278fd0f8ebfa/internal/controller/stream/resources/watcher.go#L249

As for mutations, each `Event` we receive from the watcher here:
https://github.com/illumio/cloud-operator/blob/b7b8b46c9f95868b1f56031b0ff8278fd0f8ebfa/internal/controller/stream/resources/watcher.go#L165
Already contains the group-version-kind: `Event.Object` is of type `runtime.Object`, on which we can call [`event.Object.GetObjectKind().GroupVersionKind()`](https://pkg.go.dev/k8s.io/apimachinery@v0.35.4/pkg/runtime/schema#ObjectKind). And then it's easy to add that into the `KubernetesObjectData` for each mutation.

I'm OK with doing that change in another PR.
But please remove this `api_version` field here as it will be redundant.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+
+// CiliumCIDRSet defines a CIDR range with optional exceptions.
+message CiliumCIDRSet {
+  // CIDR range (e.g., "10.0.0.0/8")
+  string cidr = 1;
+  // Exception CIDRs to exclude from the range
+  repeated string except = 2;
+  // Reference to a CiliumCIDRGroup custom resource by name
+  optional string cidr_group_ref = 3;
+}
```
rlenglet, line 235: Add the missing `cidrGroupSelector` field:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/api/cidr.go#L40-L44
```suggestion
  // Selects CiliumCIDRGroups by their labels, rather than by name
  optional LabelSelector cidr_group_selector = 4;
}
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+
+// CiliumAWSGroup defines AWS-specific security group selector fields.
+message CiliumAWSGroup {
+  // AWS security groups tags.
+  map<string, string> labels = 1;
+  // Security group IDs (Cilium supports multiple per group selector)
+  repeated string security_group_ids = 2;
+  repeated string security_group_names = 3;
+  optional string region = 4;
+  optional string vpc_id = 5;
```
rlenglet: Where does this come from?
There's no such field in the Cilium API spec:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/api/groups.go#L33-L39
```suggestion
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+  // ICMP fields for IPv4 or IPv6
+  repeated CiliumICMPField fields = 1;
+}
+
+// CiliumICMPField defines a single ICMP type/code pair.
+message CiliumICMPField {
+  // ICMP type (e.g., 8 for echo request)
+  uint32 type = 1;
+  // ICMP code (optional, family-specific)
+  optional uint32 code = 2;
```
rlenglet: Where did you find a reference to `code`?
There are only 2 fields in the `ICMPField` in the Cilium API spec:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/api/icmp.go#L76-L109
The doc is clear too:
https://docs.cilium.io/en/stable/security/policy/layer4/#limit-icmp-icmpv6-types
```suggestion
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+// Services can be selected by name or by label selector.
+// https://docs.cilium.io/en/stable/security/policy/language/#primitives-with-kubernetes-service-resolution
+message CiliumService {
+  // Kubernetes service name (used with k8s_namespace for direct reference)
+  optional string k8s_service = 1;
+  // Kubernetes namespace (optional, defaults to policy namespace)
+  optional string k8s_namespace = 2;
+  // Label selector to match services by labels instead of by name
+  optional LabelSelector k8s_service_selector = 3;
+}
```
rlenglet: Follow closer the API spec:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/api/service.go#L9-L18
Cf. also the docs: https://docs.cilium.io/en/latest/security/policy/layer3/#services-based

We can't rearrange the structure and move fields around, as we might lose the semantics.
Especially, Cilium evaluates `k8sServiceSelector` first, and if that doesn't match it evalutes `k8sService`:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/k8s/service.go#L304-L316
If we eliminate the distinction between `k8sServiceSelector` and `k8sService`, we can lose that semantics.

It should be something like:
```
message CiliumPolicyService {
  // Selects services by k8s labels and namespace
  optional CiliumPolicyK8sServiceSelectorNamespace k8s_service_selector = 1;
  // Selects service by name and namespace pair
  optional string k8s_service = 2;
}

message CiliumPolicyK8sServiceSelectorNamespace {
  LabelSelector selector = 1;
  optional string namespace = 2;
}

message CiliumPolicyK8sServiceNamespace {
  optional string service_name = 1;
  optional string namespace = 2;
}
```
Make sure to add comments to each of these fields.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+
+// CiliumPortRule defines allowed ports and protocols for Cilium policies.
```
```suggestion
// CiliumPortRule defines allowed ports and protocols for Cilium policies.
// Note that none of the L7 fields from Cilium's PortRule are included: https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/api/l4.go#L208-L258
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+// https://docs.cilium.io/en/stable/security/policy/language/#icmp
+message CiliumICMPRule {
+  // ICMP fields for IPv4 or IPv6
+  repeated CiliumICMPField fields = 1;
+}
+
+// CiliumICMPField defines a single ICMP type/code pair.
+message CiliumICMPField {
+  // ICMP type (e.g., 8 for echo request)
+  uint32 type = 1;
```
rlenglet: `type` is either an int or a string.
Model this properly to support both, e.g. using a `oneof`.
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/api/icmp.go#L108

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+
+// CiliumPort defines a single port with protocol, with optional port range support.
+message CiliumPort {
+  // Port number or named port (e.g., "80", "http")
+  string port = 1;
+  // Protocol for this port rule
+  CiliumProtocol protocol = 2;
+  // End of port range (0-65535). If specified, port must be numeric.
+  // Traffic to ports [port, end_port] will be matched.
+  optional int32 end_port = 3;
```
rlenglet: Reorder: `port`, then `end_port`, then `protocol`.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -97,6 +99,232 @@ message KubernetesNetworkPolicyData {
...
+// CiliumProtocol defines the transport protocol for Cilium port rules.
+enum CiliumProtocol {
+  CILIUM_PROTOCOL_UNSPECIFIED = 0;
+  CILIUM_PROTOCOL_TCP = 1;
+  CILIUM_PROTOCOL_UDP = 2;
+  CILIUM_PROTOCOL_SCTP = 3;
+  CILIUM_PROTOCOL_ICMP = 4;
+  CILIUM_PROTOCOL_ICMPV6 = 5;
+  CILIUM_PROTOCOL_ANY = 6;
+}
```
rlenglet: This is missing some of the possible values:
https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/api/l4.go#L12-L25
I'd recommend replacing this enum with a string to avoid the pain of listing everything.

rlenglet (approved):
Did one pass on the API spec.

rlenglet (changes requested):
Oops I meant to mark my review as "request changes."

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -485,20 +713,91 @@ message GetConfigurationUpdatesRequest {
...
+    ConfiguredNetworkPolicyData network_policy_data = 2;
+    // Indicates that all network policies have been sent in the initial snapshot.
+    NetworkPolicySnapshotComplete network_policy_snapshot_complete = 3;
+    // A mutation to a network policy in the cluster.
+    NetworkPolicyMutation network_policy_mutation = 4;
+  }
+}
+
+// Configured network policy data for snapshot or create/update operations.
+message ConfiguredNetworkPolicyData {
```
rlenglet: Add the group+version as I commented for the resources stream. Cf. https://github.com/illumio/cloud-operator/pull/392/changes#r3112633567

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -485,20 +713,91 @@ message GetConfigurationUpdatesRequest {
...
+  // Kubernetes resource name for the policy.
+  string name = 2;
+  // Kubernetes labels to apply to the policy resource.
+  map<string, string> labels = 3;
+  // Kubernetes annotations to apply to the policy resource.
+  map<string, string> annotations = 4;
+  // Policy scope - namespaced or cluster-wide.
+  oneof scope {
+    NamespacedPolicyData namespaced_policy = 5;
+    ClusterPolicyData cluster_policy = 6;
```
rlenglet: The fields in this `oneof` will grow, and we'll likely keep adding common fields outside of the `oneof`.
Keeping the tags easy to manage, start the tags in the `oneof` higher, e.g. from `100`.

```suggestion
    NamespacedPolicyData namespaced_policy = 100;
    ClusterPolicyData cluster_policy = 101;
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -485,20 +713,91 @@ message GetConfigurationUpdatesRequest {
...
+}
+
+// Namespaced policy data.
+message NamespacedPolicyData {
+  // Kubernetes namespace for the policy.
+  string namespace = 1;
+  // Policy specification based on the CNI type.
+  oneof policy {
+    KubernetesNetworkPolicyData k8s_network_policy = 2;
+    KubernetesCiliumNetworkPolicyData cilium_network_policy = 3;
```
```suggestion
    KubernetesNetworkPolicyData k8s_network_policy = 100;
    KubernetesCiliumNetworkPolicyData cilium_network_policy = 101;
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -485,20 +713,91 @@ message GetConfigurationUpdatesRequest {
...
+  oneof scope {
+    NamespacedPolicyData namespaced_policy = 5;
+    ClusterPolicyData cluster_policy = 6;
+  }
+}
+
+// Namespaced policy data.
+message NamespacedPolicyData {
+  // Kubernetes namespace for the policy.
+  string namespace = 1;
```
rlenglet: Instead of doing that, just add this as an `optional` field into `ConfiguredNetworkPolicyData` / `ConfiguredKubernetesObjectData`. So that mirrors the structure of `KubernetesObjectData` in the resources stream.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: This must support any kind of resources, not just network policies.
The only thing to change to e.g. configure `CiliumCIDRGroup`s should be to add one field into the `oneof` below.
So use a more generic name for this message.
Reusing the name `KubernetesObjectData` would highlight the symmetry with the resources stream.

```suggestion
// Configured resource data for snapshot or create/update operations.
message ConfiguredKubernetesObjectData {
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -485,20 +713,91 @@ message GetConfigurationUpdatesRequest {
...
+
+// Configured network policy data for snapshot or create/update operations.
+message ConfiguredNetworkPolicyData {
+  // Unique identifier for tracking this policy.
+  string id = 1;
+  // Kubernetes resource name for the policy.
+  string name = 2;
+  // Kubernetes labels to apply to the policy resource.
+  map<string, string> labels = 3;
+  // Kubernetes annotations to apply to the policy resource.
```
```suggestion
  // Kubernetes labels to apply to the resource.
  map<string, string> labels = 3;
  // Kubernetes annotations to apply to the resource.
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -485,20 +713,91 @@ message GetConfigurationUpdatesRequest {
...
+  }
+}
+
+// Namespaced policy data.
+message NamespacedPolicyData {
+  // Kubernetes namespace for the policy.
+  string namespace = 1;
+  // Policy specification based on the CNI type.
+  oneof policy {
+    KubernetesNetworkPolicyData k8s_network_policy = 2;
```
rlenglet: It's highly unlikely that we will ever configure vanilla k8s network policies.

```suggestion
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -485,20 +713,91 @@ message GetConfigurationUpdatesRequest {
...
+  }
+}
+
+// Cluster-wide policy data.
+message ClusterPolicyData {
+  // Policy specification based on the CNI type.
+  oneof policy {
+    KubernetesCiliumClusterwideNetworkPolicyData cilium_clusterwide_network_policy = 1;
+  }
+}
```
rlenglet: To recap my other comments on the configured data:

```suggestion
// Configured resource data for snapshot or create/update operations.
message ConfiguredKubernetesObjectData {
  // Unique identifier for tracking this resource.
  string id = 1;
  map<string, string> annotations = 2;
  string kind = 3;
  map<string, string> labels = 4;
  string name = 5;
  optional string namespace = 6;
  kind_specific scope {
    KubernetesCiliumNetworkPolicyData cilium_network_policy = 100;
    KubernetesCiliumClusterwideNetworkPolicyData cilium_clusterwide_network_policy = 101;
    // TODO: KubernetesCiliumCIDRGroupData cilium_cidr_group = 102;
  }
}
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -485,20 +713,91 @@ message GetConfigurationUpdatesRequest {
...
 
   oneof response {
-    // Updates the operator's configuration.
     Configuration update_configuration = 1;
+    // A network policy in the initial snapshot.
+    ConfiguredNetworkPolicyData network_policy_data = 2;
+    // Indicates that all network policies have been sent in the initial snapshot.
+    NetworkPolicySnapshotComplete network_policy_snapshot_complete = 3;
+    // A mutation to a network policy in the cluster.
+    NetworkPolicyMutation network_policy_mutation = 4;
```
rlenglet: We won't only send policy resources. We will also may be send `CiliumCIDRGroup` resources.
So the names of these fields should reflect that this stream supports configuring any kind of resources.
Use names similar to those in `SendKubernetesResourcesRequest`, for consistency:

```suggestion
    // Data about a Kubernetes resource to configure in the cluster.
    ConfiguredKubernetesObjectData resource_data = 2;
    // Indicates that all resource_data have been sent to report all the resources in the cluster in the initial snapshot.
    ConfiguredKubernetesResourceSnapshotComplete resource_snapshot_complete = 3;
    // A mutation to a Kubernetes resource in the cluster.
    KubernetesResourceMutation resource_mutation = 4;
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -485,20 +713,91 @@ message GetConfigurationUpdatesRequest {
...
+message NetworkPolicyMutation {
+  oneof mutation {
+    // A new network policy to create in the cluster.
+    ConfiguredNetworkPolicyData create_policy = 1;
+    // An existing network policy to update in the cluster.
+    ConfiguredNetworkPolicyData update_policy = 2;
+    // An existing network policy to delete from the cluster.
+    DeleteNetworkPolicy delete_policy = 3;
   }
 }
```
rlenglet: Rename those messages to not be specific to network policies, and to match those of the `SendKubernetesResourcesRequest` RPC. I'd just prefix those same names with "Configured".

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -45,6 +45,8 @@ message KubernetesObjectData {
     KubernetesNodeData node = 101;
     KubernetesServiceData service = 102;
     KubernetesNetworkPolicyData network_policy = 103;
+    KubernetesCiliumNetworkPolicyData cilium_network_policy = 104;
+    KubernetesCiliumClusterwideNetworkPolicyData cilium_clusterwide_network_policy = 105;
```
rlenglet, line 50:
```suggestion
    KubernetesCiliumClusterwideNetworkPolicyData cilium_clusterwide_network_policy = 105;
    // TODO: KubernetesCiliumCIDRGroupData cilium_cidr_group = 106;
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet, line 1: We must also support collecting and configuring `CiliumCIDRGroup` resources.
OK to do that in a separate PR later.

rlenglet (changes requested):

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: Do we need the `kind` field? It seems a bit redundant since the `kind_specific` field already tells the kind.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: Is it okay to keep the `DeleteConfiguredKubernetesObject` with only the ID field? For policies, this was okay since the in-memory map will store all the relevant fields.

rlenglet: Yes, that's ok!

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: Good catch. Right, you can remove `kind`.
The resources stream needs it, because for most kinds we won't send anything beyond the generic fields, and the kind_specific fields won't be set. But for the configuration stream, we will always set one of the kind_specific fields.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: added the `api_group_version` field to KubernetesObjectData

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: Added CiliumPolicyK8sServiceSelectorNamespace and CiliumPolicyK8sServiceNamespace.
Looking at the code, `k8sServiceSelector ` and `k8sService ` are mutually exclusive fields, but `egress.ToServices` can have a combination of both: https://github.com/cilium/cilium/blob/ad40b0b2aaf33ad6dc66b74466bc52418b1ebad4/pkg/policy/k8s/service.go#L304-L316

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: will do this in a separate PR

sungkyunchung: Updated `ConvertMetaObjectToMetadata` in `internal/controller/k8s_resources.go` and added a unit test for this

sungkyunchung: Added `CiliumCIDRGroup` related proto types. will add collecting in a separate PR

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: What is the purpose of collecting API group-version? API version is currently hard-coded in the Watcher and our plan for Cilium and future CNIs is also to hard-code the API version for each CNI.

Is this okay or do we actually want to support querying multiple API versions for each CNI? that would require larger changes to how we build the GVR for each resource.

rlenglet (approved):

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: As we discussed, this will be done in a future PR.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: We do want to query multiple API versions for each resource kind. Because users may be using multiple versions at once.
As we discussed, let's address this in another PR.

PR #405, opened 2026-04-28
Support collecting flows from AWS VPC CNI plugin

pavankumarinnamuri:


internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,197 @@
...
+			c.logger.Info("VPC CNI flow collector stopping")
+
+			return ctx.Err()
+		case <-ticker.C:
+			if err := c.pollAllPods(ctx); err != nil {
+				c.logger.Warn("Poll cycle failed", zap.Error(err))
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
pavankumarinnamuri: Fixed: Added client_test.go with 11 unit tests covering polling logic, log parsing, stale pod cleanup, and context cancellation.

internal/controller/stream/flows/vpccni/cluster_network_policy.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2024 Illumio, Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
pavankumarinnamuri: Fixed: Updated copyright header to 2026.

internal/controller/stream/flows/vpccni/client.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	logs, err := c.getLogsFromPod(ctx, pod.Name, sinceTime)
+	if err != nil {
+		return err
+	}
+
+	// Update last poll time
+	c.mu.Lock()
+	c.lastPollTime[pod.Name] = time.Now()
```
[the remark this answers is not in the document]
pavankumarinnamuri: Fixed: Now capture pollStart := time.Now() before GetLogs call to avoid gaps between request start and lastPollTime update.

internal/controller/collector/awsvpccni.go
```diff
@@ -0,0 +1,248 @@
...
+	// Check if the aws-eks-nodeagent container exists
+	if hasNodeagentContainer(pods.Items[0]) {
+		logger.Debug("VPC CNI with aws-eks-nodeagent detected")
+
+		return true
+	}
+
+	logger.Debug("aws-node pods found but aws-eks-nodeagent container not present")
+
+	return false
```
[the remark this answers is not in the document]
pavankumarinnamuri: Fixed: Removed Limit=1 and now check all pods for aws-eks-nodeagent container to handle rolling upgrades.

internal/controller/collector/awsvpccni.go
```diff
@@ -0,0 +1,252 @@
...
+	SrcPort   uint32 `json:"Src Port"`
+	DestIP    string `json:"Dest IP"`
+	DestPort  uint32 `json:"Dest Port"`
+	Proto     string `json:"Proto"`   // TCP, UDP, ICMP, SCTP, UNKNOWN
+	Verdict   string `json:"Verdict"` // ACCEPT, DENY, EXPIRED/DELETED
+}
+
+// flowMsgPattern extracts flow data from the embedded msg string in v1.2.2+ format.
+// Example: "Flow Info: Src IP: 10.0.1.28 Src Port: 55484 Dest IP: 10.0.1.132 Dest Port: 80 Proto TCP Verdict ACCEPT Direction egress".
+var flowMsgPattern = regexp.MustCompile(
```
ganesh-talla, line 70: Include Flow Info:

internal/controller/collector/awsvpccni.go
```diff
@@ -0,0 +1,252 @@
...
+// and new format (v1.2.2+) with embedded msg string.
+func ParseVPCCNIFlowLog(line string) (*pb.FiveTupleFlow, error) {
+	var log VPCCNIFlowLog
+
+	if err := json.Unmarshal([]byte(line), &log); err != nil {
+		return nil, ErrVPCCNINotFlowLog
+	}
+
+	// Check if this is a flow log (must have "Flow Info" in message)
+	if !strings.Contains(log.Message, "Flow Info") {
```
ganesh-talla, line 114: You are already do this check below if you include Flow Info: in the regex pattern

internal/controller/collector/vpccni.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	var (
+		srcIP, destIP, proto string
+		srcPort, destPort    uint32
+	)
+
+	switch {
+	case isOldFormat:
+		// Old format: fields are in separate JSON keys
```
ganesh-talla: make this into a new function so we can unit test this separately deterministically

internal/controller/collector/vpccni.go
```diff
@@ -0,0 +1,252 @@
...
+			return nil, ErrVPCCNIInvalidLog
+		}
+
+		srcIP = log.SrcIP
+		srcPort = log.SrcPort
+		destIP = log.DestIP
+		destPort = log.DestPort
+		proto = log.Proto
+	case isNewFormat:
+		// New format (v1.2.2+): parse from embedded msg string
```
ganesh-talla: same here also

internal/controller/collector/vpccni.go
```diff
@@ -0,0 +1,252 @@
...
+		if !ok {
+			return nil, ErrVPCCNIInvalidLog
+		}
+	default:
+		// Should not reach here due to earlier check, but be defensive
+		return nil, ErrVPCCNINotFlowLog
+	}
+
+	// Validate required fields
+	if srcIP == "" || destIP == "" {
```
ganesh-talla: I see so many error cases here why can't you delegate this to the above refactored functions so this looks clean.

internal/controller/collector/vpccni.go
```diff
@@ -0,0 +1,252 @@
...
+		protoStr = "tcp"
+	}
+
+	layer4Message, err := CreateLayer4Message(protoStr, srcPort, destPort, ipVersion)
+	if err != nil {
+		return nil, ErrVPCCNIInvalidProtocol
+	}
+
+	// Parse timestamp or use current time
+	ts := timestamppb.Now()
```
ganesh-talla: will this not cause an wrong ordering of flows if a flow is there on a node and we read logs(flows) every 5 sec. If a timestamp is not there can't we just drop this?

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,201 @@
...
+	return buf.String(), nil
+}
+
+// parseLogs parses log lines and caches flows.
+func (c *vpccniClient) parseLogs(ctx context.Context, logs string, podName string) {
+	if logs == "" {
+		return
+	}
+
+	lines := strings.Split(logs, "\n")
```
ganesh-talla: Can't you filter this at stream level?
My only concern is if the flows are very high how will the memory be handled?

Are you aware what % of logs in k8s:Node that correspond to Flows?

ganesh-talla (changes requested):
LGTM on the overall design, but I have concerns about correctness and memory safety under load that should be addressed before merge.

## Stress testing needed

Can we do a stress test monitoring operator memory? The operator polls ALL `aws-node` pods sequentially, buffering each pod's entire log response into a `bytes.Buffer` in memory with no `LimitBytes` cap on `PodLogOptions`. On EKS clusters with many nodes and high connection rates, once flow logging is enabled (`ClusterNetworkPolicy` created), ~90-99% of `aws-eks-nodeagent` output will be flow logs. The kubelet's default `containerLogMaxSize` is 10 MiB per log file, and the K8s log API returns from the active file — so each poll could buffer up to ~10 MiB per pod (more during burst, since kubelet rotation can lag under heavy writes per [kubernetes#110630](https://github.com/kubernetes/kubernetes/issues/110630)). Multiply by N nodes and this could cause significant memory spikes.

**Suggestion:** Add `LimitBytes` to `PodLogOptions` as a hard safety cap, and consider streaming line-by-line with `bufio.Scanner` instead of buffering the full response.

## Log reprocessing between consecutive polls

There is no guarantee against reprocessing the same logs across consecutive polls:

1. **`SinceTime` has second-level granularity.** `metav1.Time` serializes as RFC3339 (seconds). If `pollStart = 12:00:05.300`, the K8s API sees `SinceTime = 12:00:05`. The next poll re-fetches everything from `05.000–05.999` that was already processed. With a 5s interval, ~20% of logs get reprocessed every cycle.

2. **`SinceTime` is inclusive (`>=`).** Even with perfect precision, logs at the exact boundary timestamp appear in both polls.

3. **Clock skew.** `time.Now()` runs on the operator pod; the `"ts"` field in flow logs comes from the node's container runtime. Any drift causes gaps (missed flows) or overlaps (duplicates).

4. **Reconnection resets `lastPollTime`.** When `ManageStream` reconnects after failure, a new `vpccniClient` is created with an empty `lastPollTime` map. The first poll uses `time.Now().Add(-pollInterval)`, which may overlap with logs already processed before the failure.

The flow cache deduplicates on 5-tuple (no timestamp in key), but this only helps if the original flow is still in cache. If it was already evicted via `activeTimeout`, the duplicate gets sent upstream again. Also, `IncrementFlowsReceived()` is called before the cache dedup check, so metrics are inflated regardless.

**Suggestion:** Track the latest `"ts"` value from actual log content per pod (not `time.Now()`), and use `max(seen_ts) + 1s` as `SinceTime` for the next poll. This eliminates clock skew and the second-level precision overlap.

internal/controller/collector/vpccni.go
pavankumarinnamuri: It should drop; will change.

internal/controller/collector/vpccni.go
pavankumarinnamuri: These are validation errors (IP, protocol, timestamp) shared by both formats - moving them would duplicate code.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -459,6 +459,8 @@ enum FlowCollector {
   FLOW_COLLECTOR_FALCO = 3;
   // Indicates that OVN-Kubernetes is deployed and configured for collecting network flows.
   FLOW_COLLECTOR_OVNK = 4;
+  // Indicates that AWS VPC CNI is deployed with Network Policy Agent for collecting network flows.
+  FLOW_COLLECTOR_VPC_CNI = 5;
```
```suggestion
  // Indicates that AWS VPC CNI is deployed with Network Policy Agent for collecting network flows.
  FLOW_COLLECTOR_AWS_VPC_CNI = 5;
```

cloud-operator/templates/clusterrole.yaml
```diff
@@ -10,6 +10,9 @@ rules:
 - apiGroups: [""]
   resources: ["configmaps", "endpoints", "namespaces", "nodes", "pods", "replicationcontrollers", "serviceaccounts", "services"]
   verbs: ["get", "list", "watch"]
+- apiGroups: [""]
+  resources: ["pods/log"]
+  verbs: ["get"]
```
rlenglet: This is too broad. Revert.

```suggestion
```

We must only allow reading logs from pods in the `kube-system`. I don't think it's possible to specify a wildcard for the resource name (i.e. the pod name) to deal with the fact that the pods have random suffixes. So unfortunately, we'll have to allow reading logs from _all_ the pods in `kube-system`.
You can't do that by just adding this into the `ClusterRole`.
You must define a `Role` and `RoleBinding` in the `kube-system` namespace. It should be very similar to the `hubble-client-certs-reader` `Role` and `cloud-operator-read-hubble-client-certs` `RoleBinding` that we already define:
https://github.com/illumio/cloud-operator/blob/4cbbadec02913b6d05f525b711051bc1d9c6ef7c/cloud-operator/templates/role.yaml#L13-L24
https://github.com/illumio/cloud-operator/blob/4cbbadec02913b6d05f525b711051bc1d9c6ef7c/cloud-operator/templates/rolebinding.yaml#L16-L29

Just follow the same pattern. The only difference is that you won't be able to specify the `resourceNames` in this case.

cmd/main.go
```diff
@@ -123,6 +123,8 @@ func main() {
 	bindEnv(logger, "tls_skip_verify", "TLS_SKIP_VERIFY")
 	bindEnv(logger, "token_endpoint", "TOKEN_ENDPOINT")
 	bindEnv(logger, "verbose_debugging", "VERBOSE_DEBUGGING")
+	bindEnv(logger, "grpc_internal_logging", "GRPC_INTERNAL_LOGGING")
```
rlenglet: Delete this line. It was moved up in a previous PR to keep the variables in lexicographic order.

```suggestion
```

cmd/main.go
```diff
@@ -123,6 +123,8 @@ func main() {
...
+	bindEnv(logger, "vpc_cni_poll_interval", "VPC_CNI_POLL_INTERVAL")
```
rlenglet:
```suggestion
	bindEnv(logger, "aws_vpc_cni_logs_polling_interval", "AWS_VPC_CNI_LOGS_POLLING_INTERVAL")
```

cmd/main.go
```diff
@@ -146,6 +148,7 @@ func main() {
 	viper.SetDefault("tls_skip_verify", false)
 	viper.SetDefault("token_endpoint", "https://dev.cloud.ilabs.io/api/v1/k8s_cluster/authenticate")
 	viper.SetDefault("verbose_debugging", false)
+	viper.SetDefault("vpc_cni_poll_interval", "5s")
```
rlenglet: The hardcoded default polling interval in `kubectl` is `1s`. Use the same.
https://github.com/kubernetes/kubectl/blob/9293e393fe69a6c7c4b0852ff114691a4d97f30c/pkg/cmd/logs/logs.go#L446

```suggestion
	viper.SetDefault("aws_vpc_cni_logs_polling_interval", "1s")
```

cmd/main.go
```diff
@@ -188,6 +191,7 @@ func main() {
 		zap.Bool("tls_skip_verify", envConfig.TlsSkipVerify),
 		zap.String("token_endpoint", envConfig.TokenEndpoint),
 		zap.Bool("verbose_debugging", viper.GetBool("verbose_debugging")),
+		zap.Duration("vpc_cni_poll_interval", viper.GetDuration("vpc_cni_poll_interval")),
```
rlenglet:
```suggestion
		zap.Duration("aws_vpc_cni_logs_polling_interval", viper.GetDuration("aws_vpc_cni_logs_polling_interval")),
```

cmd/main.go
```diff
@@ -242,6 +246,7 @@ func main() {
 		IPFIXCollectorPort: viper.GetString("ipfix_collector_port"),
 		OVNKNamespace:      viper.GetString("ovnk_namespace"),
 		TlsAuthProps:       tlsAuthProps,
+		VPCCNIPollInterval: viper.GetDuration("vpc_cni_poll_interval"),
```
rlenglet:
```suggestion
		AWSVPCCNIPollingInterval: viper.GetDuration("aws_vpc_cni_logs_polling_interval"),
```

internal/controller/collector/vpccni.go
```diff
@@ -0,0 +1,250 @@
...
+// {"level":"info","ts":"2024-09-23T12:36:53.562Z","logger":"ebpf-client",
+//
+//	"msg":"Flow Info: ","Src IP":"10.0.141.167","Src Port":39197,
+//	"Dest IP":"172.20.0.10","Dest Port":53,"Proto":"TCP","Verdict":"ACCEPT"}
+//
+// New format (v1.2.2+):
+// {"level":"debug","ts":"2026-04-13T21:18:46.888Z","caller":"runtime/asm_amd64.s:1700",
+//
+//	"msg":"Flow Info: Src IP: 10.0.1.28 Src Port: 55484 Dest IP: 10.0.1.132 Dest Port: 80 Proto TCP Verdict ACCEPT Direction egress"}
+type VPCCNIFlowLog struct {
```
rlenglet:
```suggestion
// AWSVPCCNIFlowLog represents the flow log format from aws-eks-nodeagent.
//
// Old format (v1.0.x - v1.2.1):
// {"level":"info","ts":"2024-09-23T12:36:53.562Z","logger":"ebpf-client",
//
//	"msg":"Flow Info: ","Src IP":"10.0.141.167","Src Port":39197,
//	"Dest IP":"172.20.0.10","Dest Port":53,"Proto":"TCP","Verdict":"ACCEPT"}
//
// New format (v1.2.2+):
// {"level":"debug","ts":"2026-04-13T21:18:46.888Z","caller":"runtime/asm_amd64.s:1700",
//
//	"msg":"Flow Info: Src IP: 10.0.1.28 Src Port: 55484 Dest IP: 10.0.1.132 Dest Port: 80 Proto TCP Verdict ACCEPT Direction egress"}
type AWSVPCCNIFlowLog struct {
```

internal/controller/collector/awsvpccni.go
rlenglet, line 1: Rename this file into `awsvpccni.go`.
Same for the `_test.go` file.

internal/controller/collector/vpccni.go
```diff
@@ -0,0 +1,250 @@
...
+
+// isIPv6 checks if the given address is an IPv6 address.
+func isIPv6(addr string) bool {
+	for _, c := range addr {
+		if c == ':' {
+			return true
+		}
+	}
+
+	return false
```
rlenglet:
```suggestion
	return strings.ContainsRune(addr, ':')
```

internal/controller/collector/vpccni.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+// IsVPCCNIAvailable checks if AWS VPC CNI with flow logging is available in the cluster.
+// It looks for aws-node pods with the aws-eks-nodeagent container.
+// Checks multiple pods to handle rolling upgrades where some pods may not have nodeagent yet.
+func IsVPCCNIAvailable(ctx context.Context, logger *zap.Logger, k8sClient kubernetes.Interface) bool {
```
rlenglet: Please rename "VPNCNI" into "AWSVPCCNI" everywhere.

internal/controller/collector/vpccni.go
```diff
@@ -0,0 +1,250 @@
...
+	if len(pods.Items) == 0 {
+		logger.Debug("No aws-node pods found")
+
+		return false
+	}
+
+	// Check if any pod has the aws-eks-nodeagent container
+	for i := range pods.Items {
+		if hasNodeagentContainer(pods.Items[i]) {
+			logger.Debug("VPC CNI with aws-eks-nodeagent detected",
```
rlenglet:
```suggestion
			logger.Debug("AWS VPC CNI with aws-eks-nodeagent detected",
```

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,201 @@
...
+		return "", err
+	}
+	defer stream.Close() //nolint:errcheck
+
+	buf := new(bytes.Buffer)
+	if _, err := io.Copy(buf, stream); err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
```
rlenglet: This logic doesn't seem right. Use HTTP streaming instead of just polling one buffer at a time, like kubectl does:
https://github.com/kubernetes/kubectl/blob/9293e393fe69a6c7c4b0852ff114691a4d97f30c/pkg/cmd/logs/logs.go#L503-L516
Also, using `r.ReadBytes('\n')` should save you a lot of headaches for parsing lines.

Beware that you'll have to create one goroutine to consume logs from each pod, since this streaming will be blocking. And you'll have to create/stop goroutines dynamically as pods are created and deleted.

rlenglet (changes requested):

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,201 @@
...
+		return err
+	}
+
+	// Update last poll time to poll start to avoid missing logs
+	c.mu.Lock()
+	c.lastPollTime[pod.Name] = pollStart
+	c.mu.Unlock()
+
+	// Parse and cache flows
+	c.parseLogs(ctx, logs, pod.Name)
```
rlenglet: This is not sufficient.
Flows must also be ordered by start time across all pods.
Flows coming from a single pod are ordered, since we can assume that their timestamps are monotonically increasing.
But that's not true for flows from different pods. If a node hasn't seen traffic for 5 minutes, the last flow you would parse from the AWS VPC CNI pod on that node may be say several minutes behind the latest flow from another pod.
And that can mess up with flow aggregation in the backend.

So you must do the same as Hubble Relay: have an in-memory cache of flows coming from all pods, to reorder them before forwarding them, and reject flows that are too old compared to the latest flow start time received.

Maybe that logic could be implemented in the flow sink's `CacheFlow` method.

rlenglet (changes requested):

internal/controller/stream/flows/awsvpccni/client.go
pavankumarinnamuri: We decided to keep polling for now. Streaming would require managing one goroutine per pod with dynamic creation/deletion as pods come and go. The polling approach is simpler and the FlowCache now 
  handles cross-pod flow ordering with sorted insertion.

internal/controller/stream/flows/awsvpccni/client.go
pavankumarinnamuri: Addressed in FlowCache.addFlowToCache() - flows from all pods now inserted in timestamp-sorted order. Existing activeTimeout handles stale flow rejection.

cloud-operator/templates/role.yaml
```diff
@@ -22,6 +22,17 @@ rules:
   resources: ["secrets"]
   resourceNames: ["hubble-relay-client-certs"] # Name of the Secret
   verbs: ["get"]
+---
+# Role for reading AWS VPC CNI pod logs in kube-system namespace
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  namespace: kube-system
+  name: aws-vpc-cni-logs-reader
```
rlenglet: Prefix with "cloud-operator-" so anyone can understand what owns this.

cloud-operator/templates/rolebinding.yaml
```diff
@@ -27,6 +27,21 @@ roleRef:
...
+metadata:
+  namespace: kube-system
+  name: cloud-operator-read-aws-vpc-cni-logs
+subjects:
+- kind: ServiceAccount
+  name: {{ .Values.serviceAccount.name }}
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: Role
+  name: aws-vpc-cni-logs-reader
```
rlenglet: Update here accordingly.

internal/controller/collector/awsvpccni.go
```diff
@@ -0,0 +1,244 @@
...
+
+	pb "github.com/illumio/cloud-operator/api/illumio/cloud/k8sclustersync/v1"
+)
+
+const (
+	// AWSNodeLabel is the label selector for aws-node pods.
+	AWSNodeLabel = "k8s-app=aws-node"
+	// AWSNodeNamespace is the namespace where aws-node pods run.
+	AWSNodeNamespace = "kube-system"
+	// AWSEksNodeagentContainer is the container name that has flow logs.
```
rlenglet: Nit.

```suggestion
	// AWSEksNodeagentContainer is the name of the container that has flow logs.
```

internal/controller/collector/awsvpccni.go
```diff
@@ -0,0 +1,244 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+
+	pb "github.com/illumio/cloud-operator/api/illumio/cloud/k8sclustersync/v1"
+)
+
+const (
+	// AWSNodeLabel is the label selector for aws-node pods.
+	AWSNodeLabel = "k8s-app=aws-node"
+	// AWSNodeNamespace is the namespace where aws-node pods run.
```
rlenglet: Nit.
```suggestion
	// AWSNodeNamespace is the namespace of the aws-node pods.
```

internal/controller/collector/awsvpccni.go
```diff
@@ -0,0 +1,244 @@
...
+	ErrAWSVPCCNINotFlowLog       = errors.New("log line is not a flow log")
+	ErrAWSVPCCNIInvalidProtocol  = errors.New("unsupported protocol in AWS VPC CNI flow log")
+	ErrAWSVPCCNIInvalidTimestamp = errors.New("invalid or missing timestamp in AWS VPC CNI flow log")
+)
+
+// AWSVPCCNIFlowLog represents the flow log format from aws-eks-nodeagent.
+//
+// Old format (v1.0.x - v1.2.1):
+// {"level":"info","ts":"2024-09-23T12:36:53.562Z","logger":"ebpf-client",
+//
```
rlenglet, line 44:
```suggestion
```

internal/controller/collector/awsvpccni.go
```diff
@@ -0,0 +1,244 @@
...
+//	"msg":"Flow Info: ","Src IP":"10.0.141.167","Src Port":39197,
+//	"Dest IP":"172.20.0.10","Dest Port":53,"Proto":"TCP","Verdict":"ACCEPT"}
+//
+// New format (v1.2.2+):
+// {"level":"debug","ts":"2026-04-13T21:18:46.888Z","caller":"runtime/asm_amd64.s:1700",
+//
```
rlenglet, line 50:
```suggestion
```

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,201 @@
...
+	pollInterval time.Duration
+
+	mu           sync.Mutex
+	lastPollTime map[string]time.Time
+}
+
+// Run collects flows from VPC CNI by polling aws-node pod logs.
+func (c *vpccniClient) Run(ctx context.Context) error {
+	c.logger.Info("Starting VPC CNI flow collector",
+		zap.Duration("pollInterval", c.pollInterval))
```
rlenglet: Keep the case of field names consistent.

```suggestion
		zap.Duration("poll_interval", c.pollInterval))
```

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,201 @@
...
+		return "", err
+	}
+	defer stream.Close() //nolint:errcheck
+
+	buf := new(bytes.Buffer)
+	if _, err := io.Copy(buf, stream); err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
```
rlenglet: Instead of returning a string just to re-parse it later, it would be more efficient to just return the `[]bytes` and to use the `bufio` package to implement parsing more efficiently, e.g. using [`ScanLines`](https://pkg.go.dev/bufio#ScanLines), which does zero copies.

```suggestion
	return buf.Bytes(), nil
```

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// parseLogs parses log lines and caches flows.
+func (c *vpccniClient) parseLogs(ctx context.Context, logs string, podName string) {
```
rlenglet: Instead of passing the podName just for logging, it would be cleaner to just pass a logger as a parameter with the `pod` field already set.

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,201 @@
...
+
+		c.flowSink.IncrementFlowsReceived()
+
+		flowCount++
+	}
+
+	if flowCount > 0 {
+		c.logger.Debug("Parsed flows from pod",
+			zap.String("pod", podName),
+			zap.Int("flowCount", flowCount))
```
rlenglet:
```suggestion
			zap.Int("flow_count", flowCount))
```

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,201 @@
...
+func (c *vpccniClient) Run(ctx context.Context) error {
+	c.logger.Info("Starting VPC CNI flow collector",
+		zap.Duration("pollInterval", c.pollInterval))
+
+	ticker := time.NewTicker(c.pollInterval)
+	defer ticker.Stop()
+
+	// Do an initial poll immediately
+	if err := c.pollAllPods(ctx); err != nil {
+		c.logger.Warn("Initial poll failed", zap.Error(err))
```
rlenglet: All log messages in this file must make clear that they are about AWS VPC CNI. Without that context, this message doesn't make much sense.
Same for all the other log messages in this file.

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,201 @@
...
+	if flowCount > 0 {
+		c.logger.Debug("Parsed flows from pod",
+			zap.String("pod", podName),
+			zap.Int("flowCount", flowCount))
+	}
+}
+
+// cleanupStalePods removes entries from lastPollTime for pods that no longer exist.
+func (c *vpccniClient) cleanupStalePods(pods *corev1.PodList) {
+	activePods := make(map[string]bool)
```
rlenglet:
```suggestion
	activePods := make(map[string]bool, len(pods.Items))
```

internal/controller/stream/flows/awsvpccni/cluster_network_policy.go
```diff
@@ -0,0 +1,113 @@
...
+	Group:    "networking.k8s.aws",
+	Version:  "v1alpha1",
+	Resource: "clusternetworkpolicies",
+}
+
+// IsCRDAvailable checks if the ClusterNetworkPolicy CRD is registered in the cluster.
+func IsCRDAvailable(logger *zap.Logger, discoveryClient discovery.DiscoveryInterface) bool {
+	_, err := discoveryClient.ServerResourcesForGroupVersion(ClusterNetworkPolicyAPIVersion)
+	if err != nil {
+		logger.Debug("ClusterNetworkPolicy CRD not available", zap.Error(err))
```
rlenglet:
```suggestion
		logger.Warn("ClusterNetworkPolicy CRD not available", zap.Error(err))
```

internal/controller/stream/flows/awsvpccni/cluster_network_policy.go
```diff
@@ -0,0 +1,113 @@
...
+	logger.Debug("ClusterNetworkPolicy CRD is available")
+
+	return true
+}
+
+// EnsureFlowLoggingPolicy creates the ClusterNetworkPolicy for comprehensive flow logging.
+// If the policy already exists, it returns nil (idempotent).
+func EnsureFlowLoggingPolicy(ctx context.Context, logger *zap.Logger, dynamicClient dynamic.Interface) error {
+	policy := &unstructured.Unstructured{
+		Object: map[string]any{
```
rlenglet, line 49: You should set the owner reference to refer to the `Deployment` that owns this pod, so that this policy gets deleted automatically whenever cloud-operator is uninstalled. Otherwise, we're leaving this resource behind.

internal/controller/stream/flows/awsvpccni/cluster_network_policy.go
```diff
@@ -0,0 +1,113 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+
+	_, err := dynamicClient.Resource(clusterNetworkPolicyGVR).Create(ctx, policy, metav1.CreateOptions{})
+	if err != nil {
+		// Policy is cluster-scoped and persists across operator restarts.
```
rlenglet:
```suggestion
		// Policy is cluster-scoped and persists across restarts of the cloud-operator pod.
```

internal/controller/stream/flows/awsvpccni/cluster_network_policy.go
```diff
@@ -0,0 +1,113 @@
...
+
+			return nil
+		}
+
+		logger.Error("Failed to create ClusterNetworkPolicy", zap.Error(err))
+
+		return err
+	}
+
+	logger.Info("Created ClusterNetworkPolicy for comprehensive flow logging", zap.String("name", ClusterNetworkPolicyName))
```
rlenglet: In every log message in this file, make clear that it is about AWS VPC CNI.

internal/controller/stream/flows/detect.go
```diff
@@ -23,7 +25,7 @@ var _ stream.StreamClientFactory = (*FlowCollectorStreamFactory)(nil)
 // This allows flow collectors to be managed by the stream manager like other streams.
 type FlowCollectorStreamFactory struct {
 	Factory       CollectorFactory
-	CollectorName string // e.g., "Cilium", "OVN-K", "Falco"
+	CollectorName string // e.g., "Cilium", "OVN-K", "Falco", "VPC-CNI"
```
rlenglet: Be consistent about that name, e.g. with https://github.com/illumio/cloud-operator/pull/405/changes#r3352864267 and https://github.com/illumio/cloud-operator/pull/405/changes#diff-66f5dc1744cea6a0560f6c893064fe1c386ec47fac031abaa8367c1281af29fbR463.
```suggestion
	CollectorName string // e.g., "Cilium", "OVN-K", "Falco", "AWS-VPC-CNI"
```

internal/controller/stream/flows/detect.go
```diff
@@ -117,6 +119,35 @@ func DetectFlowCollector(ctx context.Context, config CollectorConfig) (pb.FlowCo
...
+	if collector.IsAWSVPCCNIAvailable(ctx, config.Logger, clientset) {
+		config.Logger.Info("Using AWS VPC CNI flow collector")
+
+		// Check CRD and create ClusterNetworkPolicy for comprehensive flow logging
+		discoveryClient := config.K8sClient.GetDiscoveryClient()
+		dynamicClient := config.K8sClient.GetDynamicClient()
+
+		if awsvpccni.IsCRDAvailable(config.Logger, discoveryClient) {
+			if err := awsvpccni.EnsureFlowLoggingPolicy(ctx, config.Logger, dynamicClient); err != nil {
+				config.Logger.Warn("Failed to create ClusterNetworkPolicy, flow logging may be limited",
```
rlenglet: Clarify in each log message that it is about AWS VPC CNI.

internal/controller/stream/flows/detect.go
```diff
@@ -117,6 +119,35 @@ func DetectFlowCollector(ctx context.Context, config CollectorConfig) (pb.FlowCo
...
+					zap.Error(err))
+			}
+		} else {
+			config.Logger.Warn("ClusterNetworkPolicy CRD not found - enable network policy in VPC CNI addon for comprehensive flow logging")
```
rlenglet:
```suggestion
			config.Logger.Warn("ClusterNetworkPolicy CRD not found - enable network policy in AWS VPC CNI addon for comprehensive flow logging")
```

rlenglet (changes requested):
Looks good!
I just have a few relatively minor comments, mostly about improving logging.

pavankumarinnamuri:
@rlenglet Re: your comment about setting the owner reference to the Deployment for the ClusterNetworkPolicy —

One thing I noticed while implementing this: ClusterNetworkPolicy is cluster-scoped, but the Deployment is namespace-scoped. Kubernetes garbage collection doesn't support cross-scope owner references, so setting the Deployment as owner won't actually auto-delete the policy on uninstall.

I went ahead and added the owner reference lookup, but wanted to flag that it won't trigger automatic cleanup. The policy does have `app.kubernetes.io/managed-by: illumio-cloud-operator` labels, so it can be cleaned up via label selector.

Should we keep the owner reference as-is, drop it and rely on labels only, or handle cleanup differently?

rlenglet:
> Should we keep the owner reference as-is, drop it and rely on labels only, or handle cleanup differently?

Oh yeah, sorry, I had forgotten about the issue with clusterwide resources owner references. Let's just forget about it, then.
Keep the labels only, as you already do.
We'll have to publish documentation with commands for users to run for full uninstallation after `helm uninstall`.

rlenglet (dismissed):
This is great, thank you!

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,198 @@
...
+	})
+
+	stream, err := req.Stream(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer stream.Close() //nolint:errcheck
+
+	buf := new(bytes.Buffer)
+	if _, err := io.Copy(buf, stream); err != nil {
```
ganesh-talla, line 137: Will this not store and copy all logs in memory again?
Why can't we do this check in parseLogs? down the stream without copying 

something like just return stream and process stream in parseLogs() like this 
```
scanner := bufio.NewScanner(stream) // stream directly, no intermediate buffer
        flowCount := 0
        for scanner.Scan() {
                line := scanner.Text()
                if len(line) == 0 {
                        continue
                } 
```

internal/controller/stream/flows/awsvpccni/client.go
```diff
@@ -0,0 +1,198 @@
...
+	pollStart := time.Now()
+
+	logs, err := c.getLogsFromPod(ctx, pod.Name, sinceTime)
+	if err != nil {
+		return err
+	}
+
+	// Update last poll time to poll start to avoid missing logs
+	c.mu.Lock()
+	c.lastPollTime[pod.Name] = pollStart
```
ganesh-talla, line 113: In last poll time you are just storing the time last poll started. 
@rlenglet Just wanted to confirm that will flows be in sync with operator time and  and written to logs.

ganesh-talla (dismissed):
LGTM, just some minor nits.

Re: memory concern from my earlier review — I've dug into this more and the risk is bounded. Kubelet enforces a default containerLogMaxSize of 10 MiB per log file, and retains containerLogMaxFiles = 5 files by default. The K8s pod log API only reads from the active (current) log file, not the rotated ones. So each poll is capped at ~10 MiB per pod at most.

That said, it would still be better to stream line-by-line from the HTTP response instead of buffering the full response into bytes.Buffer. You're already using bufio.Scanner in parseLogs(), but on a second copy of the data. 

On log rotation and skipped logs: If logs rotate between polls, lines from the rotated file are lost since the pod log API only serves the current file. With a 1s poll interval this is unlikely under normal load, but under burst writes kubelet rotation can lag (kubernetes#110630), so a few flow logs could theoretically be missed during rotation. 

Is this is an acceptable trade-off?

CC: @rlenglet @pavankumarinnamuri

internal/controller/stream/flows/awsvpccni/client.go
rlenglet: I agree. And I already left that comment too.
https://github.com/illumio/cloud-operator/pull/405#discussion_r3185423726
The trade off is that it requires creating one goroutine per pod/node.
I would prefer to do that personally.

pavankumarinnamuri: Agreed. I'll keep the current approach for now and add a TODO to track the per-pod goroutine streaming as a future improvement. Also, all the backend changes will be in prod today. One question on the operator side: should I cut a beta version, or go straight to a normal release? Wanted your call before tagging anything.

ganesh-talla (approved):
LGTM!
Great work
Some things need to be tested and discussed.

internal/controller/stream/flows/awsvpccni/client.go
ganesh-talla: @pavankumarinnamuri We discussed this on board last week.
Can you create a ticket to discuss it with Romain.

PR #406, opened 2026-04-28
Add api_group and api_version to k8s resource metadata and support multiple API versions per resource

sungkyunchung:
The resource watcher previously tracked only the API group, and the version was hardcoded to `v1`. To correctly support resources across all API groups and versions, we now track the preferred version from discovery to build the GVR for list and watch, and include the API group and version from the response object when streaming resource metadata.

- **Add `api_group` and `api_version` to resource metadata**: `KubernetesObjectData` gets two new fields derived from the unstructured response — `api_group` (e.g. `"apps"`) and `api_version` (e.g. `"v1"`).
- **Propagate discovered version through the watcher**: discovery now stores `ResourceInfo{Group, Version}` using `group.PreferredVersion`, replacing the hardcoded `v1`.


internal/controller/stream/resources/resources.go
```diff
@@ -50,28 +56,27 @@ func buildResourceApiGroupMap(resources []string, clientset kubernetes.Interface
...
-							zap.String("group", group.Name),
-							zap.String("resource", resource.Name),
-						)
+			continue
+		}
 
-						continue
-					}
+		// Query only the preferred version rather than iterating all group.Versions.
+		// All resources in our resource list always present in the preferred version,
```
rlenglet, line 66: So, if a user is using multiple versions of a group, do we only collect resources using the latest version? So we're not ingesting resources using the other versions?
That doesn't sound right.
For resources for which we only collect metadata, we need to collect all resources using all available versions in the same group.

Only for the few resource kinds for which we collect additional data, e.g. Node, Pod, Deployment, etc. do we need to check the version.

Does that makes sense?

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -41,6 +41,8 @@ message KubernetesObjectData {
   repeated KubernetesOwnerReference owner_references = 7;
   string resource_version = 8;
   string uid = 9;
+  // Full Kubernetes apiVersion string, e.g. "apps/v1" or "v1" for core resources.
+  string api_version = 10;
```
rlenglet: This is really group + version, right?
In that case, why not split this into 2 fields for consistency?
Or at least name this correctly to reflect what this really is, e.g. `group_version`.

rlenglet (changes requested):

internal/controller/stream/resources/resources.go
sungkyunchung: We previously hard‑coded the API version to `v1`. The preferred version for the resources we currently list is also `v1`, so this change will not result in any resources being missed during ingestion.
Using the preferred version is also helpful for future CRD ingestion (v2 for Cilium and v1alpha for AWS EKS). I think we don't want to hard-code this for each CNI?

Per the Kubernetes API documentation, objects created using older API versions remain accessible via the preferred API version.

> API resources are distinguished by their API group, resource type, namespace (for namespaced resources), and name. The API server handles the conversion between API versions transparently: all the different versions are actually representations of the same persisted data. The API server may serve the same underlying data through multiple API versions.

> For example, suppose there are two API versions, v1 and v1beta1, for the same resource. If you originally created an object using the v1beta1 version of its API, you can later read, update, or delete that object using either the v1beta1 or the v1 API version, until the v1beta1 version is deprecated and removed. At that point you can continue accessing and modifying the object using the v1 API.

https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-groups-and-versioning

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -41,6 +41,8 @@ message KubernetesObjectData {
   repeated KubernetesOwnerReference owner_references = 7;
   string resource_version = 8;
   string uid = 9;
+  string api_group = 10;
+  string api_version = 11;
```
rlenglet, line 45: Also add the same two fields into `ConfiguredKubernetesObjectData`, so we know what group/version to use for a configured resource. Otherwise, how could we guess?

internal/controller/stream/resources/resources.go
rlenglet: OK.

rlenglet (approved):
Nice, looks good!
I commented re: whether we need to also pass the group+version in `ConfiguredKubernetesObjectData`, but I'm not sure if that's necessary. If not, just close the discussion and this is good to go.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
sungkyunchung: Since we already build a ResourceInfo map at startup via discovery, we can share that with the configured resource enforcer so it can look up group/version from there by resource name.

PR #407, opened 2026-04-29
Cache in memory the configuration desired state

byjustina:
- Implement an in-memory cache to store ConfiguredKubernetesObjectData objects received from CloudSecure via the  GetConfigurationUpdates gRPC stream
  - Enable tracking of desired state for future reconciliation

### DIFF
--- a/api/illumio/cloud/k8sclustersync/v1/k8s_info.pb.go
+++ b/api/illumio/cloud/k8sclustersync/v1/k8s_info.pb.go
@@ -9,13 +9,14 @@
 package k8sclustersyncv1
 
 import (
+	reflect "reflect"
+	sync "sync"
+	unsafe "unsafe"
+
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
 	wrapperspb "google.golang.org/protobuf/types/known/wrapperspb"
-	reflect "reflect"
-	sync "sync"
-	unsafe "unsafe"
 )
 
 const (
--- a/api/illumio/cloud/k8sclustersync/v1/k8s_info.pb.go
+++ b/api/illumio/cloud/k8sclustersync/v1/k8s_info.pb.go
@@ -9,14 +9,13 @@
 package k8sclustersyncv1
 
 import (
-	reflect "reflect"
-	sync "sync"
-	unsafe "unsafe"
-
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
 	wrapperspb "google.golang.org/protobuf/types/known/wrapperspb"
+	reflect "reflect"
+	sync "sync"
+	unsafe "unsafe"
 )
 
 const (
--- a/internal/controller/stream/config/cache/cache.go
+++ b/internal/controller/stream/config/cache/cache.go
@@ -10,36 +10,51 @@ import (
 	pb "github.com/illumio/cloud-operator/api/illumio/cloud/k8sclustersync/v1"
 )
 
-// ConfiguredObjectCache stores configured Kubernetes objects received from CloudSecure.
-// It tracks the desired state that CloudSecure wants in the cluster.
+// ObjectCache is a generic thread-safe cache for storing objects by ID.
+// It supports atomic replacement for snapshot-based updates.
+//
+// This cache is generic to support both:
+//   - Config cache: stores *pb.ConfiguredKubernetesObjectData from CloudSecure (desired state)
+//   - Runtime cache: stores *unstructured.Unstructured from Kubernetes (actual state)
 //
 // Access patterns:
 //   - Snapshot: Build a local map, then call ReplaceAll() to atomically swap it in
 //   - Mutations: Use Insert() and Delete() which handle locking internally
-//   - Reading (for the reconciler): Use Get(), Values(), Len() which handle locking internally
+//   - Reading: Use Get(), Values(), Len() which handle locking internally
 //
 // Block on <-cache.IsReady() to wait for the first snapshot to complete before reading.
-type ConfiguredObjectCache struct {
+type ObjectCache[T any] struct {
 	mutex sync.RWMutex
 
-	// objects maps object ID to its ConfiguredKubernetesObjectData.
-	objects map[string]*pb.ConfiguredKubernetesObjectData
+	// objects maps object ID to its value.
+	objects map[string]T
 
 	// ready is closed when the first snapshot is complete.
 	ready chan struct{}
 }
 
-// NewConfiguredObjectCache creates a new cache instance.
-func NewConfiguredObjectCache() *ConfiguredObjectCache {
-	return &ConfiguredObjectCache{
-		objects: make(map[string]*pb.ConfiguredKubernetesObjectData),
+// NewObjectCache creates a new cache instance.
+func NewObjectCache[T any]() *ObjectCache[T] {
+	return &ObjectCache[T]{
+		objects: make(map[string]T),
 		ready:   make(chan struct{}),
 	}
 }
 
+// ConfiguredObjectCache is the cache type for CloudSecure configured objects.
+type ConfiguredObjectCache = ObjectCache[*pb.ConfiguredKubernetesObjectData]
+
+// TODO - The runtime cache will use unstructured objects because reconciliation may require
+// Kubernetes metadata fields not present in the proto, such as resourceVersion and ownerReferences
+
+// NewConfiguredObjectCache creates a new cache for configured objects.
+func NewConfiguredObjectCache() *ConfiguredObjectCache {
+	return NewObjectCache[*pb.ConfiguredKubernetesObjectData]()
+}
+
 // IsReady returns a channel that is closed when the first snapshot is complete.
 // Use <-cache.IsReady() to block until the cache has consistent data.
-func (c *ConfiguredObjectCache) IsReady() <-chan struct{} {
+func (c *ObjectCache[T]) IsReady() <-chan struct{} {
 	return c.ready
 }
 
@@ -47,7 +62,7 @@ func (c *ConfiguredObjectCache) IsReady() <-chan struct{} {
 // Use this for snapshot ingestion: build a local map, then call ReplaceAll to
 // swap it in. The cache remains consistent until the swap completes.
 // Also marks the cache as ready on first call to indicate cache now has valid data.
-func (c *ConfiguredObjectCache) ReplaceAll(objects map[string]*pb.ConfiguredKubernetesObjectData) {
+func (c *ObjectCache[T]) ReplaceAll(objects map[string]T) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
@@ -64,7 +79,7 @@ func (c *ConfiguredObjectCache) ReplaceAll(objects map[string]*pb.ConfiguredKube
 
 // Insert adds or updates an object in the cache.
 // Use for mutations after snapshot is complete.
-func (c *ConfiguredObjectCache) Insert(id string, obj *pb.ConfiguredKubernetesObjectData) {
+func (c *ObjectCache[T]) Insert(id string, obj T) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
@@ -73,15 +88,15 @@ func (c *ConfiguredObjectCache) Insert(id string, obj *pb.ConfiguredKubernetesOb
 
 // Delete removes an object from the cache by ID.
 // Use for mutations after snapshot is complete.
-func (c *ConfiguredObjectCache) Delete(id string) {
+func (c *ObjectCache[T]) Delete(id string) {
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
 
 	delete(c.objects, id)
 }
 
-// Get retrieves an object by ID. Returns nil if not found.
-func (c *ConfiguredObjectCache) Get(id string) *pb.ConfiguredKubernetesObjectData {
+// Get retrieves an object by ID. Returns the zero value if not found.
+func (c *ObjectCache[T]) Get(id string) T {
 	c.mutex.RLock()
 	defer c.mutex.RUnlock()
 
@@ -89,7 +104,7 @@ func (c *ConfiguredObjectCache) Get(id string) *pb.ConfiguredKubernetesObjectDat
 }
 
 // Values returns all objects in the cache, sorted by ID for consistency.
-func (c *ConfiguredObjectCache) Values() []*pb.ConfiguredKubernetesObjectData {
+func (c *ObjectCache[T]) Values() []T {
 	c.mutex.RLock()
 	defer c.mutex.RUnlock()
 
@@ -97,7 +112,7 @@ func (c *ConfiguredObjectCache) Values() []*pb.ConfiguredKubernetesObjectData {
 		return nil
 	}
 
-	result := make([]*pb.ConfiguredKubernetesObjectData, 0, len(c.objects))
+	result := make([]T, 0, len(c.objects))
 	for _, key := range slices.Sorted(maps.Keys(c.objects)) {
 		result = append(result, c.objects[key])
 	}
@@ -106,7 +121,7 @@ func (c *ConfiguredObjectCache) Values() []*pb.ConfiguredKubernetesObjectData {
 }
 
 // Len returns the number of objects in the cache.
-func (c *ConfiguredObjectCache) Len() int {
+func (c *ObjectCache[T]) Len() int {
 	c.mutex.RLock()
 	defer c.mutex.RUnlock()
 
--- a/internal/controller/stream/config/cache/cache.go
+++ b/internal/controller/stream/config/cache/cache.go
@@ -62,15 +62,8 @@ func (c *ConfiguredObjectCache) Reset() {
 }
 
 // InsertLocked adds or updates an object in the cache when the caller already holds the lock.
-// Use during snapshot ingestion after BeginSnapshot() has been called.
 func (c *ConfiguredObjectCache) InsertLocked(id string, obj *pb.ConfiguredKubernetesObjectData) {
 	c.objects[id] = obj
-}
-
-// DeleteLocked removes an object from the cache by ID when the caller already holds the lock.
-// Use during snapshot ingestion after BeginSnapshot() has been called.
-func (c *ConfiguredObjectCache) DeleteLocked(id string) {
-	delete(c.objects, id)
 }
 
 // Insert adds or updates an object in the cache.
--- a/internal/controller/stream/config/cache/cache.go
+++ b/internal/controller/stream/config/cache/cache.go
@@ -12,20 +12,28 @@ import (
 
 // ConfiguredObjectCache stores configured Kubernetes objects received from CloudSecure.
 // It tracks the desired state that CloudSecure wants in the cluster.
+
+// Access patterns:
+// 1. For snapshot ingestion, caller manages locking directly:
+//	cache.Mutex.Lock()
+//	cache.ClearLocked()
+//	cache.InsertLocked(id, obj)
+//	cache.NotifyReady()
+//	cache.Mutex.Unlock()
 //
-// The cache supports two access patterns:
-//  1. During snapshot ingestion: caller holds Mutex for entire snapshot, uses *Unlocked methods
-//  2. During mutations: caller uses regular methods which handle locking internally
-//
-// Use Ready() to wait for the first snapshot to complete before reading.
+// 2. For mutations after snapshot, use regular methods which handle locking internally.
+
+// Block on <-cache.Ready() to wait for the first snapshot to complete before reading.
 type ConfiguredObjectCache struct {
 	// Mutex is exported so callers can hold the lock across multiple operations
-	// (e.g., during snapshot ingestion). Use *Unlocked methods when holding this lock.
+	// (e.g., during snapshot ingestion). Use *Locked methods when holding this lock.
 	Mutex sync.RWMutex
 
-	// Maps object's ID as the key to its ConfiguredKubernetesObjectData as the value
+	// objects maps object ID to its ConfiguredKubernetesObjectData.
 	objects map[string]*pb.ConfiguredKubernetesObjectData
-	ready   chan struct{} // Closed when snapshot complete
+
+	// ready is closed when the first snapshot is complete.
+	ready chan struct{}
 }
 
 // NewConfiguredObjectCache creates a new cache instance.
@@ -37,32 +45,20 @@ func NewConfiguredObjectCache() *ConfiguredObjectCache {
 }
 
 // Ready returns a channel that is closed when the first snapshot is complete.
-// Use this to block until the cache has consistent data.
+// Use <-cache.Ready() to block until the cache has consistent data.
 func (c *ConfiguredObjectCache) Ready() <-chan struct{} {
 	return c.ready
 }
 
-// BeginSnapshot prepares the cache for a new snapshot.
-// It acquires the write lock and clears the cache.
-// The lock is held until EndSnapshot is called.
-// This blocks all readers until the snapshot is complete.
-func (c *ConfiguredObjectCache) BeginSnapshot() {
-	c.Mutex.Lock()
+// Reset removes all objects from the cache.
+// Caller MUST hold Mutex before calling this method.
+//
+// This method exists because the cache is a singleton with external references
+// (e.g., reconciliation loops waiting on Ready()). On stream reconnection,
+// the client clears the cache to ingest a fresh snapshot rather than creating
+// a new cache instance, which would orphan those references.
+func (c *ConfiguredObjectCache) Reset() {
 	c.objects = make(map[string]*pb.ConfiguredKubernetesObjectData)
-}
-
-// EndSnapshot completes the snapshot and releases the lock.
-// It also signals ready (idempotent - only first call closes the channel).
-func (c *ConfiguredObjectCache) EndSnapshot() {
-	// Close ready channel to signal first snapshot complete (idempotent)
-	select {
-	case <-c.ready:
-		// Already closed - do nothing
-	default:
-		close(c.ready)
-	}
-
-	c.Mutex.Unlock()
 }
 
 // InsertLocked adds or updates an object in the cache when the caller already holds the lock.
@@ -108,12 +104,12 @@ func (c *ConfiguredObjectCache) Values() []*pb.ConfiguredKubernetesObjectData {
 	c.Mutex.RLock()
 	defer c.Mutex.RUnlock()
 
-	return c.valuesUnlocked()
+	return c.valuesLocked()
 }
 
-// valuesUnlocked returns all objects sorted by ID, when you already hold the lock.
+// valuesLocked returns all objects sorted by ID when the caller already holds the lock.
 // Caller MUST hold at least a read lock.
-func (c *ConfiguredObjectCache) valuesUnlocked() []*pb.ConfiguredKubernetesObjectData {
+func (c *ConfiguredObjectCache) valuesLocked() []*pb.ConfiguredKubernetesObjectData {
 	if len(c.objects) == 0 {
 		return nil
 	}
@@ -134,18 +130,20 @@ func (c *ConfiguredObjectCache) Len() int {
 	return len(c.objects)
 }
 
-// LenUnlocked returns the number of objects when you already hold the lock.
+// LenLocked returns the number of objects when the caller already holds the lock.
 // Caller MUST hold at least a read lock.
-func (c *ConfiguredObjectCache) LenUnlocked() int {
+func (c *ConfiguredObjectCache) LenLocked() int {
 	return len(c.objects)
 }
 
-// Checks if the ready channel is closed (non-blocking).
-func (c *ConfiguredObjectCache) NotifyReady() bool {
+// NotifyReady marks the cache as ready, after a consistent snapshot has been
+// successfully inserted into the cache. This method is idempotent.
+// The mutex must be held for writes when calling this method.
+func (c *ConfiguredObjectCache) NotifyReady() {
 	select {
 	case <-c.ready:
-		return true // Channel closed = snapshot complete
+		// The channel is already closed. Do nothing.
 	default:
-		return false // Channel open = still ingesting
+		close(c.ready)
 	}
 }
--- a/internal/controller/stream/config/cache/cache.go
+++ b/internal/controller/stream/config/cache/cache.go
@@ -1,0 +1,98 @@
+// Copyright 2026 Illumio, Inc. All Rights Reserved.
+
+package cache
+
+import (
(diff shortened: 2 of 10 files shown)


api/illumio/cloud/k8sclustersync/v1/k8s_info.pb.go
rlenglet, line 1: Revert the unnecessary changes to this file.

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+
+	pb "github.com/illumio/cloud-operator/api/illumio/cloud/k8sclustersync/v1"
+)
+
+// ConfiguredObjectCache stores configured Kubernetes objects received from CloudSecure.
+// It tracks the desired state that CloudSecure wants in the cluster.
+type ConfiguredObjectCache struct {
+	mu               sync.RWMutex
+	objects          map[string]*pb.ConfiguredKubernetesObjectData
+	snapshotComplete bool
```
rlenglet: Remove this field from here. Cache readers shouldn't need to know about the state of the stream. They shouldn't have to care.
This boolean is needed to track the state of the stream, but that should be done as a local variable in `configClient.Run`.

On the other hand, readers need to know when the cache starts to be readable, i.e. got at least one consistent snapshot from a config stream.
I recommend using a simple channel, which would be closed after the first successful snapshot has been stored into the chache:


```suggestion
	ready <-chan struct{}
```

You should define a `Ready` method similar to the `Done` method in `context.Context` to return that channel to cache readers:
```
func (c *ConfiguredObjectCache) Ready() <-chan struct{} {
	return c.ready
}
```

Then readers will be able to wait for the cache to be ready using that, e.g. something like:
```
	<-cache.Ready():
	logger.Info("Configuration stream cache ready, starting reconciliation loop")
```

Define a `NotifyReady` method to close the channel safely. That method must be called by `configClient.Run` whenever it successfully completes a snapshot:

```
// NotifyReady marks the cache as ready, after a consistent snapshot has been successfully inserted into the cache.
// This method is idempotent.
// The mutex must be held for writes when calling this method.
func (c *ConfiguredObjectCache) NotifyReady {
	select {
	case <-c.ready:
		// The channel is already closed. Do nothing.
	default:
		close(c.ready)
	}
}
```

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	"sync"
+
+	pb "github.com/illumio/cloud-operator/api/illumio/cloud/k8sclustersync/v1"
+)
+
+// ConfiguredObjectCache stores configured Kubernetes objects received from CloudSecure.
+// It tracks the desired state that CloudSecure wants in the cluster.
+type ConfiguredObjectCache struct {
+	mu               sync.RWMutex
+	objects          map[string]*pb.ConfiguredKubernetesObjectData
```
rlenglet: Add a comment for this field.
Explain that each key is the object's ID.

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+
+// NewConfiguredObjectCache creates a new cache instance.
+func NewConfiguredObjectCache() *ConfiguredObjectCache {
+	return &ConfiguredObjectCache{
+		objects: make(map[string]*pb.ConfiguredKubernetesObjectData),
+	}
+}
+
+// Store adds or updates an object in the cache.
+func (c *ConfiguredObjectCache) Store(id string, obj *pb.ConfiguredKubernetesObjectData) {
```
rlenglet: In Golang, this operation is called "insert". Cf. https://pkg.go.dev/maps@go1.26.2#Insert

```suggestion
// Insert adds or updates an object in the cache.
func (c *ConfiguredObjectCache) Insert(id string, obj *pb.ConfiguredKubernetesObjectData) {
```

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// Get retrieves an object by ID. Returns nil and false if not found.
+func (c *ConfiguredObjectCache) Get(id string) (*pb.ConfiguredKubernetesObjectData, bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	obj, ok := c.objects[id]
+
+	return obj, ok
```
rlenglet:
```suggestion
	return c.objects[id]
```

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// List returns all objects in the cache.
+func (c *ConfiguredObjectCache) List() []*pb.ConfiguredKubernetesObjectData {
```
rlenglet: Name things consistently with the standard packages.
https://pkg.go.dev/maps@go1.26.2#Values
```suggestion
// Values returns all objects in the cache.
func (c *ConfiguredObjectCache) Values() []*pb.ConfiguredKubernetesObjectData {
```

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	result := make([]*pb.ConfiguredKubernetesObjectData, 0, len(c.objects))
```
rlenglet:
```suggestion
	if len(c.objects) == 0 {
		return nil
	}

	result := make([]*pb.ConfiguredKubernetesObjectData, 0, len(c.objects))
```

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	for _, obj := range c.objects {
```
rlenglet: To avoid any surprises, and to make unittesting easier, always return the list sorted by key.

```suggestion
	for _, key := range slices.Sorted(maps.Keys(c.objects)) {
		obj := c.objects[key]
```

internal/controller/stream/config/cache/cache.go
rlenglet: Why do you think this would be needed?
I don't see how useful that would be.

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+func (c *ConfiguredObjectCache) Len() int {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	return len(c.objects)
+}
+
+// Clear removes all objects from the cache and resets snapshotComplete to false.
+// This should be called on stream reconnection.
+func (c *ConfiguredObjectCache) Clear() {
```
rlenglet: I can't imagine when we would ever call this.

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// SetSnapshotComplete marks the snapshot as complete.
+// After this, only mutations should be processed, not resource data.
+func (c *ConfiguredObjectCache) SetSnapshotComplete() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.snapshotComplete = true
+}
```
rlenglet: Remove. Tracking the state of the stream doesn't belong here.
That state should remain local to `configClient.Run`.
```suggestion
```

internal/controller/stream/config/client.go
rlenglet, line 1: The locking logic is incorrect.
Reads from the cache should always be consistent.
With this implementation, before a snapshot is complete, the cache contents is inconsistent.

Reads should be blocked while ingesting a snapshot, and should not be allowed until after the first snapshot is complete.

So the logic should be:
1. When Run opens the stream:
    1. Lock the cache for writes. Until the snapshot is complete, the cache should remain locked for writes.
    1. Clear the cache.
1. For each resource received in the snapshot, put it into the cache.
1. When receiving a snapshot complete:
    1. Notify readers (e.g. the operator reconciliation code) that the cache is ready to be read.
    1. Unlock the cache.
1. When a mutation is received:
    1. Lock the cache for writes.
    1. Put/delete the resource into the cache.
    1. Unlock the cache.

When reconnecting the stream, the same will happen.

That way, cache readers will always read a consistent state of the set of resources, and can never read e.g. a partial snapshot.

To notify

rlenglet (changes requested):

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"sync"
+
+	pb "github.com/illumio/cloud-operator/api/illumio/cloud/k8sclustersync/v1"
+)
+
+// ConfiguredObjectCache stores configured Kubernetes objects received from CloudSecure.
+// It tracks the desired state that CloudSecure wants in the cluster.
+type ConfiguredObjectCache struct {
+	mu               sync.RWMutex
```
rlenglet: To make it easier for clients of this type to lock/unlock, I would simply export this field.

```suggestion
	Mutex            sync.RWMutex
```

rlenglet (changes requested):

internal/controller/stream/config/cache/cache.go
rlenglet: And document for each method whether the mutex must be locked when calling it.

ryan-er: Client meaning users of the cache? Why would client need to lock by themselves? If reads are going to block when cache is not ready anyways, and the cache provide ways to check readiness e.g. blocking `Ready()` method as you described in other comment, and if needed a non blocking `IsReady()` method.

rlenglet: > Why would client need to lock by themselves?

To consistently read multiple entries in the cache.

internal/controller/stream/config/client.go
```diff
@@ -31,6 +32,7 @@ type configClient struct {
 	verboseDebugging   bool
 	bufferedGrpcSyncer *logging.BufferedGrpcWriteSyncer
 	stats              *stream.Stats
+	cache              *cache.ConfiguredObjectCache
```
ryan-er, line 35: How will reconciler get to this `cache`?  Should the lifecycle or lifetime of the `cache` be tied to the stream client or the reconciler?

internal/controller/stream/config/cache/cache.go
rlenglet: Locking is required, because the writer and reader will be running in different goroutines. To ensure memory consistency, per the [Go Memory Model](https://go.dev/ref/mem), just like in any other programming language, synchronization is required. E.g. [using Mutexes](https://go.dev/ref/mem#locks).

internal/controller/stream/config/client.go
rlenglet: No, the cache will be created once and for all, and shared/injected between the reader and the writer.
The cache will persist across configuration stream reconnections.

internal/controller/stream/config/cache/cache.go
ryan-er: Understood, I was asking about whether it needs to be exposed when it's properly synchronized inside the Reads and Writes. Reading multiple makes sense.. But makes it weird then that some synchronization is handled/abstracted inside the cache, and some are responsible by the client/user of the cache.

Or we could perhaps have a method/function in the cache that allows for reading multiple keys at once, that way the mutex still doesn't need to be exposed, correct?

internal/controller/stream/config/client.go
ryan-er: Cool so right now it's per stream client https://github.com/illumio/cloud-operator/pull/407/changes#diff-cd518860f1361aa91f974fafe99c903ea69a58d610190eb2d585eaa579250288R45, but we can address in different PR.

internal/controller/stream/config/cache/cache.go
rlenglet: > handled/abstracted inside the cache

It can't be "abstracted inside the cache", as multiple operations must be performed atomically, as I explained in https://github.com/illumio/cloud-operator/pull/407#discussion_r3165758020.

internal/controller/stream/config/client.go
rlenglet: No, it's a singleton.
It's technically per stream client **factory**, but since that is a singleton, the cache is also technically a singleton.

rlenglet: Actually, good catch. It MUST be a singleton, and it's not right now. I will add a comment about that.

internal/controller/stream/config/factory.go
rlenglet, line 25: The cache must be a singleton injected into the factory.

```suggestion
	Stats              *stream.Stats
	Cache              *cache.ConfiguredObjectCache
```

Create and set it in `main`.

internal/controller/stream/config/factory.go
```diff
@@ -41,6 +42,7 @@ func (f *Factory) NewStreamClient(ctx context.Context, grpcConn grpc.ClientConnI
 		verboseDebugging:   f.VerboseDebugging,
 		bufferedGrpcSyncer: f.BufferedGrpcSyncer,
 		stats:              f.Stats,
+		cache:              cache.NewConfiguredObjectCache(),
```
rlenglet:
```suggestion
		cache:              f.Cache,
```

rlenglet (changes requested):

internal/controller/stream/config/cache/cache.go
```diff
@@ -3,96 +3,149 @@
...
+// Use during snapshot ingestion after BeginSnapshot() has been called.
+func (c *ConfiguredObjectCache) InsertLocked(id string, obj *pb.ConfiguredKubernetesObjectData) {
+	c.objects[id] = obj
+}
+
+// DeleteLocked removes an object from the cache by ID when the caller already holds the lock.
+// Use during snapshot ingestion after BeginSnapshot() has been called.
+func (c *ConfiguredObjectCache) DeleteLocked(id string) {
+	delete(c.objects, id)
+}
```
rlenglet, line 78: This is not needed. Snapshots only insert resources.
```suggestion
```

internal/controller/stream/config/cache/cache.go
```diff
@@ -3,96 +3,149 @@
...
 
-// SetSnapshotComplete marks the snapshot as complete.
-// After this, only mutations should be processed, not resource data.
-func (c *ConfiguredObjectCache) SetSnapshotComplete() {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-
-	c.snapshotComplete = true
+// Checks if the ready channel is closed (non-blocking).
+func (c *ConfiguredObjectCache) NotifyReady() bool {
```
rlenglet: Using the term "notify" here doesn't sound right.
How about `IsReady`?

And move this just under the `Ready` method at the top of the file, as they go together.

internal/controller/stream/config/cache/cache.go
```diff
@@ -3,96 +3,149 @@
...
 	"sync"
 
 	pb "github.com/illumio/cloud-operator/api/illumio/cloud/k8sclustersync/v1"
 )
 
 // ConfiguredObjectCache stores configured Kubernetes objects received from CloudSecure.
 // It tracks the desired state that CloudSecure wants in the cluster.
+//
+// The cache supports two access patterns:
+//  1. During snapshot ingestion: caller holds Mutex for entire snapshot, uses *Unlocked methods
```
rlenglet: It would be more logical to me if the methods that require the mutex to be locked were called `*Locked` instead of `*Unlocked`.

internal/controller/stream/config/client.go
```diff
@@ -31,16 +32,27 @@ type configClient struct {
...
+	c.cache.BeginSnapshot()
+	c.logger.Debug("Started configuration snapshot, cache locked")
+
 	for {
 		select {
 		case <-ctx.Done():
+			// If we haven't completed snapshot, unlock before returning
+			if !c.cache.NotifyReady() {
+				c.cache.EndSnapshot()
+			}
```
rlenglet: A problem here is that you're marking the cache as ready, which is incorrect.

If we're in the middle of the initial snapshot, the cache may be empty, or it may contain a partial snapshot, which is anything but ready. In this case, the cache is not ready, so you MUST NOT close the channel.

Furthermore, if we're in the middle of a snapshot after a reconnection, i.e. this is not the first snapshot, there's currently no way to revert to a consistent state (the state of the cache before `BeginSnapshot`). And the cache has already been marked as ready, because a previous snapshot succeeded. So the cache MUST NOT be unlocked in this state.

To address both cases correctly, I can think of 2 solutions:

1. Do nothing and return. If the cache is already locked, i.e., we're in the middle of a snapshot, the cache will remain locked until we reconnect and we go through another snapshot. Note that you'll need to modify `BeginSnapshot` to use `TryLock` instead of `Lock`, because the cache may already be locked.
1. When getting the snapshot, instead of modifying the cache directly, build a local variable map with the new resources. When receiving a snapshot commit message, we replace the cache's map with that local map, atomically while locking the cache. That swap only happens when receiving a commit, and until then the state of the cache remains consistent, so we can unlock and return here without any issues.

I think the 2nd solution is cleaner. It's tricky to get the logic right with `TryLock`, etc.

In any case, delete this code here.
```suggestion
```

internal/controller/stream/config/client.go
```diff
@@ -49,12 +61,22 @@ func (c *configClient) Run(ctx context.Context) error {
 		if errors.Is(err, io.EOF) {
 			c.logger.Info("Server closed the GetConfigurationUpdates stream")
 
+			// If we haven't completed snapshot, unlock before returning
+			if !c.cache.NotifyReady() {
+				c.cache.EndSnapshot()
+			}
```
rlenglet: That's incorrect. `EndSnapshot` must only be called when a snapshot is **completed successfully**.

internal/controller/stream/config/client.go
```diff
@@ -49,12 +61,22 @@ func (c *configClient) Run(ctx context.Context) error {
...
 			return nil
 		}
 
 		if err != nil {
 			c.logger.Error("Configuration stream terminated", zap.Error(err))
 
+			// If we haven't completed snapshot, unlock before returning
+			if !c.cache.NotifyReady() {
+				c.cache.EndSnapshot()
+			}
```
rlenglet: That's incorrect. `EndSnapshot` must only be called when a snapshot is **completed successfully**.

internal/controller/stream/config/client.go
```diff
@@ -68,44 +90,44 @@ func (c *configClient) Run(ctx context.Context) error {
...
-		)
+		c.handleUpdateConfiguration(update.UpdateConfiguration)
 
-		if c.verboseDebugging {
-			c.logger.Debug("verboseDebugging is true, setting log level to debug")
-			c.bufferedGrpcSyncer.UpdateLogLevel(pb.LogLevel_LOG_LEVEL_DEBUG)
-		} else {
-			c.bufferedGrpcSyncer.UpdateLogLevel(update.UpdateConfiguration.GetLogLevel())
+	case *pb.GetConfigurationUpdatesResponse_ResourceData:
+		if c.cache.NotifyReady() {
```
rlenglet: That is logically incorrect. The cache may be ready, but that may be because a snapshot successfully completed for a previous stream. We may have reconnected since. So the cache being ready means nothing about the state of the current stream.
You must maintain a separate boolean as a local variable in this method to track the state of the stream: before or after snapshot commit.
The state of the cache is not a substitute to this local boolean variable.

internal/controller/stream/config/client.go
```diff
@@ -68,44 +90,44 @@ func (c *configClient) Run(ctx context.Context) error {
...
+			zap.String("name", update.ResourceData.GetName()),
 		)
 
 	case *pb.GetConfigurationUpdatesResponse_ResourceSnapshotComplete:
-		c.logger.Info("Received configured object snapshot complete")
+		if c.cache.NotifyReady() {
+			c.logger.Warn("Received duplicate ResourceSnapshotComplete, ignoring")
+
+			return nil
+		}
```
rlenglet: That is incorrect. The cache may be ready because a snapshot was successfully completed in a previous stream. It says nothing about the current stream.

```suggestion
```

internal/controller/stream/config/client.go
```diff
@@ -68,44 +90,44 @@ func (c *configClient) Run(ctx context.Context) error {
...
-				zap.String("id", m.DeleteObject.GetId()),
-			)
-		default:
-			c.logger.Error("Received unknown configured object mutation, closing stream", zap.Any("response", resp))
+		if !c.cache.NotifyReady() {
+			c.logger.Warn("Received ResourceMutation before snapshot complete, ignoring")
 
-			return errors.New("server sent unknown configured object mutation type")
+			return nil
+		}
```
rlenglet: Do not rely on the state of the cache.
Only rely on a local boolean variable to capture the state of the stream: before or after snapshot commit.

internal/controller/stream/config/cache/cache.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// LenUnlocked returns the number of objects when you already hold the lock.
+// Caller MUST hold at least a read lock.
+func (c *ConfiguredObjectCache) LenUnlocked() int {
+	return len(c.objects)
+}
+
+// Checks if the ready channel is closed (non-blocking).
+func (c *ConfiguredObjectCache) NotifyReady() bool {
```
rlenglet: Delete this method altogether. There is no correct usage of it in `client.go`, so you should just delete it.
Add a boolean variable into the client's `Run` method to track the state of the stream (before or after snapshot commit), instead of incorrectly relying on the state of the cache, which lifetime spans multiple streams, to reflect the state of any particular stream.

rlenglet (changes requested):

internal/controller/stream/config/client.go
byjustina: Sorry, I was in the process of revising the pull request - this is an old version. I've just pushed a new version

internal/controller/stream/config/client.go
byjustina: Apologies, this was an old version - I've updated the current version to reflect this

internal/controller/stream/config/cache/cache.go
rlenglet: Comment still open. Small optimization.

internal/controller/stream/config/client.go
```diff
@@ -31,16 +32,39 @@ type configClient struct {
...
+// isSnapshotComplete checks if the cache has completed its first snapshot.
+func (c *configClient) isSnapshotComplete() bool {
+	select {
+	case <-c.cache.Ready():
+		return true
+	default:
+		return false
+	}
+}
+
```
rlenglet: Delete that. You must define a local variable in `Run` or, better, a field in `configClient` to track the state of the current stream.
The cache is potentially used across many consecutive streams, so it cannot reflect the state of the current stream.

```suggestion
```

internal/controller/stream/config/client.go
```diff
@@ -31,16 +32,39 @@ type configClient struct {
...
+	c.cache.Mutex.Lock()
+	c.cache.Reset()
+
+	c.logger.Debug("Started configuration snapshot, cache locked")
+
 	for {
 		select {
 		case <-ctx.Done():
+			// If we haven't completed snapshot, unlock before returning
+			if !c.isSnapshotComplete() {
```
rlenglet:
```suggestion
			if !c.snapshotCompleted {
```

internal/controller/stream/config/client.go
```diff
@@ -49,12 +73,22 @@ func (c *configClient) Run(ctx context.Context) error {
 		if errors.Is(err, io.EOF) {
 			c.logger.Info("Server closed the GetConfigurationUpdates stream")
 
+			// If we haven't completed snapshot, unlock before returning
+			if !c.isSnapshotComplete() {
```
rlenglet:
```suggestion
			if !c.snapshotCompleted {
```

internal/controller/stream/config/client.go
```diff
@@ -49,12 +73,22 @@ func (c *configClient) Run(ctx context.Context) error {
...
+			}
+
 			return nil
 		}
 
 		if err != nil {
 			c.logger.Error("Configuration stream terminated", zap.Error(err))
 
+			// If we haven't completed snapshot, unlock before returning
+			if !c.isSnapshotComplete() {
```
rlenglet:
```suggestion
			if !c.snapshotCompleted {
```

internal/controller/stream/config/client.go
rlenglet, line 30:
```suggestion
	stream             ConfigurationStream
	snapshotCompleted bool // True after the snapshot has been completed in the stream.
```

internal/controller/stream/config/client.go
```diff
@@ -68,44 +102,46 @@ func (c *configClient) Run(ctx context.Context) error {
...
-		)
+		c.handleUpdateConfiguration(update.UpdateConfiguration)
 
-		if c.verboseDebugging {
-			c.logger.Debug("verboseDebugging is true, setting log level to debug")
-			c.bufferedGrpcSyncer.UpdateLogLevel(pb.LogLevel_LOG_LEVEL_DEBUG)
-		} else {
-			c.bufferedGrpcSyncer.UpdateLogLevel(update.UpdateConfiguration.GetLogLevel())
+	case *pb.GetConfigurationUpdatesResponse_ResourceData:
+		if c.isSnapshotComplete() {
```
rlenglet:
```suggestion
		if c.snapshotCompleted {
```

internal/controller/stream/config/client.go
```diff
@@ -68,44 +102,46 @@ func (c *configClient) Run(ctx context.Context) error {
...
+			c.logger.Warn("Received ResourceData after snapshot complete, ignoring")
+
+			return nil
```
rlenglet: Return an error to restart the stream.
The server sent garbage, so we can't continue.

internal/controller/stream/config/client.go
```diff
@@ -68,44 +102,46 @@ func (c *configClient) Run(ctx context.Context) error {
...
+		// During snapshot, we hold the lock
+		c.cache.InsertLocked(update.ResourceData.GetId(), update.ResourceData)
+		c.logger.Debug("Stored configured object from snapshot",
 			zap.String("id", update.ResourceData.GetId()),
+			zap.String("name", update.ResourceData.GetName()),
 		)
 
 	case *pb.GetConfigurationUpdatesResponse_ResourceSnapshotComplete:
-		c.logger.Info("Received configured object snapshot complete")
+		if c.isSnapshotComplete() {
```
rlenglet:
```suggestion
		if c.snapshotCompleted {
```

internal/controller/stream/config/client.go
```diff
@@ -68,44 +102,46 @@ func (c *configClient) Run(ctx context.Context) error {
...
+			c.logger.Warn("Received duplicate ResourceSnapshotComplete, ignoring")
+
+			return nil
```
rlenglet: Return an error to restart the stream.

internal/controller/stream/config/client.go
```diff
@@ -68,44 +102,46 @@ func (c *configClient) Run(ctx context.Context) error {
...
+		}
+
+		// Snapshot complete - signal ready and release lock
+		c.cache.NotifyReady()
```
rlenglet:
```suggestion
		c.cache.NotifyReady()
		c.snapshotCompleted = true
```

internal/controller/stream/config/client.go
```diff
@@ -68,44 +102,46 @@ func (c *configClient) Run(ctx context.Context) error {
...
-			c.logger.Debug("Received update object mutation",
-				zap.String("id", m.UpdateObject.GetId()),
-			)
-		case *pb.ConfiguredKubernetesObjectMutation_DeleteObject:
-			c.logger.Debug("Received delete object mutation",
-				zap.String("id", m.DeleteObject.GetId()),
-			)
-		default:
-			c.logger.Error("Received unknown configured object mutation, closing stream", zap.Any("response", resp))
+		if !c.isSnapshotComplete() {
```
rlenglet:
```suggestion
		if !c.snapshotCompleted {
```

internal/controller/stream/config/client.go
```diff
@@ -68,44 +102,46 @@ func (c *configClient) Run(ctx context.Context) error {
...
+			c.logger.Warn("Received ResourceMutation before snapshot complete, ignoring")
 
-			return errors.New("server sent unknown configured object mutation type")
+			return nil
```
rlenglet: Return an error to restart the stream.

rlenglet (changes requested):
Looks much better, thanks!
Still, you can't rely on the "ready" state of the cache to determine whether the current stream has completed a snapshot. You need to keep state in the `configClient`, using a new bool field, to track that, because that state is specific to each stream.
I made specific recommendations to fix that.

internal/controller/stream/config/client.go
```diff
@@ -31,16 +32,39 @@ type configClient struct {
...
+	default:
+		return false
+	}
+}
+
 // Run receives configuration updates from the server until the stream closes.
 func (c *configClient) Run(ctx context.Context) error {
+	// Lock cache writes, and clear for new snapshot
+	// Readers are blocked until we unlock
+	c.cache.Mutex.Lock()
```
rlenglet: Suggestion to simplify the code: replace all the `if !c.snapshotCompleted {` tests below with a single `defer`:

```suggestion
	c.cache.Mutex.Lock()
	defer func() {
		if !c.snapshotCompleted {
			c.cache.Mutex.Unlock()
		}
	}()
```

rlenglet (changes requested):

internal/controller/stream/config/client.go
byjustina: Updated to use a local variable in Run() instead because I think that makes the per-stream scope clearer

rlenglet (dismissed):
Very nice, thank you!

rlenglet:
Please address the Linting errors.

ryan-er (dismissed):
Looks good. The only thing is we don't have a way to do multiple reads/gets and Romain spoke about exposing the mutex. Otherwise we can also have a function to read multiple keys. Ok to address in future PR IMO.

rlenglet:
@ryan-er I think the main access patterns on the read side will be  1) reading an individual key atomically, and 2) atomically computing a diff between two cache instances. The latter can likely be done by adding a single method to the current implementation. So I agree this can be done in a future PR.

byjustina:
It seems like the local linter output  (`make lint`) differs from CI lint output, but I've just updated the PR to address the linting issues reflected on CI

rlenglet:
@byjustina The golangci-lint version does matter a lot. You need to use the exact same version we're using in the build pipeline, which is also the same version that is installed in the devcontainer.
https://github.com/illumio/cloud-operator/blob/b85f6d8059f9903b358c4783cece0f51630e427d/.github/workflows/go-security-checks.yml#L39
https://github.com/illumio/cloud-operator/blob/b85f6d8059f9903b358c4783cece0f51630e427d/.devcontainer/Dockerfile#L6

I would recommend that you switch to using a devcontainer, either locally in VSCode or using GitHub Codespaces if you have some credits, to make sure that you have the right versions of the tools installed.

rlenglet:
@byjustina You can also upgrade the golangci-lint version in the devcontainer and GigHub Action workflow to the version you are using, if that's a higher version than the one we're currently using. 😀

byjustina:
@rlenglet Thanks, installing dev-containers through VSCode worked! Confirmed that the linting issues have been updated accordingly

rlenglet (approved):
Thank you!

PR #422, opened 2026-05-29
Unify cache keys, use field manager ownership for runtime cache, & add integration tests                                                          

byjustina:
Replace server-assigned CloudSecure IDs with deterministic kind/namespace/name cache keys across both config and runtime caches. This ensures the reconciler can directly compare desired state (from CloudSecure) against observed state (from Kubernetes) without needing labels, which are subject to external controller changes.                                                        

  **Key changes:**
  - Cache keys: Both caches now use `kind/namespace/name` instead of opaque server IDs
  - Ownership detection: Replace ManagedByLabel with SSA `managedFields` check for determining if a K8s object is operator-managed
  - Add reconciliation integration tests infra


internal/controller/reconciler/testdata/crds/ciliumcidrgroups.yaml
```diff
@@ -0,0 +1,113 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
byjustina: These files are pulled from the official cilium spec. I could use `x-kubernetes-preserve-unknown-fields: true` instead of pulling the entire spec but I thought this would be best since it enforces comprehensive schema validation

rlenglet: Copying code like this without preserving the copyright and license notices violates Cilium's copyright.
You must preserve the license and copyright notices as comments at the top of this file.
For example, copy this:
https://github.com/cilium/cilium/blob/65169cd77dcfd401f7b8712fbc33e04ef2cf74b4/pkg/k8s/apis/cilium.io/client/crds/v2/embed.go#L1-L2

rlenglet: Why couldn't you just download the files dynamically from Cilium's GitHub repo?
Why do these files have to be copied and committed into this repo?

rlenglet (changes requested):

internal/controller/reconciler/testdata/crds/ciliumcidrgroups.yaml
byjustina: I didn't want any of the tests to have to rely on network access & potentially be flaky if I were to download the files, but I suppose a better alternative is to download the specs once in the Makefile and keep them as generated artifacts, will update to do this instead

internal/controller/stream/config/client.go
```diff
@@ -42,6 +44,7 @@ type configClient struct {
 func (c *configClient) Run(ctx context.Context) error {
 	// Local state for THIS stream (not the cache's global state)
 	pendingSnapshot := make(map[string]*pb.ConfiguredKubernetesObjectData)
+	cloudSecureIDToKey := make(map[string]string) // maps CloudSecure-assigned Id → cache key
```
byjustina: This is needed because the `DeleteConfiguredKubernetesObject` currently only carries the id field, and we need to map the ID to the cache key otherwise we won't know what obj to delete. An option would be to modify the protobuf message to send down the `kind/namespace/name` instead since id is not really used anywhere else at this point

internal/controller/stream/config/client.go
```diff
@@ -86,7 +89,21 @@ func (c *configClient) handleConfigUpdate(ctx context.Context, resp *pb.GetConfi
...
+				zap.Error(err))
+
+			return nil
+		}
+
+		cloudSecureIDToKey[update.ResourceData.GetId()] = key
+		// Clear the CloudSecure-assigned Id before caching. The cache uses kind/namespace/name
+		// keys, and the reconciler compares config vs runtime objects with proto.Equal — runtime
+		// objects don't have this field, so leaving it would cause a permanent mismatch.
+		update.ResourceData.Id = ""
```
byjustina: Removing id from the protobuf message would eliminate this manual handling

internal/controller/stream/config/client.go
```diff
@@ -114,6 +131,35 @@ func (c *configClient) handleConfigUpdate(ctx context.Context, resp *pb.GetConfi
...
+			m.CreateOrUpdateObject.Id = ""
+		case *pb.ConfiguredKubernetesObjectMutation_DeleteObject:
+			key, ok := cloudSecureIDToKey[m.DeleteObject.GetId()]
+			if !ok {
+				c.logger.Warn("Ignoring delete for unknown CloudSecure Id", zap.String("id", m.DeleteObject.GetId()))
+
+				return nil
+			}
+
+			delete(cloudSecureIDToKey, m.DeleteObject.GetId())
```
byjustina: see comment above too

.github/workflows/integration_envtest_tests.yml
rigdomin-illumio, line 1: Let's rename to match naming structure of other integration test GH Action workflows

.github/workflows/envtest.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: "go.mod"
+          cache: true
+
+      - name: Install envtest binaries
+        run: |
+          go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
rigdomin-illumio: This should match and be pinned to what we have in go.mod; otherwise, could be misleading.

.github/workflows/integration_envtest_tests.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: "go.mod"
+          cache: true
+
+      - name: Install envtest binaries
+        run: |
+          go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+          echo "KUBEBUILDER_ASSETS=$(setup-envtest use --print path)" >> $GITHUB_ENV
```
rigdomin-illumio: This looks to be pulling k8s CLI tools?

.github/workflows/integration_envtest_tests.yml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Run envtest tests
```
rigdomin-illumio, line 39: missing step for `download-crds`.
why not use make targets here?

internal/controller/reconciler/testdata/policies/init.yaml
```diff
@@ -0,0 +1,19 @@
...
+    ingress:
+    - fromEntities:
+      - host
+    egress:
+    - toEntities:
+      - all
+      toPorts:
+      - ports:
+        - port: "53"
+          protocol: UDP
```
rigdomin-illumio: missing EOF newline

internal/controller/reconciler/setup_envtest_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return h
+}
+
+// setupSuite wires the full production pipeline against envtest:
+//
+//	fakeserver (gRPC) → config client → config cache ──→ reconciler → envtest K8s API
+//	                     resources client → runtime cache ─┘
+//
+func setupSuite(t *testing.T) *fakeserver.FakeServer {
```
rigdomin-illumio, line 121: nit:
This looks like a lot of setup steps, and I see many tests in `reconciler_integration_test`, each call this `setupSuite()`.

We can group similar tests together and use `t.Run()` with a single, shared, `setupSuite()` call.

Makefile
```diff
@@ -7,6 +7,9 @@ LOCAL_IMAGE := $(LOCAL_REGISTRY)/$(APP_NAME)
 COMMIT := $(shell git rev-parse --short HEAD)
 DATE := $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 LDFLAGS := -ldflags "-X main.Version=latest -X main.Commit=$(COMMIT) -X main.Date=$(DATE)"
+CILIUM_VERSION := $(shell grep 'github.com/cilium/cilium ' go.mod | awk '{print $$2}')
```
rigdomin-illumio: This seems very risky.
Perhaps we can look to check what is currently installed (either local or in CI) using `go list` or `gomodcache` env var.

A random change in go.mod format could affect this.

Makefile
```diff
@@ -24,6 +27,20 @@ test:
 	@echo "Running tests..."
 	go test ./...
 
+# Download Cilium CRD schemas for envtest
+.PHONY: download-crds
+download-crds:
```
rigdomin-illumio, line 36: We should protect in case of the event these curl commands should fail.

rigdomin-illumio (changes requested):
Great work! Nice coverage.

There are some things that I'm a bit hesitant on and want to think further on, for now I left a few comments.

rigdomin-illumio:
Please update your PR with latest changes from main

pavankumarinnamuri (dismissed):
_Superseded by inline review — please disregard._

internal/convert/policy.go
```diff
@@ -52,25 +43,46 @@ func ExtractResourceName(data *pb.ConfiguredKubernetesObjectData) (string, error
...
+// already-converted KubernetesObjectData for the runtime cache. Labels and
+// annotations are copied as-is so the reconciler can detect drift via proto.Equal.
+func BuildConfiguredFromMetadata(metadata *pb.KubernetesObjectData) (*pb.ConfiguredKubernetesObjectData, error) {
 	configured := &pb.ConfiguredKubernetesObjectData{
-		Id:          id,
 		Name:        metadata.GetName(),
 		Namespace:   metadata.Namespace,
 		Annotations: metadata.GetAnnotations(),
-		Labels:      filteredLabels,
+		Labels:      metadata.GetLabels(),
```
pavankumarinnamuri, line 105: Runtime cache now stores the live object's full label set, but the config cache only holds CloudSecure's labels. `reconcileObject` compares them with `proto.Equal` (reconciler.go:176), so any policy a third-party controller decorates with a label never compares equal → the reconciler re-applies on every event and every interval tick (a hot loop, not just the safety net). Suggest restricting runtime-side labels to operator-owned keys before caching, and adding a convergence assertion (resourceVersion stops changing) to `TestReconciler_ExternalUserLabelsPreserved`.

internal/convert/policy.go
```diff
@@ -52,25 +43,46 @@ func ExtractResourceName(data *pb.ConfiguredKubernetesObjectData) (string, error
...
+		return "CiliumCIDRGroup", nil
+	default:
+		return "", fmt.Errorf("unsupported kind_specific type: %T", data.GetKindSpecific())
 	}
+}
 
+// NewCacheKey builds a cache key from kind, namespace, and name.
+// Both config and runtime caches use this key format so lookups match
+// without needing labels or reverse indexes.
+func NewCacheKey(kind, namespace, name string) string {
```
pavankumarinnamuri: `NewCacheKey` returns a string, not a constructor — `New` is misleading; prefer `CacheKey(...)`. Relatedly `CacheKeyFromObj` abbreviates `Object` inconsistently with `ConvertToApplyObject`/`BuildConfiguredFromMetadata` → `CacheKeyForObject(...)`.

internal/convert/policy.go
```diff
@@ -52,25 +43,46 @@ func ExtractResourceName(data *pb.ConfiguredKubernetesObjectData) (string, error
...
-// matches the shape of the config cache. Annotations are passed through as-is
-// because the operator doesn't add any — SSA handles annotation ownership.
-func BuildConfiguredFromMetadata(id string, metadata *pb.KubernetesObjectData) (*pb.ConfiguredKubernetesObjectData, error) {
-	filteredLabels := make(map[string]string, len(metadata.GetLabels()))
-	for k, v := range metadata.GetLabels() {
-		if k != CloudSecureIDLabel && k != ManagedByLabel {
-			filteredLabels[k] = v
-		}
+// ExtractKind returns the Kubernetes kind string for a configured object.
+func ExtractKind(data *pb.ConfiguredKubernetesObjectData) (string, error) {
```
pavankumarinnamuri, line 71: `ExtractKind`, `ExtractResourceName`, and the kind label in `marshalConfiguredObjectSpecs` are three copies of the same CNP/CCNP/CIDRGroup mapping. This PR adds two more parallel switches; consider consolidating into one table (at minimum, don't leave the count higher).

internal/controller/reconciler/reconciler.go
```diff
@@ -46,10 +44,11 @@ func NewReconciler(
...
 	return &Reconciler{
-		logger:       logger,
-		client:       client,
-		configCache:  configCache,
-		runtimeCache: runtimeCache,
+		logger:            logger,
+		client:            client,
+		configCache:       configCache,
+		runtimeCache:      runtimeCache,
+		reconcileInterval: DefaultReconcileInterval,
```
pavankumarinnamuri, line 52: `reconcileInterval` is defaulted here but not a constructor parameter, so nothing can override it — the integration tests are stuck on the 5-minute timer and never exercise the periodic full-reconcile path. Suggest a `WithReconcileInterval` functional option.

internal/controller/reconciler/reconciler.go
```diff
@@ -157,7 +156,7 @@ func (r *Reconciler) processResourceChange(ctx context.Context, id string, recon
 			r.logger.Error("Full reconciliation failed", zap.Error(err))
 		}
 
-		reconcileTimer.Reset(FullReconcileInterval)
+		reconcileTimer.Reset(r.reconcileInterval)
 	} else {
 		if err := r.reconcileObject(ctx, id); err != nil {
 			r.logger.Error("Object reconciliation failed", zap.String("id", id), zap.Error(err))
```
pavankumarinnamuri: The `id`→`key` rename stopped short: `reconcileObject`/`processResourceChange` params and `zap.String("id", …)` now carry `kind/namespace/name` keys. Also `configObj.GetId()` in `applyObject`/`deleteObject` is always `""` (config client blanks it at client.go:104), so those log fields are dead. Rename to `key` and drop the empty fields.

internal/controller/stream/config/client.go
```diff
@@ -86,7 +88,21 @@ func (c *configClient) handleConfigUpdate(ctx context.Context, resp *pb.GetConfi
 			return errors.New("server sent ResourceData after snapshot complete")
 		}
 
-		pendingSnapshot[update.ResourceData.GetId()] = update.ResourceData
+		key, err := convert.CacheKeyFromObj(update.ResourceData)
+		if err != nil {
+			c.logger.Warn("Skipping unsupported resource type",
```
pavankumarinnamuri: Unsupported-kind skips (here and :142) and the unknown-delete-Id skip (:157) log at `Warn`, but these are expected (unsupported kind, or a delete racing a reconnect), not degraded operation. Downgrade to `Debug`.

internal/controller/stream/resources/client.go
```diff
@@ -302,42 +302,51 @@ func (c *resourcesClient) CreateMutationObject(metadata *pb.KubernetesObjectData
...
 //
 // For list operations (eventType is empty), the handler accumulates the configured object
 // into pendingSnapshot for later ReplaceAll — the cache is not modified directly.
 func (c *resourcesClient) newRuntimeCacheHandler(pendingSnapshot map[string]*pb.ConfiguredKubernetesObjectData) RuntimeCacheHandler {
-	return func(ctx context.Context, eventType watch.EventType, metadata *pb.KubernetesObjectData) error {
-		labels := metadata.GetLabels()
-		if labels[convert.ManagedByLabel] != convert.ManagedByValue {
-			return nil
+	return func(ctx context.Context, eventType watch.EventType, obj *unstructured.Unstructured, metadata *pb.KubernetesObjectData) error {
+		configured, err := convert.BuildConfiguredFromMetadata(metadata)
```
pavankumarinnamuri: Both `//nolint:nilerr` returns drop unsupported kinds with zero logging — a watched-but-unclassifiable resource silently vanishes. The config client logs the equivalent skip; add a `Debug` line here. Also, this conversion + `CacheKeyFromObj` runs on every watch event before the cheap `hasFieldManager` check below; for non-owned objects (the common case) it's wasted — check ownership first, build the configured object/key only when ownership might hold.

internal/controller/stream/resources/client.go
```diff
@@ -346,6 +355,17 @@ func (c *resourcesClient) newRuntimeCacheHandler(pendingSnapshot map[string]*pb.
 	}
 }
 
+// hasFieldManager reports whether the object's managedFields includes the given manager.
+func hasFieldManager(obj *unstructured.Unstructured, manager string) bool {
+	for _, mf := range obj.GetManagedFields() {
+		if mf.Manager == manager {
```
pavankumarinnamuri: `hasFieldManager` matches on Manager name only. SSA can leave multiple entries under one manager (a stale `Operation: Update`, or a `status` subresource). To mean "we SSA-own this object," also require `mf.Operation == metav1.ManagedFieldsOperationApply && mf.Subresource == ""`.

internal/controller/stream/config/cache/cache.go
```diff
@@ -115,52 +116,60 @@ func (c *ObjectCache[T]) ReplaceAll(ctx context.Context, objects map[string]T) e
...
 	c.mutex.RLock()
 	defer c.mutex.RUnlock()
 
-	return c.objects[id]
+	return c.objects[key]
 }
 
-// Values returns all objects in the cache, sorted by ID for consistency.
+// Keys returns all cache keys, sorted for consistency.
+func (c *ObjectCache[T]) Keys() []string {
```
pavankumarinnamuri, line 192: `Keys()` sorts under the lock on every call, but `reconcileAll` only unions both caches' keys into a set (order discarded) — wasted O(n log n) + allocation per full reconcile. Consider an unsorted `Range`/`ForEachKey` for that path; keep sorted `Keys()`/`Values()` for callers that need determinism. (Also: the type doc's "Reading: Get(), Values(), Len()" should mention `Keys()`.)

cmd/fakeserver/main.go
```diff
@@ -66,7 +72,10 @@ func main() {
...
 	}
 
-	// Wait indefinitely for server stop signal
+	// Keep the compiler happy about proxyServer usage
+	_ = proxyServer
+
+	// Wait for server stop signal
 	logger.Info("Server started")
-	<-fs.stopChan
+	<-fs.StopChan
```
pavankumarinnamuri, line 76: `defer fs.Stop()` together with this `<-fs.StopChan` double-closes `StopChan`: only `Stop()` closes it, so once this unblocks the deferred `Stop()` runs again, and `Stop()` has no `sync.Once` guard → panic on the second close. Make `Stop()` idempotent or drop the defer.

cmd/fakeserver/main.go
```diff
@@ -33,28 +38,29 @@ func main() {
 		logger.Error("Token could not be signed with fake secret key")
```
pavankumarinnamuri: The `SignedString` error is logged but execution falls through and uses the (empty) token. This is unrecoverable — use `logger.Fatal(...)`.

cmd/fakeserver/main.go
```diff
@@ -66,7 +72,10 @@ func main() {
 		}()
 	}
 
-	// Wait indefinitely for server stop signal
+	// Keep the compiler happy about proxyServer usage
+	_ = proxyServer
```
pavankumarinnamuri: `var proxyServer` + this `_ = proxyServer` "keep the compiler happy" is dead state; `proxyServer` is only used inside the `if *proxyFlag` block. Scope the variable into that block and delete the blank assign.

fakeserver/server.go
```diff
@@ -324,18 +302,36 @@ func tokenAuthStreamInterceptor(expectedToken string) grpc.StreamServerIntercept
 	}
 }
 
-func (fs *FakeServer) start() error {
-	logger = fs.logger
-	logger.Info("Starting FakeServer", zap.String("address", fs.address), zap.String("httpAddress", fs.httpAddress), zap.String("token", fs.token))
+// DisconnectConfigStream closes the current config responses channel, causing the
+// active GetConfigurationUpdates stream to end (client sees EOF). A new channel is
+// created so subsequent sends to ConfigResponses work for the next connected client.
+func (fs *FakeServer) DisconnectConfigStream() {
```
pavankumarinnamuri, line 338: `DisconnectConfigStream` closes `ConfigResponses` and reassigns it with no synchronization, while `GetConfigurationUpdates` ranges over the field and tests send to it → data race on the field + send-on-closed-channel panic; fails under `-race`. Guard with a mutex, or hand each stream its own channel instead of reassigning the shared field.

fakeserver/server.go
```diff
@@ -354,38 +350,38 @@ func (fs *FakeServer) start() error {
 			Certificates: []tls.Certificate{creds},
 			MinVersion:   tls.VersionTLS12,
 		})
-	serverState = fs.state
-	fs.server = grpc.NewServer(grpc.Creds(credsTLS), grpc.KeepaliveEnforcementPolicy(kaep), grpc.KeepaliveParams(kasp), grpc.StreamInterceptor(tokenAuthStreamInterceptor(fs.token)))
-	pb.RegisterKubernetesInfoServiceServer(fs.server, &server{})
+	serverState = fs.State
```
pavankumarinnamuri: Now that this PR makes the package importable and adds `:0` (OS-assigned port) support, handlers still read package-global `serverState`/`logger` that `Start()` overwrites on every call — two servers in one process clobber each other. Handlers already have the `fs *FakeServer` receiver; use `fs.State`/`fs.Logger` and drop the globals.

internal/controller/reconciler/setup_envtest_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	t.Cleanup(cancel)
+
+	// Start resources stream client (watches envtest K8s API → populates runtime cache → streams to fakeserver)
+	resourcesFactory := &resources.Factory{
+		Logger:       zap.NewNop(),
+		K8sClient:    testClient,
+		Stats:        stream.NewStats(),
+		Cache: runtimeCache,
```
pavankumarinnamuri: Not gofmt-clean — this `Cache:` field is misaligned in the `resources.Factory` literal (`gofmt -l` flags the file). CI lint will fail; run `gofmt -w`.

internal/controller/reconciler/setup_envtest_test.go
```diff
@@ -0,0 +1,156 @@
...
+	m.Run()
+}
+
+func testdataDir() string {
+	_, filename, _, _ := runtime.Caller(0)
+
+	return filepath.Join(filepath.Dir(filename), "testdata")
+}
+
+func strPtr(s string) *string {
```
pavankumarinnamuri: `strPtr` reimplements `k8s.io/utils/ptr.To`, already in the module graph (and `watcher_test.go` has yet another variant). Delete `strPtr`; use `ptr.To("...")`.

pavankumarinnamuri (changes requested):
Nice work, great coverage! A few things to address before merge — left some comments .

.github/workflows/integration_envtest_tests.yml
byjustina: Not K8s CLI tools — this is specifically for `setup-envtest`, which downloads lightweight test binaries (kube-apiserver + etcd) used by the envtest framework to run integration tests against a real API server without needing a full cluster. These are test-only binaries

internal/convert/policy.go
byjustina: I think this isn't a concern because:

1) SSA re-apply of identical desired state is a server-side no-op. Only operator-owned fields are compared by the API server, so in this case third-party labels won't be applied and resource version won't be bumped up. I verified this with a convergence test as well, so there wouldn't be a network hot loop.
2) The cache guard only gets triggered when the object actually changed (`!exists || !proto.Equal(old, obj)`), so even if a redundant watch event did arrive, the rebuilt runtime object is proto.Equal to the cached one and the notification is suppressed — it never reaches the reconciler. So I don't believe there there would be a local hot loop either.

So the worst case is bounded: one redundant no-op apply per genuine external metadata edit, plus one no-op apply per 5-min timer tick. But these are good test cases that I will be adding 👍 

  ▎ On the suggestion to restrict runtime-side labels to operator-owned keys before caching
I kept the full proto.Equal comparison for correctness. I believe selective comparison/filtering won't be able both ignore third-party metadata and detect when CloudSecure un-declares a key it previously owned.

internal/controller/reconciler/reconciler.go
byjustina: If we add an integration test for this, aren't we essentially testing out the timer functionality which we can assume works? There are existing `reconcileAll` unit tests to verify the behavior

.github/workflows/integration_envtest_tests.yml
rigdomin-illumio, line 1: Should we configure `timeout-minutes` in case some of these integration tests hang?

fakeserver/state.go
rigdomin-illumio, line 1: Let's look to add some locked accessor methods that use Mutex Locks.
Within your fakeserver `server.go` and `helper_tests.go`, there are methods that are accessing and setting some of these fields.

e.g.
- IncrementResourcesReceived()
- CheckAndClearBadInitialCommit()
- SetConnectionSuccessful()

This has potential of race conditions on ServerState fields in your tests.

internal/controller/reconciler/setup_envtest_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+var cnpGVR = schema.GroupVersionResource{Group: "cilium.io", Version: "v2", Resource: "ciliumnetworkpolicies"}
+var ccnpGVR = schema.GroupVersionResource{Group: "cilium.io", Version: "v2", Resource: "ciliumclusterwidenetworkpolicies"}
+var cidrGroupGVR = schema.GroupVersionResource{Group: "cilium.io", Version: "v2", Resource: "ciliumcidrgroups"}
+
+func TestMain(m *testing.M) {
+	// Set KUBEBUILDER_ASSETS if not already set, so tests work from IDEs
+	// without needing to manually export the variable.
+	if os.Getenv("KUBEBUILDER_ASSETS") == "" {
+		out, err := exec.Command("setup-envtest", "use", "--print", "path").Output()
```
rigdomin-illumio: Nit:
We should consider pinning k8s version here and github actions.
This could cause a different k8s version run between different runs.
Or at least clearly state what version is used for the run.

rigdomin-illumio (dismissed):
PR looks good.
Some minor comments/nits.

Otherwise I think we're pretty much ready to merge IMO.

rigdomin-illumio (dismissed):
LGTM.

rigdomin-illumio:
Fix merge conflicts

pavankumarinnamuri:
@copilot Can you do one more round of review?

pavankumarinnamuri:
@copilot Resolve the conflicts.

rigdomin-illumio:
@byjustina , please fix failing IT build

rigdomin-illumio (dismissed):

fakeserver/state.go
```diff
@@ -119,6 +119,98 @@ func (s *ServerState) RecordResourceSnapshot() {
...
+
+		return true
+	}
+
+	return false
+}
+
+// SetConnectionSuccessful marks the connection as successful and the resource
+// snapshot as complete.
+func (s *ServerState) SetConnectionSuccessful() {
```
rlenglet: That name is too vague. What connection? If this is about the resource stream, rename this to reflect that.

fakeserver/test_helpers.go
```diff
@@ -19,16 +25,22 @@ type TestConfig struct {
 	Timeout       time.Duration
 	PollInterval  time.Duration
 	EnableLogging bool
+	// AutoHandshake controls whether Start() sends the default config handshake
+	// (UpdateConfiguration + empty ResourceSnapshotComplete). Set to false when
+	// tests need to control the handshake sequence themselves.
+	AutoHandshake bool
```
rlenglet: Where does the term "handshake" come from? AFAIK, we don't use that anywhere else in the codebase.
IMHO, "AutomaticInitialConfigurationSnapshot" or something like that would clarify the semantics.

internal/controller/reconciler/reconciler.go
```diff
@@ -21,21 +21,19 @@ import (
 )
 
 const (
-	// FieldManager identifies cloud-operator as the owner of fields in Server-Side Apply.
-	FieldManager = "illumio-cloud-operator"
-
-	// FullReconcileInterval is the periodic safety net for full reconciliation,
+	// DefaultReconcileInterval is the periodic safety net for full reconciliation,
```
rlenglet:
```suggestion
	// DefaultReconcileInterval is the default interval between periodic full reconciliations done for safety,
```

internal/controller/stream/config/cache/cache.go
rlenglet: Also, if we really need to iterate over the sorted keys very often, we could maintain an additional string slice field in addition to `objects` to stores the keys already sorted.
Inserting a new key into that slice on insert, and removing a key on delete, while keeping that slice sorted is straightforward.

If this is used only for sorting (I didn't verify that), then I don't really care about the performance.

internal/controller/stream/config/client.go
```diff
@@ -86,7 +88,21 @@ func (c *configClient) handleConfigUpdate(ctx context.Context, resp *pb.GetConfi
...
+		key, err := convert.CacheKeyForObject(update.ResourceData)
+		if err != nil {
+			c.logger.Warn("Skipping unsupported resource type",
+				zap.String("name", update.ResourceData.GetName()),
+				zap.Error(err))
+
+			return nil
+		}
+
+		cloudSecureIDToKey[update.ResourceData.GetId()] = key
```
rlenglet: Remove this `cloudSecureIDToKey` map. It's a hack to work around an RPC message definition bug. Let's fix the RPC message instead.

internal/controller/stream/config/client.go
```diff
@@ -114,7 +130,39 @@ func (c *configClient) handleConfigUpdate(ctx context.Context, resp *pb.GetConfi
...
+			var ok bool
+
+			key, ok = cloudSecureIDToKey[m.DeleteObject.GetId()]
+			if !ok {
+				c.logger.Warn("Ignoring delete for unknown CloudSecure Id", zap.String("id", m.DeleteObject.GetId()))
+
+				return nil
+			}
+
+			delete(cloudSecureIDToKey, m.DeleteObject.GetId())
```
rlenglet: Instead of doing all this dance with an extra `cloudSecureIDToKey` map, etc., let's just change the `DeleteConfiguredKubernetesObject` message definition from:
https://github.com/illumio/cloud-operator/blob/e2f083eb56c4e26360b934a098bbb95e13cbccec/api/illumio/cloud/k8sclustersync/v1/k8s_info.proto#L801-L807
```
message DeleteConfiguredKubernetesObject {
  string id = 1;
}
```
to:
```
message DeleteConfiguredKubernetesObject {
  string kind = 1;
  optional string namespace = 2;
  string name = 3;
}
```
so we have all the data we need here.

The RPC spec must meet the needs of the client. Not the of the server.
If the RPC doesn't fit how the client uses the data, fix the RPC instead of doing workarounds like this.

rlenglet (changes requested):

fakeserver/test_helpers.go
byjustina: It was generated by Claude, but meant to mean the initial config snapshot (`UpdateConfiguration` + empty `ResourceSnapshotComplete`). I'll rename it to `AutoInitialConfigSnapshot`

internal/controller/stream/config/client.go
rigdomin-illumio: Please create a small PR to update the proto schema on server side

internal/controller/reconciler/setup_envtest_test.go
```diff
@@ -0,0 +1,165 @@
...
+		os.Setenv("KUBEBUILDER_ASSETS", strings.TrimSpace(string(out)))
+	}
+
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths:     []string{filepath.Join(testdataDir(), "crds")},
+		ErrorIfCRDPathMissing: true,
+	}
+
+	cfg, err := testEnv.Start()
+	if err != nil {
```
rigdomin-illumio, line 72: This will always return 0 even if underlying test execution of `m.Run()` has failed.

```
code := m.Run()
testEnv.Stop()
os.Exit(code)
```

rigdomin-illumio (dismissed):
LGTM
Update your PR with main to fix merge conflicts

rigdomin-illumio:
Please fix the panic encountered in Unit Tests

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -780,8 +780,6 @@ message GetConfigurationUpdatesResponse {
 
 // Configured object data for snapshot or create/update operations.
 message ConfiguredKubernetesObjectData {
-  // Unique identifier for tracking this object.
-  string id = 1;
```
sungkyunchung, line 988: Can we set 1 as reserved to prevent from reusing it

sungkyunchung (changes requested):

rigdomin-illumio (dismissed):

rigdomin-illumio:
PR is good to me, but please get a clean build for this.

rigdomin-illumio (dismissed):

rigdomin-illumio:
Need to update this PR with latest changes in Main.

PR #409, opened 2026-05-05
Add reconciliation loop and runtime state cache

byjustina:

PR #423, opened 2026-06-01
Merge configuration create and update mutations into one

rlenglet:
The desired state cache and the reconciliation loop implemented in https://github.com/illumio/cloud-operator/pull/409 don't handle create and update mutations differently.
Simplify the API and the backend by merging both mutation messages into one.

### DIFF
--- a/api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
+++ b/api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
@@ -756,7 +756,7 @@ message GetConfigurationUpdatesRequest {
 // Configured Kubernetes objects:
 //   1. (0+) resource_data, one message for each object in the initial snapshot.
 //   2. (1) resource_snapshot_complete, indicates the initial snapshot is complete.
-//   3. (0+) resource_mutation, one message for each object create/update/delete.
+//   3. (0+) resource_mutation, one message for each object create-or-update or delete.
 //
 // On stream reconnection, the server sends a new full snapshot of configured objects.
 message GetConfigurationUpdatesResponse {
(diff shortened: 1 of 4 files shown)


api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -810,13 +810,13 @@ message ConfiguredKubernetesObjectSnapshotComplete {}
...
 message ConfiguredKubernetesObjectMutation {
   oneof mutation {
-    // A new configured object to create in the cluster.
-    ConfiguredKubernetesObjectData create_object = 1;
-    // An existing configured object to update in the cluster.
-    ConfiguredKubernetesObjectData update_object = 2;
+    // A configured object to create or update in the cluster.
+    ConfiguredKubernetesObjectData create_or_update_object = 1;
     // An existing configured object to delete from the cluster.
     DeleteConfiguredKubernetesObject delete_object = 3;
```
rigdomin-illumio, line 816: nit:
change to 2?

rigdomin-illumio (dismissed):
LGTM

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
rlenglet: Normally, tags should not be reused, unless the semantics is identical to the previous field. That's not the case here.
https://protobuf.dev/programming-guides/proto3/#deleting

internal/controller/stream/config/client.go
```diff
@@ -140,18 +140,11 @@ func (c *configClient) handleUpdateConfiguration(config *pb.GetConfigurationUpda
 // handleMutation processes a configured object mutation (create/update/delete).
 func (c *configClient) handleMutation(mutation *pb.ConfiguredKubernetesObjectMutation) error {
 	switch m := mutation.GetMutation().(type) {
-	case *pb.ConfiguredKubernetesObjectMutation_CreateObject:
-		c.cache.Insert(m.CreateObject.GetId(), m.CreateObject)
+	case *pb.ConfiguredKubernetesObjectMutation_CreateOrUpdateObject:
+		c.cache.Insert(m.CreateOrUpdateObject.GetId(), m.CreateOrUpdateObject)
```
byjustina: I think this should be updated to include the error handling from this [PR](https://github.com/illumio/cloud-operator/pull/424/changes#diff-78303f75f743dbc4fffb5f59f2cbf5b8e3a5051e43a6f2d0c688e48722e15959)

rlenglet: Agreed. I was going to do that after #424 was merged. Thanks!

rigdomin-illumio (approved):

PR #435, opened 2026-06-25
Add support for eks vpc cni policy

wonjun-illumio:
**Adds full ingest and enforce support for AWS VPC CNI network-policy CRDs (`networking.k8s.aws/v1alpha1`) on EKS, streamed to and reconciled from CloudSecure alongside existing core and Cilium resources:**

- `ClusterNetworkPolicy` — cluster-scoped admin/baseline policy (tier, priority, Accept/Deny/Pass actions, FQDN egress)
- `ApplicationNetworkPolicy` — namespaced policy with standard Kubernetes NetworkPolicy semantics plus FQDN (`domainNames`) egress

### Changes

**Resource discovery & routing**

- `resources.go`: added `clusternetworkpolicies` and `applicationnetworkpolicies` to `ManagedResourceNames` 
- `client.go`: dispatch routes both kinds to the AWS converter via `IsAWSResource`; existing NotFound/Forbidden skip logic preserved so missing CRDs / RBAC denials on non-EKS clusters are gracefully skipped.

**Conversion (ingest: unstructured → protobuf)**

- `convert/aws.go`, `convert/aws_types.go`: converters and local CRD struct copies for both kinds (no new module dependencies).
  - `ClusterNetworkPolicy`: cluster-scoped; `subject` (pods/namespaces), `tier`, `priority`, ingress/egress rules with per-rule `action`, ports (`portNumber`/`portRange`/`namedPort`), and egress peers carrying `networks` (CIDR) + `domainNames` (FQDN).
  - `ApplicationNetworkPolicy`: namespaced; reuses upstream `networking/v1` ingress types and the existing standard-NetworkPolicy port/ipBlock helpers, with an AWS-specific egress rule/peer to carry `domainNames`.
  - Strict parsing with explicit wrapped errors (no silent swallowing).

**Enforce (protobuf → CRD, Server-Side Apply)**

- `convert/policy.go`: added both kinds to `ExtractResourceName`, `setConfiguredKindSpecific`, and `marshalConfiguredObjectSpecs`.
  - `ClusterNetworkPolicy` marshals directly via `protoToMap` (proto mirrors the CRD spec 1:1).
  - `ApplicationNetworkPolicy` uses a dedicated spec marshaler: reconstructs `policyTypes` from the ingress/egress bools, renders peers under `from`/`to`, and emits numeric ports as JSON ints (the CRD field is `IntOrString`) so they aren't misread as named ports.

**Proto (`k8s_info.proto`)**

- New messages: `KubernetesAWSClusterNetworkPolicyData` (+ `AWSNetworkPolicySubject`, `AWSNetworkPolicyPodSelector`, `AWSNetworkPolicyIngressRule`, `AWSNetworkPolicyEgressRule`, `AWSNetworkPolicyPort`, `AWSPortNumber`, `AWSPortRange`, `AWSNetworkPolicyIngressPeer`, `AWSNetworkPolicyEgressPeer`) and `KubernetesAWSApplicationNetworkPolicyData` (+ `AWSApplicationNetworkPolicyRule`, `AWSApplicationNetworkPolicyPeer`).
- Extended the `kind_specific` oneofs (no renumbering/removal — wire-compatible, `buf breaking` passes):
  - `KubernetesObjectData`: `aws_cluster_network_policy = 107`, `aws_application_network_policy = 108`
  - `ConfiguredKubernetesObjectData`: `aws_cluster_network_policy = 103`, `aws_application_network_policy = 104`
- Regenerated `k8s_info.pb.go`

**RBAC (`clusterrole.yaml`)**

- `networking.k8s.aws`: `clusternetworkpolicies`, `applicationnetworkpolicies` — granted `get/list/watch/create/update/patch/delete` (reconciler needs write verbs for enforce).

### Tests

- Converter unit tests covering objects and edge cases for both kinds: pod/namespace subjects, `portNumber`/`portRange`/`namedPort`, Accept/Deny/Pass actions, CIDR + FQDN egress, `matchExpressions`, nil, and unsupported-kind paths.
- Enforce/apply tests assert correct kind, resource name, and rendered CRD spec (camelCase, nested `portNumber` as int, `from`/`to`, `policyTypes` reconstruction, `domainNames`).
- Reconciler tests cover apply-when-different and delete-orphaned for both kinds.
- Resource-wiring tests assert both names are in `resourceList` / `ManagedResourceNames` and routed to the AWS converter

### Notes

- contains temporary logging message, will be deleted in official PR
- Manual EKS validation pending


internal/convert/policy.go
```diff
@@ -208,6 +231,192 @@ func marshalConfiguredObjectSpecs(data *pb.ConfiguredKubernetesObjectData) (stri
 	return kind, resourceName, specFields, nil
 }
 
+// marshalAWSApplicationNetworkPolicySpec builds the ApplicationNetworkPolicy CRD
+// spec map from the proto. The proto mirrors KubernetesNetworkPolicyData (ingress/
+// egress bools + rule arrays), which does not map 1:1 to the CRD, so the spec is
+// constructed explicitly rather than via protoToMap.
```
sungkyunchung: I don't see a good advantage of doing hard-coded marshal with unmatched proto from the CRD? Can we fix the proto for ApplicationNetworkPolicy to match the CRD so we can simply call protoToMap?

internal/convert/aws.go
```diff
@@ -0,0 +1,365 @@
...
+			ingressEnabled = true
+		case "Egress":
+			egressEnabled = true
+		}
+	}
+
+	var podSelector *pb.LabelSelector
+	if len(spec.PodSelector.MatchLabels) > 0 || len(spec.PodSelector.MatchExpressions) > 0 {
+		podSelector = convertMetaV1LabelSelector(&spec.PodSelector)
+	}
```
sungkyunchung: I don't think we need this check. `podSelector: {}` means selecting all pods. Let's pass down this empty selector instead of nil to keep this meaningful.

internal/convert/policy.go
```diff
@@ -197,6 +206,20 @@ func marshalConfiguredObjectSpecs(data *pb.ConfiguredKubernetesObjectData) (stri
 			specFields = map[string]any{"spec": specMap}
 		}
 
+	case *pb.ConfiguredKubernetesObjectData_AwsClusterNetworkPolicy:
+		kind = "ClusterNetworkPolicy"
+
+		var specMap map[string]any
+
+		specMap, err = protoToMap(ks.AwsClusterNetworkPolicy)
```
sungkyunchung, line 212: protojson marshal with `EmitUnpopulated: false` will omit `priority` field if it's set to zero. We can set `priority` to zero if it's missing from the map after marshal, which sounds a bit hacky... thoughts?

sungkyunchung (changes requested):
Nice work on the proto design, full coverage of both EKS VPC CNI policy types. One thing worth revisiting is `KubernetesAWSApplicationNetworkPolicyData` models the spec slightly different from the CRD, which results in custom marshaling codes instead of simple `protoToMap` call. Also, please add a few more edge cases for conversion tests. Thanks for getting this started!

internal/convert/policy.go
wonjun-illumio: Yea, I will fix the proto for ApplicationNetworkPolicy to use protoToMap. One question before editing proto..

The ANP CRD is asymmetric between ingress and egress:
- ingress uses upstream networkingv1.NetworkPolicyIngressRule -> peers are podSelector, namespaceSelector, and ipBlock (no domainNames)
- egress uses AWS's own ApplicationNetworkPolicyEgressRule -> peers include domainNames along with previous 3 fields

So the only real difference between an ingress peer and an egress peer is domainNames. Two ways I can model this in proto:

Symmetric (one peer type) - a single AWSApplicationNetworkPolicyPeer with podSelector, namespaceSelector, ipBlock, and domainNames, used for both from and to.

Asymmetric (mirror the CRD) — separate ingress peer (no domainNames) and egress peer (with domainNames), like the CRD.

I prefer A (symmetric) because the asymmetry is a single field, and on ingest we read from upstream NetworkPolicyIngressRule which has no domain field, so we structurally can't populate it there anyway. Both symmetric and asymmetric will require port normalizer at the end of protoToMap, since ANP CRD's port field is IntOrString.

internal/convert/policy.go
wonjun-illumio: sounds good to me. I don't see other way of doing this cleanly

internal/convert/aws.go
```diff
@@ -0,0 +1,365 @@
...
+
+	return out
+}
+
+func convertAWSApplicationNetworkPolicy(anp *awsApplicationNetworkPolicy) *pb.KubernetesAWSApplicationNetworkPolicyData {
+	spec := anp.Spec
+
+	var ingressEnabled, egressEnabled bool
+
+	for _, policyType := range spec.PolicyTypes {
```
rigdomin-illumio: Log unknown types just in case.

internal/convert/policy.go
```diff
@@ -208,6 +231,192 @@ func marshalConfiguredObjectSpecs(data *pb.ConfiguredKubernetesObjectData) (stri
...
+	case pb.Port_PROTOCOL_TCP_UNSPECIFIED:
+		return "TCP"
+	default:
+		return "TCP"
+	}
+}
+
+// awsLabelSelectorToMap renders a proto LabelSelector as a CRD selector map, or
+// nil when the selector is empty.
+func awsLabelSelectorToMap(sel *pb.LabelSelector) map[string]any {
```
rigdomin-illumio: For AWS NetworkPolicy and ApplicationNetworkPolicy, 
`podSelector: {}` should be support, as it represents `match everything`

Currently, empty pod selection above will result in `nil` pod selection to be sent to CS.
This affects pod selector and peer selectors.

e.g.
we want:
```
spec:
  podSelector: {}       # every pod in the namespace
  policyTypes: ["Ingress", "Egress"]
```

but we get:
```
spec:
  policyTypes: ["Ingress", "Egress"]
  # podSelector is MISSING
```

rigdomin-illumio (changes requested):
Overall PR looks great!
One concern is the support for empty pod selector.

rigdomin-illumio:
General comment:
We may want to consider versioning compatibility checks:
e.g.
`v1alpha1 CRD requires VPC CNI addon v1.15+ and EKS Kubernetes 1.25+ (GA on 1.28+)`

internal/convert/policy.go
sungkyunchung: That's reasonable, but I'd prefer asymmetric to mirror the CRD as closely as possible. It's easier to track future CRD updates. On the ingest side it's safe, but on the enforcement side we don't want to accidentally set the field and have the API server reject the apply.

pavankumarinnamuri:
woo ho @wonjun-illumio — first PR! 🎉

pavankumarinnamuri (changes requested):
Thanks for the PR — the proto design and overall structure look solid, and test coverage is good. Two issues need to be fixed before merge:

1. **Empty selector handling in ApplicationNetworkPolicy** — empty  and empty peer selectors mean "select all" in the CRD, but the conversion and enforce marshaler currently drop them. This produces invalid CRDs or changes semantics. See comments on  and .

2. **ClusterNetworkPolicy  is dropped** —  is a required CRD field that allows , but the proto uses plain , so  omits it. Please make it  so explicit zero values are preserved.

Also added a few smaller notes: the ANP proto/CRD mismatch that forces custom marshaling, the widened  RBAC for the flow-logging CNP, and a duplicate label-selector helper. Please consider adding round-trip and empty-selector test cases as well.

pavankumarinnamuri (changes requested):
.

internal/convert/aws.go
```diff
@@ -0,0 +1,365 @@
...
+		switch policyType {
+		case "Ingress":
+			ingressEnabled = true
+		case "Egress":
+			egressEnabled = true
+		}
+	}
+
+	var podSelector *pb.LabelSelector
+	if len(spec.PodSelector.MatchLabels) > 0 || len(spec.PodSelector.MatchExpressions) > 0 {
```
pavankumarinnamuri: Empty `podSelector` is dropped here, but in the CRD `podSelector` is a required field and an empty selector means "all pods in the namespace" (standard `LabelSelector` semantics). Converting it to `nil` and then omitting it on enforce produces an invalid CRD that fails validation. Consider always converting the selector so empty is preserved as `{}`.

internal/convert/policy.go
pavankumarinnamuri: Returning `nil` for an empty selector drops the field entirely. For ANP peers this changes semantics: an empty `podSelector`/`namespaceSelector` means "select all pods/namespaces", and omitting it can leave a peer with no selector set, violating the CRD's mutual-exclusion rules. Empty selectors should render as `{}`.

internal/convert/policy.go
```diff
@@ -208,6 +231,192 @@ func marshalConfiguredObjectSpecs(data *pb.ConfiguredKubernetesObjectData) (stri
 	return kind, resourceName, specFields, nil
 }
 
+// marshalAWSApplicationNetworkPolicySpec builds the ApplicationNetworkPolicy CRD
+// spec map from the proto. The proto mirrors KubernetesNetworkPolicyData (ingress/
```
pavankumarinnamuri: The custom marshaling is necessary because the proto models ANP like the existing standard `KubernetesNetworkPolicyData` (ingress/egress bools + rule arrays) rather than mirroring the CRD (podSelector + policyTypes + ingress/egress rule arrays). This is consistent with existing patterns, but it does introduce round-trip complexity. Consider adding a test that round-trips an ANP through `ConvertUnstructuredToAWSResource` → `ConvertToApplyObject` to catch selector/field-loss bugs.

internal/convert/policy.go
```diff
@@ -208,6 +231,192 @@ func marshalConfiguredObjectSpecs(data *pb.ConfiguredKubernetesObjectData) (stri
...
+// egress bools + rule arrays), which does not map 1:1 to the CRD, so the spec is
+// constructed explicitly rather than via protoToMap.
+func marshalAWSApplicationNetworkPolicySpec(data *pb.KubernetesAWSApplicationNetworkPolicyData) map[string]any {
+	spec := map[string]any{}
+
+	if sel := awsLabelSelectorToMap(data.GetPodSelector()); sel != nil {
```
pavankumarinnamuri: Because `awsLabelSelectorToMap` returns `nil` for an empty selector, an empty `podSelector` is omitted from the enforced CRD. The CRD requires `podSelector`, so the apply will fail. Empty selectors should be emitted as `{}`.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -354,6 +356,162 @@ message CiliumPolicyPort {
...
+// EKS. The operator ingests customer-authored policies and applies CloudSecure
+// authored ones. Field names match the CRD so protojson (camelCase) output maps
+// directly. Schema pinned to:
+// https://github.com/aws/amazon-network-policy-controller-k8s/blob/59b675390bf78a3303e4e60be71a5d057845e61e/api/v1alpha1/clusternetworkpolicy_types.go
+
+// KubernetesAWSClusterNetworkPolicyData contains data specific to an AWS
+// ClusterNetworkPolicy object.
+message KubernetesAWSClusterNetworkPolicyData {
+  // Evaluation order within a tier (0-1000); lower is higher precedence.
+  int32 priority = 1;
```
pavankumarinnamuri: `priority` should be `optional int32`. The CRD requires priority and allows `0`, but `protoToMap` uses `EmitUnpopulated=false`, so a priority of `0` is dropped on enforce and the SSA apply fails validation. Making it proto3 `optional` preserves explicit zero values.

cloud-operator/templates/clusterrole.yaml
```diff
@@ -22,10 +22,11 @@ rules:
...
   resources: ["ciliumcidrgroups", "ciliumnetworkpolicies", "ciliumclusterwidenetworkpolicies"]
   verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
-# AWS VPC CNI - operator creates ClusterNetworkPolicy for flow logging when VPC CNI is detected
+# AWS VPC CNI - operator creates a ClusterNetworkPolicy for flow logging and
+# ingests/reconciles ClusterNetworkPolicy resources for policy support.
 - apiGroups: ["networking.k8s.aws"]
-  resources: ["clusternetworkpolicies"]
-  verbs: ["create"]
+  resources: ["clusternetworkpolicies", "applicationnetworkpolicies"]
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
```
pavankumarinnamuri, line 30: Granting `delete` here is correct for reconciler-managed policies, but note that the operator also creates a flow-logging `ClusterNetworkPolicy` that is not intended to be reconciled. Today it is protected by the absence of the `cloud.illum.io/resource-id` label in the runtime-cache handler, but the widened RBAC scope increases risk if that guard is ever bypassed. Consider documenting this invariant or scoping the flow-logging CNP creation to a separate, non-reconciled path.

internal/convert/aws.go
```diff
@@ -0,0 +1,365 @@
...
+
+// convertAWSANPEgressPeers converts ApplicationNetworkPolicy egress peers, which
+// additionally carry FQDN domain names.
+func convertAWSANPEgressPeers(peers []awsANPEgressPeer) []*pb.AWSApplicationNetworkPolicyPeer {
+	if len(peers) == 0 {
+		return nil
+	}
+
+	out := make([]*pb.AWSApplicationNetworkPolicyPeer, 0, len(peers))
+	for _, p := range peers {
```
pavankumarinnamuri, line 346: This helper duplicates `convertLabelSelectorToProto` in `k8s_resources.go`. Consider reusing the existing helper to reduce duplication.

pavankumarinnamuri (changes requested):

wonjun-illumio:
Changes made according to review:
1. remodel proto design for `ApplicationNetworkPolicy` to exactly mirror from AWS crd
2. correct empty pod selectors - should be all fixed
3. changed `ClusterNetworkPolicy's` priority to optional, which should also fix protojson marshal with `EmitUnpopulated: false` omitting priority when set to 0
4. log unknown `ApplicationNetworkPolicy` policy types
5. Deleted the local `convertMetaV1LabelSelector` and reused the shared `convertLabelSelectorToProto`
6. RBAC invariant documentation + additional guard check
7. more testing



> General comment: We may want to consider versioning compatibility checks: e.g. `v1alpha1 CRD requires VPC CNI addon v1.15+ and EKS Kubernetes 1.25+ (GA on 1.28+)`

How do we want this to be achieved exactly? There is VPC CNI CRD check within detect.go:130-137. Not sure if this is enough or what is being asked.
```
  if awsvpccni.IsCRDAvailable(config.Logger, discoveryClient) {
      if err := awsvpccni.EnsureFlowLoggingPolicy(...); err != nil {
          config.Logger.Warn("... flow logging may be limited", ...)
      }
  } else {
      config.Logger.Warn("ClusterNetworkPolicy CRD not found - enable network policy in AWS VPC CNI addon ...")
  }
```

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -354,6 +357,193 @@ message CiliumPolicyPort {
...
+// Egress peers additionally allow CIDR networks and FQDN domain names.
+message AWSNetworkPolicyEgressPeer {
+  // Selects pods by namespace and pod labels.
+  AWSNetworkPolicyPodSelector pods = 1;
+  // Selects entire namespaces by label.
+  LabelSelector namespaces = 2;
+  // Networks lists CIDR blocks (max 25).
+  repeated string networks = 3;
+  // DomainNames lists FQDN peers, wildcard-capable (max 25); valid only with
+  // "Accept" or "Pass" actions.
```
sungkyunchung: It seems CNP does not support domain names with "Pass" anymore. Please fix the comment. https://github.com/aws/amazon-network-policy-controller-k8s/commit/82547210cc7a354ec015c1751883ad047617441f

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -354,6 +356,162 @@ message CiliumPolicyPort {
...
+// AWSNetworkPolicyPort selects a port; exactly one form is set, mirroring the
+// CRD's port {portNumber | portRange | namedPort}.
+message AWSNetworkPolicyPort {
+  AWSPortNumber port_number = 1;
+  AWSPortRange port_range = 2;
+  optional string named_port = 3;
+}
+
+// AWSPortNumber is a single port number with its protocol.
+message AWSPortNumber {
```
sungkyunchung: Can we keep the `AWSNetworkPolicy*` prefix for all CNP sub fields

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -354,6 +356,162 @@ message CiliumPolicyPort {
...
+  // IP protocol: "TCP", "UDP", or "SCTP" (default "TCP").
+  string protocol = 1;
+  // Port number (1-65535).
+  int32 port = 2;
+}
+
+// AWSPortRange is an inclusive port range with its protocol.
+message AWSPortRange {
```
sungkyunchung: Can we keep the AWSNetworkPolicy* prefix for all CNP sub fields

sungkyunchung (changes requested):
Thanks for addressing all the comments! The conversion code looks much cleaner now. LGTM other than some nit comments.

internal/convert/policy.go
```diff
@@ -28,8 +28,24 @@ const (
...
+
+	// FlowLoggingComponentValue marks the operator-created flow-logging
+	// ClusterNetworkPolicy.
+	FlowLoggingComponentValue = "flow-logging"
 )
 
+// IsOperatorInfrastructure reports whether the given labels belong to an
+// operator-created infrastructure object, which must never be treated as
+// desired-state-managed, so the runtime-cache handler and reconciler skip them.
+func IsOperatorInfrastructure(labels map[string]string) bool {
```
byjustina: BTW I removed operator-managed labels checks and opted for field manager ownership check instead, in this [PR](https://github.com/illumio/cloud-operator/pull/422) which hasn't been merged yet

pavankumarinnamuri:
@copilot Can you do one more round of review?

sungkyunchung (dismissed):

rigdomin-illumio (dismissed):
LGTM

internal/convert/policy_test.go
```diff
@@ -171,6 +172,245 @@ func TestConvertToApplyObject_CiliumCIDRGroup(t *testing.T) {
...
+	require.Len(t, egress, 1)
+	eRule, ok := egress[0].(map[string]any)
+	require.True(t, ok)
+	to, ok := eRule["to"].([]any)
+	require.True(t, ok)
+	toPeer, ok := to[0].(map[string]any)
+	require.True(t, ok)
+	assert.Equal(t, []any{"*.amazonaws.com"}, toPeer["domainNames"])
+}
+
```
byjustina: Can you add a round trip test for CNP too?

internal/convert/policy.go
byjustina: I'm not sure that this is necessary -  in the protobuf you set the priority as an optional, which should generate a pointer and treat it as 0 even if not included
```
  // Optional so an explicit priority of 0 (valid, highest precedence) is
  // preserved through protojson instead of being dropped as a zero value.
  optional int32 priority = 1;
```

can you add a test to validate this behavior where the priority is unset?

byjustina (dismissed):

rigdomin-illumio (dismissed):
LGTM

rigdomin-illumio:
Update your branch with latest main to fix merge conflicts

sungkyunchung:
ANP is exclusive to EKS Auto Mode. It doesn't exist on standard EKS clusters with VPC CNI. On Auto Mode, VPC CNI runs entirely off-cluster on AWS-managed infrastructure (no aws-node DaemonSet or pods), so flow ingestion is architecturally not possible on Auto Mode. 

For the current scope, ANP ingestion adds dead code. We should limit this PR to CNP support and track ANP separately once we have a plan for Auto Mode flow ingestion.

pavankumarinnamuri:
@copilot Research & verify the claim in this [comment:](https://github.com/illumio/cloud-operator/pull/435#issuecomment-5037741817) is ApplicationNetworkPolicy (networking.k8s.aws/v1alpha1) exclusive to EKS Auto Mode and absent on standard EKS VPC CNI clusters? Please cite the amazon-network-policy-controller-k8s repo and AWS EKS Auto Mode docs.
 Also, can you do one more round of review on this PR while you're at it?

pavankumarinnamuri (dismissed):

pavankumarinnamuri:
@copilot   Can you verify this claim with sources and give a clear confirmed/refuted verdict:

  On AWS EKS, ApplicationNetworkPolicy (networking.k8s.aws/v1alpha1) FQDN/domainNames egress rules only work on EKS Auto Mode nodes — so on a standard EKS VPC CNI cluster, ANP effectively degrades to a
  regular Kubernetes NetworkPolicy with the FQDN part ignored. Also: is the ANP CRD documented/supported on standard EKS at all, or only on Auto Mode?

  Please cite the AWS EKS docs and the amazon-network-policy-controller-k8s / aws-network-policy-agent repos.

pavankumarinnamuri:
@copilot We're making ApplicationNetworkPolicy ingest-only in this PR — watch/stream it to CloudSecure but don't enforce/reconcile it. ClusterNetworkPolicy stays full ingest + enforce. Can you list the exact code changes needed to drop ANP from the enforce path while keeping ingest? Check the ConfiguredKubernetesObjectData proto oneof, policy.go (ExtractResourceName / setConfiguredKindSpecific / marshalConfiguredObjectSpecs), ManagedResourceNames vs resourceList in resources.go, the ANP RBAC verbs in clusterrole.yaml, and any dead tests.

pavankumarinnamuri:
@copilot whoa, easy there 😄 I asked for a list, not a whole PR — you went full agent mode on me. Can you roll back / reset to the commit right before the one you just pushed, and instead just drop the change list in a comment? I'll do the edits myself. 🙏

pavankumarinnamuri:
@copilot+kimi-k2.7-code Can you do one more round of review?

pavankumarinnamuri (dismissed):

pavankumarinnamuri:
Please don't merge this.

rigdomin-illumio (dismissed):

internal/convert/policy.go
```diff
@@ -93,6 +95,8 @@ func setConfiguredKindSpecific(configured *pb.ConfiguredKubernetesObjectData, so
 		configured.KindSpecific = &pb.ConfiguredKubernetesObjectData_CiliumClusterwideNetworkPolicy{CiliumClusterwideNetworkPolicy: ks.CiliumClusterwideNetworkPolicy}
 	case *pb.KubernetesObjectData_CiliumCidrGroup:
 		configured.KindSpecific = &pb.ConfiguredKubernetesObjectData_CiliumCidrGroup{CiliumCidrGroup: ks.CiliumCidrGroup}
+	case *pb.KubernetesObjectData_AwsClusterNetworkPolicy:
```
rigdomin-illumio, line 98: Didn't we re-introduce `AwsApplicationNetworkPolicy` support?
If so, let's add it back.

rigdomin-illumio (approved):
LGTM

pavankumarinnamuri (approved):

PR #437, opened 2026-07-06
Log Hubble Relay agent-node connectivity periodically

ganesh-talla:
Periodically call Hubble Relay's `Observer.ServerStatus` RPC, which reports its connectivity to the Cilium agents running on all nodes, and log the response along with the flow collection stat logs, to help troubleshoot the cases where the operator reports zero flows with no error.

Sample log output:

```
INFO  Hubble Relay server status   {"numConnectedNodes": 8, "numUnavailableNodes": 0, "unavailableNodes": []}
INFO  Hubble Relay server status   {"numConnectedNodes": 6, "numUnavailableNodes": 2, "unavailableNodes": ["node-a", "node-b"]}
WARN  Hubble Relay cannot reach some or all of its agent nodes; cluster network flows may be incomplete or missing   {"numConnectedNodes": 6, "numUnavailableNodes": 2, "unavailableNodes": ["node-a", "node-b"]}
WARN  Hubble Relay cannot reach some or all of its agent nodes; cluster network flows may be incomplete or missing   {"numConnectedNodes": 0, "numUnavailableNodes": 0, "unavailableNodes": []}
```

### DIFF
--- a/internal/controller/collector/hubble_watchdog.go
+++ b/internal/controller/collector/hubble_watchdog.go
@@ -1,0 +1,89 @@
+// Copyright 2026 Illumio, Inc. All Rights Reserved.
+
+package collector
+
+import (
+	"context"
+	"time"
+
+	observer "github.com/cilium/cilium/api/v1/observer"
+	"go.uber.org/zap"
+)
+
+// monitorRelayPeerHealth polls Hubble Relay's ServerStatus RPC while a flow
+// stream is active and logs how many agent nodes Relay is currently connected
+// to. This surfaces a blind spot that is otherwise invisible to the operator.
+//
+// Relay maintains its own persistent connections to the per-node Cilium agents
+// (hubble-peer on :4244) and fans their flows in to every consumer. When those
+// relay->agent connections drop and Relay does not reconnect, it keeps the
+// operator's GetFlows stream open while delivering no flows: stream.Recv()
+// blocks with no error and the operator silently logs flows_received: 0.
+//
+// The operator cannot repair the relay->agent hop: that connection is managed
+// entirely by Relay and is shared across all consumers, so reconnecting the
+// operator's own stream would not help. This watchdog is therefore observability
+// only. It logs a Warn whenever Relay reports one or more unavailable nodes
+// (naming them) so operators and support can immediately distinguish "relay
+// healthy but no traffic" from "relay cannot reach its agents", instead of
+// investigating a silent flows_received: 0. It returns when ctx is done.
+func (fm *CiliumFlowCollector) monitorRelayPeerHealth(ctx context.Context) {
+	interval := fm.relayStatusCheckInterval
+	if interval <= 0 {
+		interval = HubbleRelayStatusCheckInterval
+	}
+
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			connected, unavailable, unavailableNodes, ok := fm.queryRelayPeerStatus(ctx)
+			if !ok {
+				// The status RPC itself failed; skip this tick. A genuinely dead
+				// Relay is handled by the flow stream's own error path.
+				continue
+			}
+
+			switch {
+			case unavailable > 0:
+				fm.logger.Warn("Hubble Relay cannot reach some agent nodes; cluster network flows may be incomplete",
+					zap.Uint32("connected_nodes", connected),
+					zap.Uint32("unavailable_nodes", unavailable),
+					zap.Strings("unavailable_node_names", unavailableNodes),
+				)
+			case connected == 0:
+				fm.logger.Warn("Hubble Relay reports zero connected agent nodes; no cluster network flows will be collected until Relay reconnects to its agents",
+					zap.Uint32("connected_nodes", connected),
+				)
+			default:
+				fm.logger.Info("Hubble Relay peer status healthy",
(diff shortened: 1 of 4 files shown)


internal/controller/collector/hubble_watchdog.go
```diff
@@ -0,0 +1,89 @@
...
+// blocks with no error and the operator silently logs flows_received: 0.
+//
+// The operator cannot repair the relay->agent hop: that connection is managed
+// entirely by Relay and is shared across all consumers, so reconnecting the
+// operator's own stream would not help. This watchdog is therefore observability
+// only. It logs a Warn whenever Relay reports one or more unavailable nodes
+// (naming them) so operators and support can immediately distinguish "relay
+// healthy but no traffic" from "relay cannot reach its agents", instead of
+// investigating a silent flows_received: 0. It returns when ctx is done.
+func (fm *CiliumFlowCollector) monitorRelayPeerHealth(ctx context.Context) {
```
rlenglet: Nits.
1. The name is "Hubble Relay".
2. Why introduce new terms ("peer" and "health")??? The RPC name uses terms "server" and  "status", and the relevant fields are using term "node". Use those terms.

```suggestion
func (fm *CiliumFlowCollector) monitorHubbleRelayServerStatus(ctx context.Context) {
```

internal/controller/collector/hubble_watchdog.go
```diff
@@ -0,0 +1,89 @@
...
+				)
+			}
+		}
+	}
+}
+
+// queryRelayPeerStatus issues a single, timeout-bounded ServerStatus RPC and
+// returns the connected/unavailable node counts and unavailable node names. ok
+// is false if the status could not be determined.
+func (fm *CiliumFlowCollector) queryRelayPeerStatus(ctx context.Context) (connected, unavailable uint32, unavailableNodes []string, ok bool) {
```
rlenglet: Nits re: naming.

```suggestion
func (fm *CiliumFlowCollector) queryHubbleRelayServerStatus(ctx context.Context) (numConnectedNodes, numUnavailableNodes uint32, unavailableNodes []string, ok bool) {
```

internal/controller/collector/hubble_watchdog.go
```diff
@@ -0,0 +1,89 @@
...
+	rpcCtx, cancel := context.WithTimeout(ctx, relayStatusRPCTimeout)
+	defer cancel()
+
+	resp, err := fm.client.ServerStatus(rpcCtx, &observer.ServerStatusRequest{})
+	if err != nil {
+		fm.logger.Debug("Failed to query Hubble Relay ServerStatus", zap.Error(err))
```
rlenglet: Why only debug level? If you ignore this error it doesn't need to be at Error level, but it should at least be at Warn level.

internal/controller/collector/hubble_watchdog.go
```diff
@@ -0,0 +1,89 @@
...
+				)
+			case connected == 0:
+				fm.logger.Warn("Hubble Relay reports zero connected agent nodes; no cluster network flows will be collected until Relay reconnects to its agents",
+					zap.Uint32("connected_nodes", connected),
+				)
+			default:
+				fm.logger.Info("Hubble Relay peer status healthy",
+					zap.Uint32("connected_nodes", connected),
+				)
+			}
```
rlenglet: Nits re: naming.
1. Use the same terms as in Cilium's protobuf definitions, for consistency.
2. We use camelCase for log field names in this codebase.

```suggestion
			switch {
			case unavailable > 0:
				fm.logger.Warn("Hubble Relay cannot reach some agent nodes; cluster network flows may be incomplete",
					zap.Uint32("numConnectedNodes", connected),
					zap.Uint32("numUnavailableNodes", unavailable),
					zap.Strings("unavailableNodes", unavailableNodes),
				)
			case connected == 0:
				fm.logger.Warn("Hubble Relay reports zero connected agent nodes; no cluster network flows will be collected until Relay reconnects to its agents",
					zap.Uint32("numConnectedNodes", connected),
					zap.Uint32("numUnavailableNodes", unavailable),
				)
			default:
				fm.logger.Info("Hubble Relay server status is healthy",
					zap.Uint32("numConnectedNodes", connected),
					zap.Uint32("numUnavailableNodes", unavailable),
				)
			}
```

rlenglet: I have other issues with this:

1. I am against having 3 distinct log messages for the 3 different cases. It will make filtering the logs unnecessarily complex (to get the stats over time, I would have to write a regexp that select all 3 messages??!!). Reduce this to 2 distinct logs:
    1. One log at Info level, logged unconditionally, which includes the stats.
    2. One more log at Warn level for each error condition.
1. Logging the full list of unavailable nodes can be too long. There may be hundreds of unavailable nodes. I would sort the list and only log a subset.

rlenglet (changes requested):

internal/controller/collector/hubble_watchdog.go
ganesh-talla: The reason is if hubble relay itself is not accesible then the strem recv will error out and force a restart.

But will change the log level to warn level

internal/controller/collector/hubble_watchdog.go
ganesh-talla: Good point I added sorting and only took top10 unavailable nodes

ganesh-talla:
@rlenglet Resolved all comments Please review

rlenglet (approved):

PR #441, opened 2026-07-14
prevent ReplaceAll deadlock by making snapshot notification non-blocking

pavankumarinnamuri:
ReplaceAll sent SnapshotReplaced on the unbuffered resourceChanged channel with a blocking send, coupling the resource snapshot's completion to a consumer actively draining the channel.

The reconciler is the only consumer, and it does not start draining until both the config and runtime caches are ready. When the config cache has not yet become ready, the reconciler is not draining, so the runtime cache's ReplaceAll blocks indefinitely on the send.

While ReplaceAll is blocked the resource stream never completes its snapshot, the processing flag is never cleared, and the health check starts failing — causing the liveness probe to restart the pod in a loop with no resources ever ingested.

Make sure that sending the SnapshotReplaced notification is never blocking, by first draining all the notifications already in the resourceChanged channel.

### DIFF
--- a/internal/controller/stream/config/cache/cache.go
+++ b/internal/controller/stream/config/cache/cache.go
@@ -111,10 +111,25 @@ func (c *ObjectCache[T]) ReplaceAll(ctx context.Context, objects map[string]T) e
 
 	c.mutex.Unlock()
 
+	// Honor cancellation deterministically before attempting to notify, so a
+	// cancelled snapshot is never reported as successful.
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
+	// Notify consumers of the full replacement without blocking the caller.
+	// SnapshotReplaced is idempotent ("the whole cache was replaced, reconcile
+	// everything"), so the send is buffered by one and coalescing: if a prior
+	// SnapshotReplaced is still queued because the consumer is not draining yet
+	// (e.g. the reconciler is still waiting on the other cache to become ready,
+	// including across stream reconnects that call ReplaceAll again), dropping
+	// this one is correct and must not block. This fully decouples snapshot
+	// completion from consumer readiness and prevents a startup/reconnect
+	// deadlock that would otherwise wedge the resource stream and trip the
+	// liveness probe.
 	select {
-	case <-ctx.Done():
-		return ctx.Err()
 	case c.resourceChanged <- SnapshotReplaced:
+	default:
 	}
 
 	return nil
--- a/internal/controller/stream/config/cache/cache.go
+++ b/internal/controller/stream/config/cache/cache.go
@@ -95,7 +95,10 @@ func (c *ObjectCache[T]) ResourceChanged() <-chan string {
 // Use this for snapshot ingestion: build a local map, then call ReplaceAll to
 // swap it in. The cache remains consistent until the swap completes.
 // Also marks the cache as ready on first call to indicate cache now has valid data.
-// Sends SnapshotReplaced on the resourceChanged channel to signal a full snapshot replacement.
+// Sends SnapshotReplaced on the resourceChanged channel to signal a full snapshot
+// replacement. The notification is always delivered: if the single-slot buffer is
+// full, ReplaceAll drains the stale value and retries, ensuring SnapshotReplaced
+// is enqueued without blocking indefinitely.
 func (c *ObjectCache[T]) ReplaceAll(ctx context.Context, objects map[string]T) error {
 	c.mutex.Lock()
 
@@ -103,6 +106,9 @@ func (c *ObjectCache[T]) ReplaceAll(ctx context.Context, objects map[string]T) e
 
 	// Mark ready (idempotent - safe to call on reconnect)
 	select {
+	case <-ctx.Done():
+		c.mutex.Unlock()
+		return ctx.Err()
 	case <-c.ready:
 		// Already closed
 	default:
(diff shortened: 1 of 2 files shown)


rlenglet (dismissed):

internal/controller/stream/config/cache/cache.go
```diff
@@ -52,9 +53,14 @@ const SnapshotReplaced = "*"
...
-		ready:           make(chan struct{}),
-		resourceChanged: make(chan string),
+		objects: make(map[string]T),
+		ready:   make(chan struct{}),
+		// Buffered by one so a snapshot swap (ReplaceAll) can enqueue its
+		// SnapshotReplaced notification without blocking on a consumer that is
+		// not yet draining (e.g. the reconciler still waiting on another cache to
+		// become ready). Without the buffer the first snapshot can deadlock the
+		// resource stream, which then fails its liveness probe and crash-loops.
+		resourceChanged: make(chan string, 1),
```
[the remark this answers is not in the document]
pavankumarinnamuri: @rlenglet  Copilot is actually right. buffer of 1 isn't enough — on a stream reconnect ReplaceAll gets called again, buffer still has the old SnapshotReplaced in it (reconciler never drains it, it's stuck in waitForCaches since EO is disabled), so the second send blocks and we're back to the deadlock.

verified it live with **gops** on the pod — goroutine 24 was sitting in reconciler.waitForCaches (reconciler.go:133) for ~1152 mins with nothing draining resourceChanged, and the pod had a restart with Exit 2 from the liveness kill, which lines up exactly with the reconnect case.

fixed it properly now — kept the buffer but made the send non-blocking + coalescing (default: drop when full, SnapshotReplaced is idempotent so dropping a duplicate is fine), and moved the ctx check up so cancellation stays deterministic. added a regression test that deadlocks on the old code and passes on the new one.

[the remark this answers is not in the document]
pavankumarinnamuri:
```
 goroutine 24 [select, 1152 minutes]:
  reconciler.(*Reconciler).waitForCaches(...)
        /workspace/internal/controller/reconciler/reconciler.go:133
  reconciler.(*Reconciler).Run(...)
        /workspace/internal/controller/reconciler/reconciler.go:89
  created by main.main
        /workspace/cmd/main.go:251
```

pavankumarinnamuri:
@rlenglet following up on your "drain the channel / lock" point — implemented the drain-then-send in ReplaceAll, but I noticed a narrow race: if an Insert/Delete refills the single-slot buffer between the drain and the send, ReplaceAll hits `default` and leaves a per-ID signal queued instead of SnapshotReplaced, so the reconciler does reconcileObject(id) instead of a full reconcileAll.

It only happens when the reconciler is actively draining (not the deadlock case we're fixing), and the 5-min FullReconcileInterval self-heals it. To fully close it I'd serialize all sends (ReplaceAll + Insert + Delete) under a notifyMu so the drain+send is atomic — the lock you mentioned. That touches Insert/Delete too, not just ReplaceAll.

Do you want the notifyMu now, or is the 5-min reconcile fallback acceptable for this PR?

internal/controller/stream/config/cache/cache.go
```diff
@@ -105,10 +111,33 @@ func (c *ObjectCache[T]) ReplaceAll(ctx context.Context, objects map[string]T) e
...
+	// single-slot buffer the drain frees the slot, so the send does not block and
+	// the snapshot notification is not lost. This decouples snapshot completion
+	// from consumer readiness and prevents the startup/reconnect deadlock that
+	// would otherwise wedge the resource stream and trip the liveness probe.
+	select {
+	case <-c.resourceChanged:
+		// Drained a stale queued value.
+	default:
+		// Buffer already empty.
+	}
```
rlenglet: 1. You should check `ctx.Done()` in the same `select`. That would make the code cleaner.
2. In case we increase the size of the channel in the future, it would be good to receive/drain _all_ the values from the channel, and not just limit to one. So do that in a `for` loop.


```suggestion
	// Notify consumers of the full replacement without blocking the caller.
	// SnapshotReplaced supersedes any per-ID signal already queued (it means
	// "reconcile everything", which subsumes any single object change), so drain
	// a stale queued values first and then enqueue SnapshotReplaced.
	// This decouples snapshot completion from consumer readiness and prevents
	// the startup/reconnect deadlock that would otherwise wedge the resource
	// stream and trip the liveness probe.

Drain:
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case _, ok := <-c.resourceChanged:
			// Drained a stale queued value.
			if !ok {
				break Drain
			}
		default:
			// Channel is empty.
			break Drain
		}
	}
```

internal/controller/stream/config/cache/cache.go
```diff
@@ -105,10 +111,33 @@ func (c *ObjectCache[T]) ReplaceAll(ctx context.Context, objects map[string]T) e
...
+
 	select {
-	case <-ctx.Done():
-		return ctx.Err()
```
rlenglet, line 110: Restore those lines.
Every `select` must start with this case, to be safe.

internal/controller/stream/config/cache/cache.go
```diff
@@ -105,10 +111,33 @@ func (c *ObjectCache[T]) ReplaceAll(ctx context.Context, objects map[string]T) e
...
 	case c.resourceChanged <- SnapshotReplaced:
+	default:
+		// A concurrent Insert/Delete refilled the slot between the drain and this
+		// send. That signal already tells the consumer there is work to do, so
+		// skipping keeps ReplaceAll non-blocking without losing the wakeup.
```
rlenglet: It is impossible for any other value to be sent into the channel since the lock is still held at this point.
Remove this case and this comment.
The channel getting drained just above while holding the lock should do the trick already.
We shouldn't take the risk of not sending this `SnapshotReplaced` into the channel.
```suggestion
```

rlenglet (changes requested):

internal/controller/stream/config/cache/cache.go
pavankumarinnamuri: Lock isn't actually held at this point — the send is after c.mutex.Unlock(), and Insert/Delete send unlocked too. I'll serialize all three under the lock so the drain + guaranteed send holds, then remove the default:

rlenglet: Ah good point. I misread the Unlock above as being in a defer.

rlenglet: But then the problem is that the other sends are not Locked either?! So we can never be sure that the channel is really drained here if other goroutines can send into the channel while we’re holding the lock here.

We may need to rethink this whole solution. Especially: why are we calling ReplaceAll while no go routine is listening on the channel?! That is the core of the problem. The cache must never be modified while no goroutine is listening on the channel.

rlenglet: I found how to solve this safely **after** unlocking:
```
Send:
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case c.resourceChanged <- SnapshotReplaced:
			break Send
		case <-c.resourceChanged:
			// Received a stale queued value from the full channel. Retry sending.
		}
	}
```

pavankumarinnamuri: @copilot can you implement this in internal/controller/stream/config/cache/cache.go, in the ReplaceAll method?

We're fixing the startup/reconnect deadlock. resourceChanged is a buffered channel of size 1, and right now the notification uses a non-blocking send with a default: case that can silently drop the SnapshotReplaced signal. I want to guarantee SnapshotReplaced is always delivered, never block indefinitely, and still honor context cancellation — without adding any new mutex and without touching Insert or Delete.

Here's what I need:

1. Replace the notification logic in ReplaceAll (both the drain select block and the send select block, plus the separate up-front if err := ctx.Err(); err != nil { return err } check) with this single self-draining send loop, right after c.mutex.Unlock():

      // Notify consumers of the full replacement. SnapshotReplaced supersedes any
      // per-ID signal already queued (it means "reconcile everything", which
      // subsumes any single object change). The loop guarantees SnapshotReplaced is
      // delivered without blocking indefinitely: if the single-slot buffer is full,
      // it drains the stale value and retries the send. This decouples snapshot
      // completion from consumer readiness and prevents the startup/reconnect
      // deadlock that would otherwise wedge the resource stream and trip the
      // liveness probe.
```
Send:
      for {
              select {
              case <-ctx.Done():
                      return ctx.Err()
              case c.resourceChanged <- SnapshotReplaced:
                      break Send
              case <-c.resourceChanged:
                      // Buffer was full; drained a stale queued value, retry the send.
              }
      }
```

2. In the "mark ready" select earlier in ReplaceAll (the one that does close(c.ready)), make case <-ctx.Done(): return ctx.Err() the first case.
3. Please don't add any new mutex/lock, and don't modify Insert or Delete.
4. Update the ReplaceAll doc comment and any test comment that says the

rlenglet (dismissed):

rlenglet (approved):

PR #451, opened 2026-07-29
Remove Falco fallback when no supported CNI is detected

pavankumarinnamuri:
Previously the operator fell back to the Falco flow collector whenever no supported CNI (Cilium+Hubble, OVN-Kubernetes, AWS VPC CNI) was detected. This started a Falco listener that silently received nothing when Falco was not deployed, masking the lack of flow visibility.

DetectFlowCollector now reports FLOW_COLLECTOR_DISABLED and returns a nil factory when no supported CNI is found, logging a warning instead of falling back. main.go only registers the flow-collector stream when a collector is available; the network-flows stream still opens and sends keepalives so the backend sees the cluster with flow collection reported as disabled.

### DIFF
--- a/internal/controller/stream/flows/detect.go
+++ b/internal/controller/stream/flows/detect.go
@@ -13,7 +13,6 @@ import (
 	"github.com/illumio/cloud-operator/internal/controller/stream"
 	"github.com/illumio/cloud-operator/internal/controller/stream/flows/awsvpccni"
 	"github.com/illumio/cloud-operator/internal/controller/stream/flows/cilium"
-	"github.com/illumio/cloud-operator/internal/controller/stream/flows/falco"
 	"github.com/illumio/cloud-operator/internal/controller/stream/flows/ovnk"
 	"github.com/illumio/cloud-operator/internal/pkg/tls"
 )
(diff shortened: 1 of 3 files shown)


internal/controller/stream/flows/detect.go
```diff
@@ -148,15 +147,12 @@ func DetectFlowCollector(ctx context.Context, config CollectorConfig) (pb.FlowCo
...
 	}
 
-	// Default to Falco
-	config.Logger.Info("Using Falco flow collector")
+	// No supported CNI detected. Flow collection is disabled: the operator does not
+	// fall back to any collector. A nil factory signals the caller not to register a
+	// flow-collector stream. The reported type is FLOW_COLLECTOR_DISABLED so the
+	// backend can surface the lack of flow visibility.
+	config.Logger.Warn("No supported CNI detected; network flow collection is disabled. " +
+		"Configure a supported CNI (Cilium+Hubble, OVN-Kubernetes, or AWS VPC CNI) to enable flow visibility.")
```
rlenglet:
```suggestion
	// No supported flow exporter detected. Flow collection is disabled: the operator does not
	// fall back to any collector. A nil factory signals the caller not to register a
	// flow-collector stream. The reported type is FLOW_COLLECTOR_DISABLED so the
	// backend can surface the lack of flow visibility.
	config.Logger.Warn("No supported flow exporter detected; network flow collection is disabled. " +
		"Configure a supported CNI plugin (Cilium+Hubble, OVN-Kubernetes, or AWS VPC CNI) to enable flow visibility.")
```

rlenglet (changes requested):
LGTM. Just one minor suggestion on the wording of the Warn log message.

rlenglet (approved):

PR #452, opened 2026-07-30
Send workload pod-template labels to CloudSecure

ganesh-talla:
## Context

Workload controllers — Deployment, ReplicaSet, StatefulSet, DaemonSet, Job, CronJob — are already watched, but they flow through the generic path in `ConvertMetaObjectToMetadata`, which has no case for them. As a result we only sent each workload's own top-level `metadata.labels` — never `spec.template.metadata.labels` (the labels Kubernetes stamps onto the pods those controllers create) nor `spec.selector.matchLabels` (how the controller selects its pods).

This enriches the workload objects we already send so the backend can correlate pods to their owning workload by label. Direction is operator → CloudSecure, on the existing `SendKubernetesResources` stream — no new stream or RPC.

### DIFF
--- a/api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
+++ b/api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
@@ -51,6 +51,7 @@ message KubernetesObjectData {
     KubernetesCiliumNetworkPolicyData cilium_network_policy = 104;
     KubernetesCiliumClusterwideNetworkPolicyData cilium_clusterwide_network_policy = 105;
     KubernetesCiliumCIDRGroupData cilium_cidr_group = 106;
+    KubernetesWorkloadData workload = 107;
   }
 }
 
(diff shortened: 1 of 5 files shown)


api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
```diff
@@ -94,6 +95,19 @@ message KubernetesPodData {
   repeated string ip_addresses = 1;
 }
 
+// KubernetesWorkloadData contains labels common to workload controller objects
+// (Deployment, ReplicaSet, StatefulSet, DaemonSet, Job, CronJob): the labels
+// stamped onto the pods they create, so the backend can correlate pods to their
+// owning workload by label.
```
```suggestion
// given to the pods they create, so the backend can correlate pods to their
// owning workload by labels.
```

internal/convert/k8s_resources.go
```diff
@@ -33,10 +33,71 @@ func NewCoreResourceConverter(clientset kubernetes.Interface, logger *zap.Logger
...
 }
 
+// workloadKinds is the set of controller kinds whose pod-template labels we send.
+var workloadKinds = map[string]struct{}{
+	"Deployment":  {},
+	"ReplicaSet":  {},
+	"StatefulSet": {},
+	"DaemonSet":   {},
+	"Job":         {},
+	"CronJob":     {},
```
rlenglet: List those in lexicographic order.

rlenglet (changes requested):
Looks good! Just a couple of cosmetic comments.

api/illumio/cloud/k8sclustersync/v1/k8s_info.proto
ganesh-talla: Done in 87fd978 — applied the suggested wording and regenerated the .pb.go so the generated comment stays in sync.

internal/convert/k8s_resources.go
ganesh-talla: Done in 87fd978 — reordered workloadKinds lexicographically (CronJob, DaemonSet, Deployment, Job, ReplicaSet, StatefulSet).

rlenglet (approved):
LGTM, thanks!

pavankumarinnamuri (commented):
Please do not merge this now; it will go in the next release. You can pull  your branch's head commit into the backend.

ganesh-talla:
> Please do not merge this now; it will go in the next release. You can pull your branch's head commit into the backend.
I also have a k8sclustersync PR up which pulls this branch's head commit in BE.

pavankumarinnamuri (approved):

